PR #12, opened 2020-01-14
added make command to deploy to google run deploy

awalias:
@kiwicopple could use your input.

1) Should this be a part of the public repo or is it specific to internal infrastructure?
2) if it's useful for public, where should I put dependency information (needs gcloud cli)
3) <s>also where to put instructions on how to get deploy keys (secrets.json)</s> actually not needed
4) comments on make file addition style welcome

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,8 @@ ENV MIX_ENV=prod \
     DB_PASSWORD=postgres \
     DB_PORT=5432 \
     MIX_ENV=prod \
-    APP_PORT=4000 \
-    APP_HOSTNAME=localhost \
+    PORT=4000 \
+    HOSTNAME=localhost \
     SECRET_KEY_BASE=SOMETHING_SUPER_SECRET
 
 RUN apt-get update
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ help:
 	@echo "make pull.{dev}       # pull all the latest docker images"
 	@echo "make local:db         # start a Postgres database on port 5432"
 
+	@echo "\nGOOGLE RUN\n"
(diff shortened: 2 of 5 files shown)


kiwicopple:
> Should this be a part of the public repo or is it specific to internal infrastructure?

Definitely belongs in the repo. 
I think there is actually a ‘Cloud Run’ button that we can put in the readme that allows for one-click installs ( like the Heroku button). I remember seeing it on HN, will dig it up. 

> if it's useful for public, where should I put dependency information (needs gcloud cli)

I think keep the Make command, and your comments in the help section are great - It’s bound to evolve, so we will add more docs at that point

kiwicopple:
See Cloud Run buttons and related discussion here: 
https://news.ycombinator.com/item?id=20758896

[This comment](https://news.ycombinator.com/item?id=20763458) is a bit worrying though.

> Now I just wait for websocket support to move my backend to cloud run as well 

Perhaps cloud run doesn't support websockets? Or perhaps it doesn't detect an open connection and then shuts it down when the connection is still alive?

awalias:
from https://cloud.google.com/run/docs/issues 

"gRPC and WebSockets support

Cloud Run (fully managed) currently does not support HTTP streaming. Therefore, inbound requests with protocols like WebSockets and gRPC (streaming RPCs) are not supported.

As an exception, Cloud Run (fully managed) services support unary gRPC for inbound requests.

For outbound requests, both WebSockets and gRPC are supported on Cloud Run (fully managed)."

super annoying :(

kiwicopple:
I am going to merge the update for realtime to use PORT and HOSTNAME instead of APP_PORT and APP_HOSTNAME. This means we will need to change some of the docker files on monorepo. I  think the naming is cleaner.

PR #19, opened 2020-02-19
Fix typo in Readme: UDPATE -> UPDATE

giovannibonetti:


kiwicopple:
Yes, that one is quite important 😂 

Thanks for contributing @giovannibonetti

Issue #33, opened 2020-05-19
Refactor the notifiers logic and plan for connectors


kiwicopple:
At the moment our decoder gen_server is doing a lot of work. We should (maybe) refactor it into smaller gen_servers, specifically this part here:

https://github.com/supabase/realtime/blob/1d7cf1162eb40e5f3c6c9349e2509cd7f7f8914c/server/lib/realtime/replication.ex#L202

This part is where we parse the data changes and then send it to various channel topics:

![image](https://user-images.githubusercontent.com/10214025/82281557-3597c500-99c4-11ea-9acb-6b17b71a244f.png)

It will be better to separate this out. because we eventually want this structure (comments/improvements:

![image](https://user-images.githubusercontent.com/10214025/82281824-e9995000-99c4-11ea-902e-0c8d0d4629bf.png)

fracek:
I think this one was fixed by #36. You can add connectors by adding a new genserver and notify it from the replication genserver.

kiwicopple:
Great - remaining questions:

- let's say I create a genserver called "connectors" which stores all the different locations that a customer wants to send the events too
- I create a genserver called "webhook" which will POST the event to a URL

As a customer, I want to start the server and have it send the events to 2 endpoints:

- https://webhook.site/44e4457a-77ae-4758-8d52-17efdf484853
- https://webhook.site/c451c09e-6363-45fe-aeff-9cf577681058

How would I pass those to the server? Env vars don't seem like a good way, as it needs to be an array (and other connectors might need additional config).

Ideas:

- Maybe some sort of JSON file that the "connectors" genserver can read from? To update the sources/connectors we can then just restart the "connectors" genserver which will read the file again?

fracek:
One option is to use config providers in `releases.exs` and add a [provider](https://hexdocs.pm/elixir/Config.Provider.html) for the configuration.

Another (and IMHO better) option is to use an agent to hold the configuration. At the beginnig you can keep it simple (simply read from a json configuration file), but can be easily improved by adding configuration through HTTP endpoints/consul etc. The important thing is that all processes read the configuration from the agent and subscribe to be notified when the configuration changes so that the data they need.

fracek:
Can you expand more on how you plan to allow customers to configure the service from the dashboard? How will this configuration be accessible by realtime? (http, direct db access, etc.) I'm trying to think of a design that's easy to setup but that can also be expanded in the future to let customers load the configuration from multiple sources.

[...]

PR #36, opened 2020-05-20
Move realtime subscribers notification to separate module

fracek:
This is related to #33.

I moved the notification logic to a separate genserver that is a sibling of the replication genserver.


kiwicopple:
Solid. this is a great improvement thanks @fracek

PR #46, opened 2020-05-30
Add configurable webhook connector

fracek:
This is related to what we discussed in #33. 

* `ConfigurationManager` module that for now simply reads a json configuration file if specified by the user.
 * `TransactionFilter` to parse and check if the transaction filter match a transaction. It only checks for schema/table name but should be fairly easy to extend it to support more types of filters
 * `WebhookConnector` that, given a set of webhook, will filter the one that are valid for the given connection and calls them.


server/lib/realtime/transaction_filter.ex
```diff
@@ -0,0 +1,100 @@
...
+      [schema, table] -> {:ok, %Filter{schema: schema, table: table, condition: nil}}
+      _ -> {:error, "malformed relation filter"}
+    end
+  end
+
+  defp name_matches(nil, change_name), do: true
+  defp name_matches(filter_name, change_name) do
+    filter_name == change_name
+  end
+end
```
kiwicopple, line 100: This is tidy

kiwicopple:
This is awesome thanks @fracek . I'll merge it in.
 
Also, I've added you to our org so you can just branch. I can't wait to try this out on some of the databases

PR #52, opened 2020-06-05
refactor: extract runtime config to releases.exs

soedirgo:
This will probably be affected by #48 in the future, but for now we should follow the recommended way to set runtime configs as explained [here](https://hexdocs.pm/mix/1.9.0/Mix.Tasks.Release.html#module-runtime-configuration).

Closes #31.

PR #53, opened 2020-06-06
fix: incorrect env config handling

soedirgo:
The config handling in #52 was incorrect, so this PR is to remedy that. Other changes I made:

- Moved the handling of unset envs to `config.exs` and `releases.exs`--this shouldn't be `application.ex`'s concern as default values can be set at build/release.
- It should now be possible to run the app with `mix phx.server` (without using mix releases). On the flip side, the configs in `config.exs` (used in `mix phx.server`) should mirror the defaults in `releases.exs` (used in `mix release`).

### DIFF
--- a/server/config/releases.exs
+++ b/server/config/releases.exs
@@ -1,14 +1,17 @@
 import Config
 
-app_hostname = System.get_env("HOSTNAME")
-app_port = System.get_env("PORT")
-db_host = System.get_env("DB_HOST")
-db_port = System.get_env("DB_PORT")
-db_name = System.get_env("DB_NAME")
-db_user = System.get_env("DB_USER")
-db_password = System.get_env("DB_PASSWORD")
-db_ssl = System.get_env("DB_SSL")
-slot_name = System.get_env("SLOT_NAME")
+# These defaults mirror the ones in config.exs, remember not to change one
+# without changing the other.
+app_hostname = System.get_env("HOSTNAME", "localhost")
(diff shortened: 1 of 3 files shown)


server/config/releases.exs
```diff
@@ -1,14 +1,17 @@
...
+app_hostname = System.get_env("HOSTNAME", "localhost")
+app_port = String.to_integer(System.get_env("PORT", "4000"))
+db_host = System.get_env("DB_HOST", "localhost")
+db_port = String.to_integer(System.get_env("DB_PORT", "5432"))
+db_name = System.get_env("DB_NAME", "postgres")
+db_user = System.get_env("DB_USER", "postgres")
+db_password = System.get_env("DB_PASSWORD", "postgres")
+# HACK: There's probably a better way to set boolean from env
+db_ssl = System.get_env("DB_SSL", "true") === "true"
+slot_name = String.to_atom(System.get_env("SLOT_NAME", "temporary"))
```
kiwicopple, line 14: this one will fail if we want to pass an actual "slot_name" in.

We use :temporary atom for the default, but it could also be something like "slot_123" - and this will get converted to :slot_123. We aren't doing this yet, so this is fine, but could you please create a new issue to allow for both scenarios @soedirgo ?

soedirgo: 👍 wasn't actually sure about this part (as well as `configuration_file`). I'll amend the commit.

PR #76, opened 2020-11-05
Support event filter

### LINKED ISSUE #73: Event parser
# Feature request

Right now we are blindly sending all events over the websockets. For various reasons (primarily security), a developer may want to filter which events are sent out. In the future this could also extend to include JSON transformer discussed [here](https://github.com/supabase/realtime/issues/66#issue-686574252).

## Describe the solution you'd like

We already have the Configuration manager for adding webhooks:

```ex
    config = [
	%Configuration.Webhook{ 
	  event: "*",
	  relation: "public:todos",
        }
      ]
```

we can probably re-use this for Realtime:

```ex
    config = [
	%Configuration.Realtime{
	  event: "*",
	  relation: "public:todos",
        },
	%Configuration.Realtime{
	  event: "*",
	  relation: "public:users:id",
        },
      ]
```

If this config is present, the realtime server will stop sending events for everything, and only send events matching the rules specified, although I can see cases where people want to allow everything, and then uses these rules to enrich _some_ routes. 

In this case the Realtime server will only send events that happen on the `todos` table (including all the columns? ie: `todos:id.eq.1`), and on the `public:users:id.eq.*` channel.

This is a workaround using security through obscurity. For example, if the user ID is a `uuid`, then it will be fairly difficult to listen to other users' data - `supabase.from('users:id = eq.21a285d7-6028-4967-a2a9-fb5afdc9a4b4').on('*', () => {})`

Finally, in the future we may extend this with a transformer:

```ex
    config = [
	%Configuration.Realtime{
	  event: "*",
	  relation: "public:todos",
          transformer: '.[0] | {message: .commit.message, name: .commit.committer.name}' # fake, using jq filter
        }
      ]
```

## Describe alternatives you've considered

None come to mind, but I'm not particularly happy with this part: 

> I can see cases where people want to allow everything, and then uses these rules to enrich _some_ routes

Also I don't particularly like how the "Channels" config is now mixed in with the "Connectors" config.

## Additional context

- Related: [Configuration Manager](https://github.com/supabase/realtime/issues/48) for handling config
- Related: [JSON parser](https://github.com/supabase/realtime/issues/66) for filtering the payload data

edit: @awalias changed:
`supabase.from('users:id.eq.21a285d7-6028-4967-a2a9-fb5afdc9a4b4').on('*', () => {})` to
`supabase.from('users:id = eq.21a285d7-6028-4967-a2a9-fb5afdc9a4b4').on('*', () => {})`

soedirgo:

Feature.

Closes #73.

## What is the new behavior?

Default config is:

```elixir
realtime: [
  %Realtime{
    event: "*",
    relation: "*"
  },
  %Realtime{
    event: "*",
    relation: "*:*"
  },
  %Realtime{
    event: "*",
    relation: "*:*:*"
  }
]
```

This allows all events to be broadcasted to all room types:
- `realtime:*`
- `realtime:<schema>`
- `realtime:<schema>:<table>`
- `realtime:<schema>:<table>:<column>`

## Additional context

I removed the leading `realtime:` for `relation` from our discussion. Looks cleaner imo, but let me know if you want to add it back.

One problem: there's no way to 1. allow events to `realtime:*`, and 2. allow events only for specific schemas, at the same time. This is because the relevant config `relation` is the same: `*`

### DIFF
--- a/dev/postgres/mnt/init-permissions.sh
+++ b/dev/postgres/mnt/init-permissions.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-set -e
-
-{ echo "host replication $POSTGRES_USER 0.0.0.0/0 trust"; } >> "$PGDATA/pg_hba.conf"
--- a/examples/next-js/pages/index.js
+++ b/examples/next-js/pages/index.js
@@ -19,18 +19,21 @@ export default class Index extends React.Component {
   componentDidMount() {
     this.socket.connect()
     this.addChannel('realtime:*')
+    this.addChannel('realtime:public')
     this.addChannel('realtime:public:todos')
     this.addChannel('realtime:public:users')
-    this.addChannel('realtime:public:users:id=eq.2')
+    this.addChannel('realtime:public:users:id=eq.16')
+    this.addChannel('realtime:public:users:id=eq.17')
+    this.addChannel('realtime:public:users:id=eq.18')
     this.fetchData()
   }
   addChannel(topic) {
     let channel = this.socket.channel(topic)
     channel.on('INSERT', msg => this.messageReceived(topic, msg))
-    // channel.on('*', msg => console.log('INSERT', msg))
-    // channel.on('INSERT', msg => console.log('INSERT', msg))
-    // channel.on('UPDATE', msg => console.log('UPDATE', msg))
-    // channel.on('DELETE', msg => console.log('DELETE', msg))
+    channel.on('*', msg => console.log('*', msg))
+    channel.on('INSERT', msg => console.log('INSERT', msg))
+    channel.on('UPDATE', msg => console.log('UPDATE', msg))
+    channel.on('DELETE', msg => console.log('DELETE', msg))
     channel
       .join()
       .receive('ok', () => console.log('Connecting'))
--- a/server/lib/realtime/configuration.ex
+++ b/server/lib/realtime/configuration.ex
@@ -1,15 +1,34 @@
 defmodule Realtime.Configuration do
   defmodule(WebhookEndpoint, do: defstruct([:endpoint]))
   defmodule(Webhook, do: defstruct([:event, :relation, :config]))
+  defmodule(Realtime, do: defstruct([:event, :relation]))
 
-  defmodule(Configuration, do: defstruct([:webhooks]))
+  defmodule(Configuration, do: defstruct([:webhooks, :realtime]))
 
   @doc """
   Load Configuration from a json file.
   """
   def from_json_file(nil) do
-    {:ok, %Configuration{webhooks: []}}
+    {:ok,
+     %Configuration{
+       webhooks: [],
+       realtime: [
+         %Realtime{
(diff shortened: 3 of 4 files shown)


kiwicopple:
> One problem: there's no way to 1. allow events to realtime:*, and 2. allow events only for specific schemas, at the same time.

Yeah I thought this might be the case from our discussion last Friday. We may have to introduce a special symbol to represent "whole database" to avoid breaking changes. eg:

```ex
realtime: [
  %Realtime{
    event: "_",
    relation: "*"
  },
  %Realtime{
    event: "*",
    relation: "*"
  },
  %Realtime{
    event: "*",
    relation: "*:*"
  },
  %Realtime{
    event: "*",
    relation: "*:*:*"
  }
]
```

I don't know if this is a good idea, it feels a bit dirty, so worth a discussion

kiwicopple:
just a thought - we might want to add an "enabled" flag, so that we can turn this off without deleting the config

soedirgo:
Do you mean adding an `enabled` field to each config? Wouldn't commenting out the configs be fine?

kiwicopple:
Yeah exactly 

> Wouldn't commenting out the configs be fine?

I'm thinking of when we are managing the config via an API. We will want to fetch the config and check against a boolean ( in the Dashboard we will add toggle)

soedirgo:
Hmm, shouldn't this be handled on the frontend side? I'm thinking there'd be a page that lists all configs, each config entry will be associated with a yaml entry, and every time the user toggles/creates/deletes an entry, we create a new `.yml` and `SIGUSR1` the realtime server.

Assuming of course we're handling config update like PostgREST, with the single source of truth being the `.yml` file (which tbh I couldn't think of a better way).

kiwicopple:
> Assuming of course we're handling config update like PostgREST, with the single source of truth being the .yml file (which tbh I couldn't think of a better way).

Yeah the yaml is a short term approach and isn't best practice for dynamically updating config. We will add a Postgres database + API to this server, and users will be able to make changes without a yml file. So really we are engineering in a future state. It doesn't have to be now, though it seems simple enough.

What are the next steps on this PR? Can I help?

soedirgo:
Actually, I just realized the problem you're referring to: if the `.yml` is the source of truth we can't simply do toggle (the old config will just be gone from the page). Still, not sure if an `enabled` field is the way to go, I haven't seen this approach in the wild.

For this PR, I'm not sure what the target is right now, previously I was assuming JSON and the simple config style we started with, but we started going off on another direction.

Rather than wait until we decide on the "ideal" config, I say we start simple and iterate on that (maybe in a feature branch). So beyond the change in `notify_subscribers/1`, we need to:
- Switch to `.yml` (should we?),
- Deal with the "one problem" at the top,

and call it a day.

On that note, do you have an example in mind for how the dynamic config thing should be like? Zookeeper maybe?

kiwicopple:
OK great - let me pick this up. Agreed, we can close this out and iterate on the rest later!

PR #77, opened 2020-11-22
Version bump pg to 8.5.1

w3b6x9:

* `pg` version bump from `7.17.1` to `8.5.1`
* add additional steps to Next js example's `Getting Started` section

Node version: 14.11.0

Next js example doesn't work.

## What is the new behavior?

Next js example now works as intended.

## Additional context

I had trouble getting the Next js example up and running and it had to do with my Node version. I tried it with my default node v14.11.0 and it didn't work but it worked with v13.7.0. 

I think it would be best to just version bump pg so now it works with more major versions of node. Tried it with v10.21.0, v13.7.0 and v14.11.0 and they all worked.


kiwicopple:
Legend, thanks @w3b6x9

PR #82, opened 2020-11-30
Add database connection retries

w3b6x9:

Prevents a temporary database connection error from shutting down entire server.

When connection to database fails then the server shuts down because it tries to restart the epgsql process too quickly.

## What is the new behavior?

When connection to database fails then the server will retry with exponential backoff and jitter, preventing server from shutting down.

## Additional context

Tested this locally by restarting postgres.

~**NOTE:** I think I discovered a better way as I was wrapping up this PR. Please hold off on merging.~

**NOTE:** Ready for review!


kiwicopple:
@soedirgo could you please review this one? I think it will solve some of the 500's we're seeing in production actually

soedirgo:
Looks good to me, tried this on a Supabase Postgres AMI and it can carry on connecting even when I blocked inbound connections inbetween. Nice work @w3b6x9!

PR #89, opened 2020-12-16
Build arm64 binaries

### LINKED ISSUE #86: Publish arm build for Graviton
https://github.com/supabase/infrastructure/issues/343

add arm64 binary download to release page

darora:
# Changes introduced

- Breaks out build jobs into separate github jobs to allow for parallel execution, given there are no inter-dependencies
- Adds a new arm64 docker build
- Pins the OS for the native x86 build to u18 (from ubuntu-latest), which I _believe_ will imminently break once u-latest moves to u20 https://github.com/actions/virtual-environments/issues/1816

The arm64 docker build is only executed on releases, so the increased CI times shouldn't present a significant overhead for iterating on PRs. Additionally, an extremely straightforward speed-up would be to publish a base docker image with the appropriate OTP/elixir versions installed on the u18 base.

Note that unlike our other services, were publishing a separate tarball for arm64 here, rather than a single tarball with two binaries.

Fixes #86 

## Testing performed

On a graviton instance, connecting against a local test database, the service came up w/o any issues. However, I haven't performed any testing of the actual functioning of the service beyond that.

```
ubuntu@ip-172-31-39-228:~/realtime$ uname -a
Linux ip-172-31-39-228 5.4.0-1029-aws #30~18.04.1-Ubuntu SMP Tue Oct 20 11:08:38 UTC 2020 aarch64 aarch64 aarch64 GNU/Linux
ubuntu@ip-172-31-39-228:~/realtime$ ./bin/realtime start
2020-12-16 02:30:00.342 [info] Running RealtimeWeb.Endpoint with cowboy 2.6.3 at :::4000 (http)
2020-12-16 02:30:00.343 [info] Access RealtimeWeb.Endpoint at http://localhost:4000
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,119 +3,181 @@ name: Release
 on:
   push:
     branches:
-      - master
+      - da/arm64
 
 jobs:
-  release:
-    name: Release
+  # release:
+  #   name: Release
 
+  #   runs-on: ubuntu-latest
+
+  #   steps:
+  #     - uses: actions/checkout@v2
+
+  #     - name: Get version before
+  #       run: echo VERSION_BEFORE=$(curl -s https://api.github.com/repos/supabase/realtime/releases/latest | jq .name -r) >> $GITHUB_ENV
+
+  #     - name: Release on GitHub
+  #       run: npx semantic-release -p        \
+  #         @semantic-release/commit-analyzer \
+  #         @semantic-release/github          \
+  #         @semantic-release/release-notes-generator
+  #       env:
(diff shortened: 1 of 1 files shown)


darora:
Example run of the arm64 release task: https://github.com/supabase/realtime/actions/runs/424718176

soedirgo (approved):
Tested this on a Supabase Postgres instance. Nice one @darora 👍

darora:
Note that I've pushed another commit eb6140a , which should be a no-op for now, but will ensure that the x86 builds don't break for our prod envs once the `ubuntu-latest` OS moves to ubuntu 20.

PR #97, opened 2021-01-06
fix: handle TRUNCATE events

soedirgo:

Bug fix.

[This code](https://github.com/supabase/realtime/blob/1220728423ed012ca9dfcd18e79188e4bb96e1fe/server/lib/realtime/subscribers_notification.ex#L74-L98) is not happy when it receives a `TRUNCATE` event. `TRUNCATE` events are not broadcasted.

## What is the new behavior?

`TRUNCATE` events are broadcasted (added the event type in the default config). Event filters are applied (schema and table only). A `TRUNCATE` event looks like this on the client:

```json
{
  "commit_timestamp": "2021-01-06T17:26:35Z",
  "schema": "public",
  "table": "bar",
  "type": "TRUNCATE"
}
```

## Additional context

Re: the discussion on Slack: turns out I was wrong, the decoder + adapter handles `TRUNCATE` in a way that also gives schema and table information. The resulting `TruncatedRelation` is what's received in `SubscribersNotification`.

### DIFF
--- a/server/lib/realtime/subscribers_notification.ex
+++ b/server/lib/realtime/subscribers_notification.ex
@@ -49,7 +49,7 @@ defmodule Realtime.SubscribersNotification do
       table_topic = "#{schema_topic}:#{change.table}"
       # Logger.debug inspect(change, pretty: true)
 
-      # Get only the config which includes this event type (INSERT | UPDATE | DELETE)
+      # Get only the config which includes this event type (INSERT | UPDATE | DELETE | TRUNCATE)
       event_config =
         Enum.filter(realtime_config, fn config ->
           change.type in config.events
@@ -95,6 +95,9 @@ defmodule Realtime.SubscribersNotification do
               end
             end)
           end
+
+        "TRUNCATE" ->
+          nil
       end
     end
   end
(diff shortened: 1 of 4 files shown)


server/lib/realtime/subscribers_notification.ex
```diff
@@ -95,6 +95,9 @@ defmodule Realtime.SubscribersNotification do
               end
             end)
           end
+
+        "TRUNCATE" ->
+          nil
```
soedirgo, line 100: Deliberately empty - `TRUNCATE` notification is done in "Shout to specific table" above.

kiwicopple:
Awesome, thanks for adding this one Bobbie. 

Still not sure how we can handle this on the client. I guess there is no way to know what data is truncated without re-querying the database?

Either way, this is good, thanks for the fix 🙏 . Feel free to merge

soedirgo:
> I guess there is no way to know what data is truncated without re-querying the database?

Yeah, since `TRUNCATE` is basically "nuke a table from the orbit". If the user needs the data, they should be using `DELETE`.

PR #98, opened 2021-01-08
fix: replication fails to trap on :init_db_conn

soedirgo:

Bug fix.

When the application fails to connect to the DB at start up (`:econnrefused`), it crashes. This is because `handle_continue/2` returns a `:stop` instead of falling back to the `:EXIT` trap.

## What is the new behavior?

Failing to connect during `init` will run the retry logic instead of crashing.

@w3b6x9 would love to hear your input here. Is this the right behavior?

### DIFF
--- a/server/lib/realtime/replication.ex
+++ b/server/lib/realtime/replication.ex
@@ -66,7 +66,7 @@ defmodule Realtime.Replication do
         {:noreply, %State{state | connection: epgsql_pid}}
 
       {:error, reason} ->
-        {:stop, reason}
+        retry(state)
     end
   end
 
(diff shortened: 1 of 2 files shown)


w3b6x9:
@soedirgo I purposefully set the return to `:stop` so that the user who spins up Realtime server gets feedback immediately.

This way they know that if the server crashed on start then either their db config was incorrect or their db server is down. In either case, they can confirm on their end and try to get a successful db conn. After that, if db connection is refused, Realtime server will retry and user will have a higher degree of confidence that it'll eventually connect and continue to work as expected.

However, if you think this is necessary then I recommend

```
Kernel.send(self(), {:EXIT, self(), "db conn refused"})
{:noreply, state}
```

instead of

```
handle_info({:EXIT, self(), reason}, state)
```

or just extract the retry logic from `handle_info({:EXIT, self(), reason}, state)` function to another function and have both `handle_continue` and `handle_info({:EXIT, self(), reason}, state)` call that new function.

soedirgo:
Got it, thanks! I think retrying during initial connection is preferred here, since the user should be able to detect failures from stderr. Also, one connection attempt per 5mins doesn't sound too bad...

PR #100, opened 2021-01-10
Update Realtime README breaking change version from 0.9.8 to 0.10.0

w3b6x9:

Realtime README update

Breaking change is v0.9.8

## What is the new behavior?

Breaking change is v0.10.0

## Additional context

Assumed the next release was going to be v0.9.8 but it's actually v0.10.0.


w3b6x9:
@soedirgo quick README update. Thanks!

PR #101, opened 2021-01-13
fix: huge WAL caused by slow replication

### LINKED ISSUE #81: Large inserts choke micro instances
inserting 1 million rows at the same time cause excessive CPU usage for an extended period afterwards, along with ~50% RAM:

image

test code here: https://raw.githubusercontent.com/supabase/benchmarks/feat/firestore-benchmarks/db/migrations/20201124141199_read_data.sql 

ruled out COPY being the issue since the following snippet also results in the same behaviour:

```sql
do $$
begin
for r in 1..1000000 loop
insert into public.read(id, slug) values(r, r);
end loop;
end;
$$;
```

Restarting the service (process) does not resolve the issue

soedirgo:

Bug fix.

Replication was slow because of calling `Enum.reverse/1` on the previous messages in the transaction on every message, which is both incorrect ((0), (1), (2) arrives as (1), (0), (2)) and unnecessary (we just need to do it once on `Commit`).

`handle_call/3` has a default timeout of 5s. Large messages (e.g. inserting 1 million rows) may take longer that this, which causes the GenServer to crash. Set it to `:infinity` instead.

## What is the new behavior?

Fix the above issues.

## Additional context

Closes #81.


server/lib/realtime/replication.ex
```diff
@@ -282,7 +282,7 @@ defmodule Realtime.Replication do
 
     %{
       state
-      | transaction: {lsn, %{txn | changes: Enum.reverse(truncated_relations ++ txn.changes)}}
+      | transaction: {lsn, %{txn | changes: Enum.reverse(truncated_relations) ++ txn.changes}}
```
w3b6x9, line 285: @soedirgo small optimization: 

```suggestion
      | transaction: {lsn, %{txn | changes: Enum.reverse(truncated_relations, txn.changes)}}
```

See: https://hexdocs.pm/elixir/Enum.html#reverse/2

Further optimization:

Since `process_message(%Truncate{} = msg, state)` is now doing `for` loop, `reverse`, and `concat`, we can combine all those operations like this:


```
  defp process_message(
         %Truncate{truncated_relations: truncated_relations},
         %State{transaction: {lsn, txn}} = state
       ) do
    truncated_relations_changes =
      generate_truncated_relations(state, truncated_relations, txn.changes)

    %{state | transaction: {lsn, %{txn | changes: truncated_relations_changes}}}
  end

  defp generate_truncated_relations(
         %State{relations: relations} = state,
         [truncated_relation | truncated_relations],
         acc
       ) do
    %{namespace: namespace, name: name} = Map.get(relations, truncated_relation)

    generate_truncated_relations(state, truncated_relations, [
      %TruncatedRelation{
        type: "TRUNCATE",
        schema: namespace,
        table: name
      }
      | acc
    ])
  end

  defp generate_truncated_relations(_state, [], acc) do
    acc
  end
```

soedirgo: > Small optimization

Wow, TIL!

> Since process_message(%Truncate{} = msg, state) is doing for loop, reverse, and concat, we can combine all those operations like this

Hmm, the imperative code is much clearer to me. Truncate is not in the hot path, so clarity trumps here.

kiwicopple:
Nice one @soedirgo - seems like a big win. lgtm 👍

kiwicopple (approved):

awalias:
wow this is amazing! will get it shipped in KPS once merged

soedirgo:
Oh yeah just a heads up @awalias, this version needs `JWT_SECRET` env for securing channels

Issue #102, opened 2021-01-13
Truncate Option 3 (Cascade + Restart Identity) Missing


w3b6x9:
# Bug report

## Describe the bug

The following function is missing `3 -> [:casecade, :restart_identity]` from `case options`:

```
  defp decode_message_impl(
         <<"T", number_of_relations::integer-32, options::integer-8, column_ids::binary>>
       ) do
    truncated_relations =
      for relation_id_bin <- column_ids |> :binary.bin_to_list() |> Enum.chunk_every(4),
          do: relation_id_bin |> :binary.list_to_bin() |> :binary.decode_unsigned()

    decoded_options =
      case options do
        0 -> []
        1 -> [:cascade]
        2 -> [:restart_identity]
      end

    %Truncate{
      number_of_relations: number_of_relations,
      options: decoded_options,
      truncated_relations: truncated_relations
    }
  end
```

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Select `Cascade` and `Restart Identity` when truncating data from table.
2. Genserver terminates and raises following error: `(CaseClauseError) no case clause matching: 3`

## Expected behavior

1. Select `Cascade` and `Restart Identity` when truncating data from table.
2. Everything continues to work without error.

## Additional context

Also, `commit_timestamp` is missing from `%TruncatedRelation` in `process_message(%Truncate{} = msg, state)`

PR #103, opened 2021-01-16
Add Truncate option 3 and add commit_timestamp to all DB Changes

w3b6x9:

Bug fix

Selecting Cascade and Restart Identity when truncating data from table results in the following error: `(CaseClauseError) no case clause matching: 3`

Addresses issue #102. 

## What is the new behavior?

Selecting Cascade and Restart Identity when truncating data from table doesn't raise error and everything works as expected.

## Additional context

- Tested to make sure both realtime channel and webhook messages contained the right commit timestamps.
- `%Truncate{}` options (e.g. `2 -> [:restart_identity]`) aren't used anywhere else so we can consider removing altogether.
- I did some refactoring while making sure that all the Changes (i.e. `NewRecord`, `UpdatedRecord`, `DeletedRecord`, `TruncatedRelation`) had `commit_timestamp`s during creation.

### DIFF
--- a/server/lib/decoder/decoder.ex
+++ b/server/lib/decoder/decoder.ex
@@ -94,7 +94,6 @@ defmodule Realtime.Decoder do
   end
 
   defp decode_message_impl(<<"R", id::integer-32, rest::binary>>) do
-    
     [
       namespace
       | [name | [<<replica_identity::binary-1, _number_of_columns::integer-16, columns::binary>>]]
@@ -190,6 +189,7 @@ defmodule Realtime.Decoder do
         0 -> []
         1 -> [:cascade]
         2 -> [:restart_identity]
+        3 -> [:cascade, :restart_identity]
       end
 
     %Truncate{
--- a/server/lib/realtime/replication.ex
+++ b/server/lib/realtime/replication.ex
@@ -163,11 +163,10 @@ defmodule Realtime.Replication do
     %State{state | conn_retry_delays: []}
   end
 
-  defp process_message(%Begin{} = msg, state) do
-    %{
+  defp process_message(%Begin{final_lsn: final_lsn, commit_timestamp: commit_timestamp}, state) do
+    %State{
       state
-      | transaction:
-          {msg.final_lsn, %Transaction{changes: [], commit_timestamp: msg.commit_timestamp}}
+      | transaction: {final_lsn, %Transaction{changes: [], commit_timestamp: commit_timestamp}}
     }
   end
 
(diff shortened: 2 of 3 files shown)


server/lib/decoder/decoder.ex
```diff
@@ -190,6 +189,7 @@ defmodule Realtime.Decoder do
         0 -> []
         1 -> [:cascade]
         2 -> [:restart_identity]
+        3 -> [:cascade, :restart_identity]
```
w3b6x9, line 192: These options aren't used anywhere so we can consider removing them entirely from `%Truncate{}`. Adding this in so it doesn't raise an error when truncating with Cascade and Restart Identity options.

server/lib/realtime/replication.ex
```diff
@@ -300,23 +356,6 @@ defmodule Realtime.Replication do
   end
 
   defp notify_subscribers(%State{transaction: {_current_txn_lsn, txn}}) do
-    txn
```
w3b6x9, line 303: This reverts https://github.com/w3b6x9/realtime/commit/e47148365ac89201a02c95745074d756ebfdd264. `commit_timestamp` is now added at the source when creating structs (i.e. `NewRecord`, `UpdatedRecord`, `DeletedRecord`, `TruncatedRelation`.

server/lib/realtime/replication.ex
```diff
@@ -209,83 +208,140 @@ defmodule Realtime.Replication do
...
+  defp process_message(
+         %Truncate{truncated_relations: truncated_relations},
+         %State{
+           relations: relations,
+           transaction: {lsn, %{commit_timestamp: commit_timestamp, changes: changes} = txn}
+         } = state
+       )
+       when is_list(truncated_relations) and is_list(changes) and is_map(relations) do
+    new_changes =
+      Enum.reduce(truncated_relations, changes, fn truncated_relation, acc ->
```
w3b6x9, line 317: @soedirgo I know we talked about this and you decided to go with the [imperative code](https://github.com/supabase/realtime/pull/101#discussion_r556978624) when I suggested the [further optimization](https://github.com/supabase/realtime/pull/101#discussion_r556927646). As I was refactoring, I found that using `reduce` encapsulated the more complex optimization logic from my suggestion while making it clearer.

I don't mean to undermine what you decided to go with so more than happy to revert the changes here. Let me know!

soedirgo: 👍 this looks good to me actually, I was mostly worried about introducing extra helper functions.

PR #99, opened 2021-01-08
Add channels auth via JWT verification

w3b6x9:
## What kind of change does this PR introduce?

Feature

## What is the current behavior?

Channels are not secured

## What is the new behavior?

Channels are secured by setting SECURE_CHANNELS=true and JWT_SECRET=your-secret

## Additional context

- By default in development mode everything should still work.
- To test out channels auth set `SECURE_CHANNELS=true` and `JWT_SECRET` (e.g. secret123).  I've been using http://jwtbuilder.jamiekurtz.com to generate JWTs and to play around with expiration and claims.
- I've been using the next-js example and passing the token param to Socket in realtime/examples/next-js/pages/index.js: `this.socket = new Socket(REALTIME_URL, { params: { token: 'token123' } })`

- To validate claims set `JWT_CLAIM_VALIDATORS`(e.g. {"iss": "Online JWT Builder", "aud": "www.example.com"}).

This addresses issue #95.

PR #105, opened 2021-01-19
fix: use JWT_SECRET for SECRET_KEY_BASE

soedirgo:
This is a followup to #99. All this means is it's no longer required for users to specify `SECRET_KEY_BASE` (which we don't really use anyway).


server/config/releases.exs
```diff
@@ -55,13 +55,6 @@ config :realtime,
...
-  System.get_env("SECRET_KEY_BASE") ||
-    raise """
-    environment variable SECRET_KEY_BASE is missing.
-    You can generate one by calling: mix phx.gen.secret
-    """
-
 config :realtime, RealtimeWeb.Endpoint,
   http: [:inet6, port: app_port],
-  secret_key_base: secret_key_base
+  secret_key_base: jwt_secret
```
w3b6x9, line 60: @soedirgo 

> which we don't really use anyway

I think you're right. I don't think Realtime server is using it to sign/verify anything. 

Perhaps we should consider removing it altogether to prevent future confusion?

For folks who are setting `SECURE_CHANNELS` to `false`, `jwt_secret` isn't mandatory. If user doesn't input `jwt_secret`, then `secret_key_base` just defaults to a meaningless empty string.

soedirgo: Yeah, probably better to just remove it. cc @kiwicopple

kiwicopple: agreed!

Issue #107, opened 2021-01-24
SECURE_CHANNELS can't be boolean


kapoko:
# Bug report

SECURE_CHANNELS should be a string, number or null.

## Describe the bug

Setting SECURE_CHANNELS to `true` or `false` in docker-compose file throws error. The docs say this value should be boolean but it is not allowed.

## To Reproduce

Set `supabase/realtime` docker image environment value `SECURE_CHANNELS` to `true` (or `false`).

## Expected behavior

No error, or the docs to specify the value should be `'true'` or `'false'` (with quotes). 

## Screenshots

![image](https://user-images.githubusercontent.com/17455539/105627913-de42d200-5e39-11eb-9a12-32e5210e9101.png)

## System information

- OS: macOS 10.14.6 (Mojave)
- Docker version 19.03.14
- Version of supabase/realtime: v0.10.4
- Version of Node.js: [e.g. 10.10.0]

soedirgo:
Thanks for the report @kapoko!

This looks like a problem with YAML in general (see https://github.com/docker/compose/issues/1788), all environment variables are strings, but I agree that the types in the docs can make it confusing. But since it was mentioned in the context of command line arguments, it's technically correct.

@w3b6x9 any thoughts on this? I think a minor improvement would be to mention examples alongside the options without specifying types.

w3b6x9:
Thanks for bringing this to our attention, @kapoko!

@soedirgo Thanks for looking into this! I agree, let's have another example to better demonstrate how to pass in these env vars. I'll put up a PR, link this issue, and we can discuss futher!

PR #108, opened 2021-01-24
Convert SECURE_CHANNELS value from boolean to string

w3b6x9:

Docs update

## What is the new behavior?

`SECURE_CHANNELS` and `JWT_CLAIM_VALIDATORS` are labeled as `string` values.

## Additional context

I think this is the better call considering they technically are strings coming in, and the application picks it up and converts them to boolean and JSON object.

Related issue: #107

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ ENV MIX_ENV=prod \
     MIX_ENV=prod \
     PORT=4000 \
     HOSTNAME=localhost \
-    JWT_SECRET=SOMETHING_SUPER_SECRET
+    JWT_SECRET=SOMETHING_SUPER_SECRET \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -13,7 +13,8 @@ ENV MIX_ENV=prod \
     MIX_ENV=prod \
     PORT=4000 \
     HOSTNAME=localhost \
-    JWT_SECRET=SOMETHING_SUPER_SECRET
+    JWT_SECRET=SOMETHING_SUPER_SECRET \
+    SECURE_CHANNELS=true
```
w3b6x9, line 17: @soedirgo since this is a `Dockerfile`, doesn't matter if it's `true` or `'true'`.

PR #112, opened 2021-02-03
Update dependencies and other minor chores

fracek:

I updated the dependencies, including bumping Phoenix to 1.5.

I also fixed warnings during compilation.

**Note:** Plug.Session requires a `secret_key_base` to sign/encrypt cookies, I added one for dev but we need one for production. What strategy do you prefer for this type of secret?


soedirgo:
> Plug.Session requires a secret_key_base to sign/encrypt cookies

Is it mandatory or can we omit it? We removed it in https://github.com/supabase/realtime/pull/105 because it doesn't seem to be doing anything useful.

fracek:
> Is it mandatory or can we omit it? We removed it in #105 because it doesn't seem to be doing anything useful.

Yes, Plug complains (and there's an internal server error on `/`) if that config value is not set. I wonder if it makes sense to just remove Plug.Session since I don't think we use it? 

/cc @kiwicopple

w3b6x9:
@fracek this is awesome! Thanks for doing this!

Only caveat with removing `Plug.Session` is that you'll now get an error when navigating to the only view (e.g. `localhost:4000/`) rendered by Phoenix server.

I think we should just hardcode it and add `secret_key_base` to both `config.exs` and `releases.exs`. It's more trouble than it's worth since we're not doing anything meaningful with it and we'll have to pass it back in for prod.

fracek:
> I think we should just hardcode it and add `secret_key_base` to both `config.exs` and `releases.exs`. It's more trouble than it's worth since we're not doing anything meaningful with it and we'll have to pass it back in for prod.

Agree, I wanted to make sure it was the best thing to do before using a not-so-secret secret in production.

kiwicopple:
Alternatively we could reuse the JWT secret - 

```
secret_key_base = System.get_env("JWT_SECRET", "DEFAULT_VALUE")
```

Seems nicer than hardcoding an arbitrary value

fracek:
I'm looking at this again, the `secret_key_base` needs to be at least 64 bytes, which is much longer than the current JWT secret. My proposal is to get this value from env or use a sensible default (for dev).

server/config/dev.exs
```diff
@@ -19,7 +19,8 @@ config :realtime, RealtimeWeb.Endpoint,
   debug_errors: true,
   code_reloader: true,
   check_origin: false,
-  watchers: []
+  watchers: [],
+  secret_key_base: "Kyvjr42ZvLcY6yzZ7vmRUniE7Bta9tpknEAvpxtaYOa/marmeI1jsqxhIKeu6V51"
```
w3b6x9, line 23: @fracek I don't think you have to set this here since you set it in `config.exs`.

server/config/config.exs
```diff
@@ -37,6 +37,9 @@ jwt_secret = System.get_env("JWT_SECRET", "")
 # must equal 1610078130.
 jwt_claim_validators = System.get_env("JWT_CLAIM_VALIDATORS", "{}")
 
+# The secret key base to built the cookie signing/encryption key.
+session_secret_key_base = System.get_env("SESSION_SECRET_KEY_BASE", "Kyvjr42ZvLcY6yzZ7vmRUniE7Bta9tpknEAvpxtaYOa/marmeI1jsqxhIKeu6V51")
```
w3b6x9, line 41: @fracek I think this is fine! We'll have to do this in `releases.exs` as well.

w3b6x9:
Closing this PR in favor of PR #128

PR #113, opened 2021-02-06
fix: enable recovery when publication is dropped

w3b6x9:

Bug fix

After dropping publication, the `realtime` server immediately retries. After creating a new publication, the server is unable to recover.

## What is the new behavior?

After dropping publication, the `realtime` server retries with backoff and jitter. Existing replication slot is dropped. When publication is created, the server now recovers and creates a new replication slot.

## Additional context

This is still WIP as I need to continue testing different edge cases. The way this fix works is:

1. Server recognizes that publication has been dropped. Retries with backoff and jitter, which will prevent spikes in CPU usage.
2. Server drops replication slot. If replication slot fails to be dropped, then server will continue retrying.
3. When server has successfully dropped replication slot, it will check if a new publication has been created.
4. When new publication has been created, then server will create a replication slot and start replication.

### DIFF
--- a/server/lib/adapters/adapter_conn_retry.ex
+++ b/server/lib/adapters/adapter_conn_retry.ex
@@ -15,6 +15,14 @@ defmodule Realtime.Adapters.ConnRetry do
     GenServer.call(__MODULE__, :reset_delay)
   end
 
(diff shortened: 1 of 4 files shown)


w3b6x9:
Closing this draft b/c database conn retrying commits have been reverted.

Issue #95, opened 2020-12-28
Phoenix tokens for channel auth


kamikazechaser:
# Feature request

## Is your feature request related to a problem? Please describe.

At the moment, channels aren't secured. Anyone with the endpoint and public key can listen to **all** db changes. This means that realtime streaming cannot be used at all on the client side. This has been mentioned as a caveat in supabase/supabase#237. However in the main docs, there is no such warning given:
* https://supabase.io/docs/client/subscribe
* https://supabase.io/docs/realtime/client/initializing

## Describe the solution you'd like

Update documentation to reflect that client-side web socket aren't authorized at the moment. Add Phoenix token verification to **all** channels.

The Phoenix token itself [seems to be a JWT](https://elixirforum.com/t/how-is-phoenix-token-different-from-jwt/2349/2) except that the payload is not in JSON. So maybe a bit of decoding/re-encoding the same token in the authorization header from JSON to "Whatever encoding Phoenix uses" might just work. Alternatively, let the burden of creating Phoenix tokens lie with the developer using the same JWT secret.

kiwicopple:
Thanks for this @kamikazechaser - we're checking if we can use our exiting JWT setup first, but if that doesn't work we'll switch to the Phoenix tokens. 

Really appreciate the ideas here 🙏

awalias:
anything else required in the rest of the stack to get this integrated?
- [ ] realtime-js to send Authorization Header?
- [ ] Worker to set SECURE_CHANNELS=true and JWT_SECRET=blah (also maybe SECRET_BASE?)
- [ ] Release new kps build with v0.10.1

awalias:
also how should we manage the breaking change with existing applications? Thinking of the scenario where we upgrade people's KPS instance

emschwartz:
If you wanted to avoid making it a sudden breaking change, you could give
the current public key the ability to subscribe to all channels, create a
new restricted public key that is limited to the row level permissions, and
deprecate or recommend against using the original public key.

>

kiwicopple:
> realtime-js to send Authorization Header?

This one is most urgent, although it is actually in `supabase-js`. 


[...]

Issue #109, opened 2021-01-28
v0.10.x JWT documentation seems unclear?


acupofjose:
## Question

Using the current release of `@supabase/realtime-js` I am unable to connect to the latest release of `@supabase/realtime`. Following the documentation of the breaking changes in v0.10, what are client opening requests supposed to include?

A `token` query param? A `header`?

## To Reproduce

**docker-compose.yml**
```yml
version: "3"
services:
  realtime:
    image: supabase/realtime:latest
    ports:
      - "4000:4000"
    environment:
      DB_HOST: db
      DB_NAME: postgres
      DB_USER: postgres
      DB_PASSWORD: postgres
      DB_PORT: 5432
      PORT: 4000
      HOSTNAME: localhost
      JWT_SECRET: SOMETHING_SUPER_SECRET
      SECURE_CHANNELS: "true"
    depends_on:
      - db
  db:
    image: supabase/postgres
    ports:
      - "5432:5432"
    command:
      - postgres
      - -c
      - wal_level=logical
    environment:
      POSTGRES_PASSWORD: postgres
```

**index.js**
```js
const { RealtimeClient } = require("@supabase/realtime-js")

const client = new RealtimeClient("ws://localhost:4000/socket")
client.onOpen(() => console.log("Socket opened."))
client.onClose(() => console.log("Socket closed."))
client.onError((e) => console.log("Socket error", e.message))
client.connect()
```

**Terminal Output**
```bash
realtime_1  | 2021-01-28 16:01:20.162 [info] Running RealtimeWeb.Endpoint with cowboy 2.6.3 at :::4000 (http)
realtime_1  | 2021-01-28 16:01:20.162 [info] Access RealtimeWeb.Endpoint at http://localhost:4000
realtime_1  | 2021-01-28 16:03:25.018 [info] REFUSED CONNECTION TO RealtimeWeb.UserSocket in 114µs
realtime_1  |   Transport: :websocket
realtime_1  |   Serializer: Phoenix.Socket.V1.JSONSerializer
realtime_1  |   Connect Info: %{}
realtime_1  |   Parameters: %{"vsn" => "1.0.0"}
realtime_1  | 2021-01-28 16:03:26.033 [info] REFUSED CONNECTION TO RealtimeWeb.UserSocket in 100µs
```

**Application Output**
```bash
realtime-test-js % node .
Socket error undefined
Socket closed.
Socket error undefined
Socket closed.
```

## Expected behavior

A connection to the realtime socket.

## System information

[...]

PR #122, opened 2021-02-11
feat: change socket param from token to apikey

w3b6x9:

feature

socket param auth token key is 'token'

## What is the new behavior?

socket param auth token key is 'apikey'

## Additional context

Related issue: #95


soedirgo (approved):
LGTM! I wonder though if we should allow `"token"` for a while, just so things like [this](https://github.com/supabase/realtime/issues/109) don't break?

w3b6x9:
@soedirgo good point! Yea, I was conflicted when I was updating the code b/c of that and I made sure to update the `realtime-js` docs ([ref](https://github.com/supabase/realtime-js/pull/76)).

@kiwicopple how do you normally handle these types of situations? Should we leave `token` in there and add some kind of deprecation warning?

PR #123, opened 2021-02-15
[WIP] Add workflows triggered by realtime events

fracek:
This PR implements the workflows we discussed earlier.

Workflows are defined using [Amazon States Language](https://states-language.net/) and reside in a database controlled by realtime.

Workflows are stored in the database, the table is synchronised with an ETS table. We use realtime events to synchronise the workflows between these two tables.

The workflow trigger (`realtime/workflows/trigger.ex`) is responsible for looking up which workflows match a given event and start the execution. 

Workflows can be `transient` or `persistent`. Transient workflows are executed inside a `Task`, if the realtime server restarts or if they crash they are not restarted. Persistent workflows are stored as Oban jobs, they are persisted across restarts.

### The interpreter

The current interpreter implementation compiles a `gen_statem` at runtime from a quoted module. This is less than ideal because it also means we need to unload the module when we are finished executing the workflow.

We need to implement persistent workflows in such a way that workflows' execution is resumed from the state they were interrupted, this can be done by "breaking" the interpreter in several persistent jobs, each one interpreting one state and transitioning to the next. The challenge here is to deal with `Parallel` and `Map` states, since they run a number of sub-workflows in parallel. My idea is to enqueue a persistent job for each sub-workflow and one job that polls for the jobs completion. This approach has one downside (two if you don't like polling): pruning old jobs also requires checking that the job result is not needed by any other job.

### What's left to do

 - [x] Add a new set of configurations for the workflows database. Users should be able to select the schema.
 - [x] Validate rest requests and return meaningful error responses
 - [x] Define a plugin-like architecture for ASL Resources (e.g. one for aws lambda functions, one for simple http post requests, etc)
 - [x] Ignore changes happening on workflows schema/tables by default? It's very easy to trigger a workflow in response to events generated by another workflow.
  - [ ] Implement an interpreter, possibly re-using parts of the [`states_language`](https://github.com/entropealabs/states_language) library
   - [ ] Implement http post request resources. Workflow metadata should be included in the header (`X-Supabase-Workflow-Id`, `X-SUpabase-Execution-Id`, `X-Supabase-Workflow-Name`, etc). To implement this the interpreter needs to carry around a context object with all this information.
   - [ ] Log state events to a table (optional, if `log_type == :postgres`). These events can be used by users to debug their workflows.

 
I will try to keep this list updated as we progress through it.


server/config/config.exs
```diff
@@ -60,13 +63,27 @@ config :realtime,
...
-  pubsub: [name: Realtime.PubSub, adapter: Phoenix.PubSub.PG2]
+  pubsub_server: Realtime.PubSub,
+  secret_key_base: session_secret_key_base
+
+config :realtime, Realtime.Repo,
+  hostname: db_host,
+  port: db_port,
+  database: db_name,
+  username: db_user,
+  password: db_password
```
kiwicopple, line 81: These settings are currently the same as the realtime listerner here: 
https://github.com/supabase/realtime/blob/af6344c7746e8a8af6a11a9b498721c1f97e339b/server/lib/realtime/application.ex#L24

Even though they will be the same in many cases, it might be useful to split them up into a "workflows" database connection and a "realtime" database connection

The reasoning is that:

- Perhaps Postgres doesn't scale as a store for workflows + logs, so we have to switch to something else
- Perhaps in the future we want to split the "realtime" source into several inputs

image


From our conversation this morning, it seems that you're listening to the changes in Postgres, using the same listener that can trigger the workflows - this separate config could prevent the user from triggering "infiinte" workflows (and we should defintiely namespace everything to a `workflows` schema inside the database)

server/lib/realtime/workflows/manager.ex
```diff
@@ -0,0 +1,153 @@
...
+        id: workflow_data.id,
+        trigger: workflow_data.trigger,
+        definition: workflow_data.definition
+      }
+      :ets.insert(table, {workflow.id, workflow})
+    end)
+    Enum.each(queue, fn txn -> do_handle_notification(table, txn) end)
+    new_state = %State{state | state: :sync, queue: []}
+    {:noreply, new_state}
+  end
```
w3b6x9, line 89: @fracek you can simplify this with `init` returning `{:ok, state, {:continue, :load_workflows}}` and `handle_continue`:

```suggestion
  @impl true
  def init(_config) do
    :ets.new(@table_name, [:named_table, :protected]) # returns @table_name
    {:ok, %{}, {:continue, :load_workflows}}
  end

  @impl true
  def handle_continue(:load_workflows, state) do
    Workflows.list_workflows()
    |> Enum.each(fn workflow_data ->
      workflow = %{
        id: workflow_data.id,
        trigger: workflow_data.trigger,
        definition: workflow_data.definition
      }
      :ets.insert(table, {workflow.id, workflow})
    end)

    {:noreply, state}
  end
```

This way, `handle_continue` ensures that all workflows are loaded before other messages are processed and you don't have to keep track of `state` and `queue`. You don't even need `table` in state as you can just reference `@table_name`.

server/lib/realtime/workflows/manager.ex
```diff
@@ -0,0 +1,153 @@
...
+
+  def start_link(config) do
+    GenServer.start_link(__MODULE__, config, name: __MODULE__)
+  end
+
+  @doc """
+  Send notification for txn to the workflow manager.
+  """
+  def notify(txn) do
+    GenServer.cast(__MODULE__, {:notify, txn})
```
w3b6x9, line 25: @fracek we should consider changing this to a `call`:

```suggestion
    GenServer.call(__MODULE__, {:notify, txn})
```

I'm worried about `GenServer.cast` (`Realtime.Workflows.Manager.notify(txn)`) async message because `Realtime.Workflows.Trigger.notify(txn)` is called directly afterwards and it depends on `Manager` to keep `:ets` table up to date.

server/lib/realtime/workflows/trigger.ex
```diff
@@ -0,0 +1,22 @@
+defmodule Realtime.Workflows.Trigger do
```
w3b6x9, line 1: @fracek perhaps we should consider moving the logic in `Trigger.notify` to `Manager` if its only purpose is to get `workflows_for_change` from `Manager` and then invoking workflows.

server/lib/realtime/workflows/execution_manager.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Realtime.Workflows.ExecutionManager do
+
+  require Logger
+
+  alias Realtime.Workflows.TransientExecutionManager
+  alias Realtime.Workflows.PersistentExecutionManager
+
+  def start_workflow_execution(workflow, execution, opts \\ []) do
```
w3b6x9, line 8: @fracek

> Workflows can be transient or persistent. Transient workflows are executed inside a Task, if the realtime server restarts or if they crash they are not restarted. Persistent workflows are stored as Oban jobs, they are persisted across restarts.

I was just wondering why transient workflows aren't inserted to the Oban queue as well. 

Here's how I thought about it: both transient and persistent workflows are inserted into Oban queue, and `is_persistent` is passed to `args` when inserting to Oban as well. When `Oban.Worker` processes a job, it'll route the args to a transient `Task` or a persistent `:gen_statem` (assuming you're going to use it for the interpreter) based on `args.is_persistent`.

I bring this up b/c if a user kicks off a one off workflow, and `realtime` crashes and restarts, then the user won't really know what happened to the executed workflow. If we alert them that something went wrong then they'll have to manually kick off the same one off workflow? Seems like something we can take care of server-side and make sure it's successfully completed.

However, I might be misunderstanding something, especially since I don't know how you're going to build the interpreter yet. Just thought I throw this out there!

server/test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -10,7 +10,7 @@ defmodule RealtimeWeb.RealtimeChannelTest do
     {:ok, socket: socket}
   end
 
-  test "INSERTS are broadcasts to the client", %{socket: socket} do
+  test "INSERTS are broadcasts to the client", %{socket: _socket} do
```
w3b6x9, line 13: @fracek you can just remove these:

```suggestion
  test "INSERTS are broadcasts to the client" do
```

w3b6x9:
@fracek I left ya a couple of comments. Awesome work on this so far!!! 🔥

server/lib/realtime/workflows/manager.ex
fracek: Great suggestion, I will add this.

server/lib/realtime/workflows/execution_manager.ex
fracek: The idea behind transient workflows is that we don't touch the database, the trade-off is that if the server crashes or is updated the execution is lost. The uses cases we discussed with @kiwicopple were as follows:

* Transient: for example IoT applications or starting the workflow from the rest API. In the first case we don't care about the individual execution since we're probably going to get hundreds more in the next few seconds, in the second case we don't care because the http handler was killed together with the server so we won't be able to send the response back.
* Persistent: anything else. The idea is to use oban to help us build an interpreter that can span multiple restarts. I'm thinking for example to use the scheduling functionality to implement the `Wait` state.

More precisely, we touch the database only once (or twice if we include information about the workflow termination) with transient workflows, while for persistent workflows we touch it once (or twice) + 3 times for each state (1 to insert the job, 1 to change state to executing, 1 to change state to completed).

Hope what I wrote makes sense, if not let me know and I can expand more.

server/lib/realtime/workflows/manager.ex
fracek: Now I wonder if it even needs to be a genserver, we only need to initialise the ets table which can be done once at startup.

w3b6x9: @fracek I think you're right!

fracek:
I updated the first comment. Overall you can use this code to test how workflows work. I need to figure out a good implementation for the persistent workflow interpreter and then it's in a good shape! Event logging is a non-issue and handling resources is already implemented (just need to set the correct headers).

chasers:
Closing. Working on Workflows somewhere else now. To be announced!

PR #129, opened 2021-03-05
feat: enable recovery when publication is dropped

w3b6x9:

feature

When publication is dropped and there's some change in the database then the `realtime` server shuts down.

## What is the new behavior?

When publication is dropped and there's some change in the database then the `realtime` server will:

1. drop the existing replication slot
2. retry with exponential backoff and jitter and check to see if publication exists
3. when publication does exist then new replication slot will be created
4. continue to notify subscribers of database changes


kiwicopple:
wow nice @w3b6x9, glad you're catching all these ones. I didnt even think to test this. 

Thanks!

server/lib/realtime/application.ex
```diff
@@ -60,13 +60,13 @@ defmodule Realtime.Application do
...
-        Realtime.Replication,
+        Realtime.DatabaseReplicationSupervisor,
         # You can provide a different WAL position if desired, or default to
         # allowing Postgres to send you what it thinks you need
-        epgsql: epgsql_params,
-        slot: slot_name,
-        wal_position: {"0", "0"},
-        publications: ["supabase_realtime"]
+        epgsql_params: epgsql_params,
+        publications: ["supabase_realtime"],
```
kiwicopple, line 67: Perhaps this should be passed as a config variable?

Also it looks like this can connect to multiple publications now? If that's correct, this is amazing

w3b6x9: @kiwicopple yep, good idea! I'll add `publications` as a config var.

Unfortunately, realtime still can only accept one publication.

Original code: 

```
publication_names =
      Keyword.get(config, :publications)
      |> Enum.map(fn pub -> ~s("#{pub}") end)
      |> Enum.join(",")
```

Refactored code found in this PR:

```
publications
           |> Enum.intersperse(",")
           |> IO.iodata_to_binary()
           |> String.replace("'", "\\'")
```

I only refactored the code as accepting multiple publications is a bit more complicated and I didn't want to pile that on while there were already a lot of moving parts with the PR. However, I can certainly enable actual multiple publications in the future if there's a demand for it! Just let me know!

kiwicopple: Great - one publication per server is enough for now 😅 

In the future it could be nice to have different publications (eg, Private tables on one publication, public tables on another), but that's getting ahead of ourselves.

Thanks Wenbo!

kiwicopple (approved):

Issue #132, opened 2021-04-04
Suggestion: document query abilities in README


rattrayalex:
# Improve documentation

## Link 

The readme

## Describe the problem

As someone evaluating Supabase, I am curious how powerful the realtime capabilities are (and/or will be). A major challenge in realtime subscriptions in general is getting quick responses to complex queries. If you only want updates in very specific situations, relying on unpowerful queries can result in a large number of clients receiving a large number of irrelevant updates, which may cause production problems and high costs.

Hasura, one provider I might compare when considering Supabase, allows arbitrarily complex queries and re-runs them once per second to "implement" realtime subscriptions. This approach has obvious tradeoffs/problems/limitations.

I'd like to know how Supabase/Realtime compares – can I only query on row ID? Can I query based on equality only, or what operations are supported? Based on a single field only, or can I combine OR's and AND's? Can I only query on fields that are indexed? How are the filters implemented?

rattrayalex:
Ah, this appears to be [documented in the supabase docs](https://supabase.io/docs/reference/javascript/subscribe#listening-to-row-level-changes): 

> You can listen to individual rows using the format `{table}:{col}=eq.{val}` - where `{col}` is the column name, and `{val}` is the value which you want to match.

kiwicopple:
Hi @rattrayalex - I'll leave this open if you want to contribute a change to the Readme?

Otherwise let's close this :)

rattrayalex:
Sure! Put up https://github.com/supabase/realtime/pull/133 but feel free to edit / suggest edits liberally!

martineboh:
All examples currently listens only to a single column. Any example on listening with two or more columns like `{table}:{col}=eq.{val}:{col2)=eq.{val2}`. A use case is listening for comments that has columns is_deleted=true and author=me

rattrayalex:
Ah, does it work with multiple columns @kiwicopple ?

kiwicopple:
it will only work for a single column. We'll have to figure out more complicated strategies (probably as part of our workflow engine)

w3b6x9:
@rattrayalex thanks for creating this issue and putting up PR #133!

> I'd like to know how Supabase/Realtime compares – can I only query on row ID? Can I query based on equality only, or what operations are supported? Based on a single field only, or can I combine OR's and AND's? Can I only query on fields that are indexed? How are the filters implemented?

[...]

PR #133, opened 2021-04-05
Docs update: Document subscription topic limitations

rattrayalex:
Addresses https://github.com/supabase/realtime/issues/132

I'm not sure if other limitations should be documented as well, but doing so can help users understand how their application architecture must map to the tool and decide whether it is right for them. Another example might be scalability limitations, though my guess is those are both high and highly dependent on the application/context.


w3b6x9:
@rattrayalex thanks!

Related PR: #148

PR #135, opened 2021-04-19
fix: Add docker image for arm64 architectures

samtgarson:

Fixes https://github.com/supabase/cli/issues/23

There is no docker build shipped for ARM architectures, so the realtime container fails when running the CLI on M1 Macs.

## What is the new behavior?

Building a docker images locally fixes the issue, this _should_ publish a compatible image for ARM architectures to Docker Hub.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -194,6 +194,15 @@ jobs:
       - uses: actions/checkout@v2
         if: env.VERSION_BEFORE != env.VERSION_AFTER
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@master
+        with:
+          platforms: arm64,amd64
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@master
+
       - name: Upload build to Docker Hub
         if: env.VERSION_BEFORE != env.VERSION_AFTER
         uses: docker/build-push-action@v1
@@ -202,3 +211,5 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
           repository: supabase/realtime
           tags: latest,v${{ env.VERSION_AFTER }}
+          builder: ${{ steps.buildx.outputs.name }}
+          platforms: linux/amd64,linux/arm64


kiwicopple:
wow that was fast @samtgarson. 

There is a strange CI error coming through on GH Actions:

```
Error: 9.708 [error] Unable to load crypto library. Failed with error:
":load_failed, Failed to load NIF library /home/runner/work/_temp/.setup-elixir/otp/lib/crypto-4.4.2/priv/lib/crypto: 'libcrypto.so.1.0.0: cannot open shared object file: No such file or directory'"
OpenSSL might not be installed on this system.
```

I don't think it's related to this PR because you edited the `release.yml` . Will need @soedirgo or @w3b6x9 to confirm

w3b6x9:
@samtgarson thanks for the PR!

@kiwicopple confirming that

```
Error: 9.708 [error] Unable to load crypto library. Failed with error:
":load_failed, Failed to load NIF library /home/runner/work/_temp/.setup-elixir/otp/lib/crypto-4.4.2/priv/lib/crypto: 'libcrypto.so.1.0.0: cannot open shared object file: No such file or directory'"
OpenSSL might not be installed on this system.
```

is not introduced by this PR.

Saw the above error in the most recent commit https://github.com/supabase/realtime/commit/26a34c8e135564d544ecb79f5e983d661d893ace on `master` and here's its failed check https://github.com/supabase/realtime/runs/2238618027.

w3b6x9:
Note: I've created PR https://github.com/supabase/realtime/pull/136 to address the "Unable to load crypto library" error.

.github/workflows/release.yml
```diff
@@ -202,3 +211,5 @@ jobs:
           password: ${{ secrets.DOCKER_PASSWORD }}
           repository: supabase/realtime
           tags: latest,v${{ env.VERSION_AFTER }}
```
kiwicopple, line 213: Looks like it didn't push a new tag - perhaps we need 2 `Upload build to Docker Hub` steps, each with different architectures/tags?

soedirgo: I think it's because it's using `docker/build-push-action@v1` instead of `v2`. Will address it in another PR.

w3b6x9: @kiwicopple looks like it's fine to define multiple platforms in a single step (see: https://github.com/docker/build-push-action/blob/master/docs/advanced/multi-platform.md).

@samtgarson i think we'll have to use `v2` of  `docker/build-push-action` to get this to work.

.github/workflows/release.yml
```diff
@@ -194,6 +194,15 @@ jobs:
...
+        with:
+          platforms: arm64,amd64
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@master
+
       - name: Upload build to Docker Hub
         if: env.VERSION_BEFORE != env.VERSION_AFTER
         uses: docker/build-push-action@v1
```
w3b6x9, line 208: i think this is the change we'll have to make: `docker/build-push-action@v1` -> `docker/build-push-action@v2`

.github/workflows/release.yml
w3b6x9: update: looks like @soedirgo beat me to it! See https://github.com/supabase/realtime/commit/eed8a4dbec310069f3eaad68319d70f0b8e3dcd6#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R217

samtgarson:
Thanks @w3b6x9 @soedirgo @kiwicopple 🙌 🙌 

I was on UK time so missed all this but good catch 😂 I know you're busy, thanks for getting this in—works a charm with the new image.

PR #139, opened 2021-04-29
feat: enable IP version specification for DB connect

w3b6x9:

feature

User can't pass in the IP version they want to use to connect to db.

## What is the new behavior?

User can pass in either IPv4 or IPv6 in order to connect to db.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,9 +132,10 @@ DB_NAME                 # {string}      Postgres database name
 DB_USER                 # {string}      Database user
 DB_PASSWORD             # {string}      Database password
 DB_PORT                 # {number}      Database port
+DB_IP_VERSION           # {string}      (options: 'IPv4'/'IPv6') Connect to database via either IPv4 or IPv6. Disregarded if database host is an IP address (e.g. '127.0.0.1') and recommended if database host is a name (e.g. 'db.abcd.supabase.co') to prevent potential non-existent domain (NXDOMAIN) errors.
 SLOT_NAME               # {string}      A unique name for Postgres to track where this server has "listened until". If the server dies, it can pick up from the last position. This should be lowercase.
 PORT                    # {number}      Port which you can connect your client/listeners
-SECURE_CHANNELS         # {string}     (options: 'true' or 'false') Enable/Disable channels authorization via JWT verification.
+SECURE_CHANNELS         # {string}      (options: 'true'/'false') Enable/Disable channels authorization via JWT verification.
 JWT_SECRET              # {string}      HS algorithm octet key (e.g. "95x0oR8jq9unl9pOIx"). Only required if SECURE_CHANNELS is set to true.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -132,9 +132,10 @@ DB_NAME                 # {string}      Postgres database name
 DB_USER                 # {string}      Database user
 DB_PASSWORD             # {string}      Database password
 DB_PORT                 # {number}      Database port
+DB_IP_VERSION           # {string}      (options: 'IPv4'/'IPv6') Connect to database via either IPv4 or IPv6. Suggested if database host is an IP address and recommended if database host is a name (string) to prevent potential non-existent domain (NXDOMAIN) errors.
```
kiwicopple, line 135: I see that this matches up with the epsql library but it seems like an odd way to do it, as you have an ENV VAR declaring the "type" of another ENV VAR. I'm not against this, but wanted to get your thoughts @w3b6x9 :

If I was building from scratch I would have 3 Options:

- `DB_IP_V6`
- `DB_IP_V4`
- `DB_HOST` (dns/cname)

(or something similar) 

I would make at least one of these required. If multiple are added, I would use them in that order (v6, v4, hostname).

w3b6x9: > ENV VAR declaring the "type" of another ENV VAR

@kiwicopple good point!

> If I was building from scratch I would have 3 Options:
> 
> - DB_IP_V6
> - DB_IP_V4
> - DB_HOST (dns/cname)

@kiwicopple Fair proposal!

For our use case, when the host is `db.abcdefabcdefabcdefab.supabase.co`, I assume we'll still be passing that via `DB_HOST`. Are we also going to pass in `db.abcdefabcdefabcdefab.supabase.co` via `DB_IP_V4` just so we can tell Realtime to connect via IPv4? Or are we exclusively passing in `db.abcdefabcdefabcdefab.supabase.co` via `DB_IP_V4` only?

I think this can get confusing then because you can pass in `db.abcdefabcdefabcdefab.supabase.co` via `DB_IP_V4` so Realtime will know to connect via IPv4 while technically `DB_IP_V4` is meant for an IPv4-compatible address and not a dns/cname. Or do I have that wrong and either ip address or dns/cname can work for any of the three options?

w3b6x9:
@kiwicopple I think we should stay the course and I've also added this new [commit](https://github.com/supabase/realtime/pull/139/commits/eb1485219a30d490e796cd9ad59ef262250766a0). Let me know what you think!

README.md
w3b6x9: @kiwicopple I think we should stay the course and I've also added this new [commit](https://github.com/supabase/realtime/pull/139/commits/eb1485219a30d490e796cd9ad59ef262250766a0). Let me know what you think!

kiwicopple: Got it! Good one @w3b6x9 👍

PR #141, opened 2021-05-01
copy pasta fix

dhenson02:

Example/docs minor typo fix in a log

Says "UPDATE" when listening to "DELETE"

## What is the new behavior?

Says "DELETE" when listening to "DELETE"

## Additional context

Node server example file


w3b6x9:
@dhenson02 good catch, thanks!

PR #142, opened 2021-05-01
remove unnecessary setInterval()

dhenson02:

Example/docs cleanup

Runs empty interval timer every 10 seconds

## What is the new behavior?

Not to do that

The listeners added throughout the file take care of this [via the Event Loop](https://nodejs.org/en/docs/guides/event-loop-timers-and-nexttick/#phases-overview).  If you are seeing otherwise, your connections are probably closing unexpectedly and require monitoring + logging out of the scope of a simple example server.

## Additional context

Node server example file


kiwicopple:
Not entirely sure why that was in there tbh - I think for testing?

Let's remove it - if it really is a problem then we can fix it properly

Issue #125, opened 2021-02-16
RAM explosion on huge # of inserts


awalias:
I am now killing the realtime service when it chews up to 70% on RAM on an instance,

The process restarts but gets caught in a loop of filling up memory and getting killed (with a 3 second restart wait each time), it would be nice if the application could constrain itself and slowly work through the backlog of events (given it has ~670Mb of RAM to play with on the free tier)

I can reliably reproduce this by [inserting 1 million (small) rows of (id, slug) at once using COPY](https://raw.githubusercontent.com/supabase/benchmarks/master/db/migrations/20201124141199_read_data.sql) (make sure to **not** drop the publication first if you're testing this)

an alternative would be to manage this with swap space on the instance but it would be much nicer to handle this in the application itself

I have a live instance which is currently caught in this boot loop so dm me for access (no idea if it will clear itself after some time)

abc3:
Memory issue can reproduce on the working machine `INSERT INTO users (name) SELECT * FROM generate_series(1, 100000);`
It seems the GC doesn't want itself to clean the heap, only can be done after the manual call `:erlang.garbage_collect(pid)`. It may be because `txn` are passed synchronously to all functions, this influences throughput too.

<img width="676" alt="Screenshot 2021-02-16 at 13 34 41" src="https://user-images.githubusercontent.com/1172600/108057762-ba9b3200-705b-11eb-846a-c47486aadf0b.png">

I will try later to check async callings.

fracek:
Agree with @abc3. If switching to async doesn't work, we may need to add backpressure between replication and the rest of the system.

abc3:
Sync calls don't fault. Looks like messages from DB coming too fast and passing to WebSockets too slow, so they accumulate in VM and eat all free RAM.

fracek:
So the only way to solve this is slow down reading from the DB if the changes start accumulating downstream (e.g. WebSockets?)

abc3:
During slow down, the messages will accumulate in epgsql process.
Maybe it can be done with logging in the DETS, then reading with some defined RPS for each export (websocket/connectors).

I think it will be useful for RAM optimization to add a memory handler to managing signals like system_memory_high_watermark, send inactive processes with hard work (Replication/SubscribersNotification) to hibernate mode.

w3b6x9:
[...]

PR #147, opened 2021-05-05
[WIP] Store DB changes on disc

abc3:

This PR allows saving memory on large transactions using the disc as storage for DB changes.
Related to that #125 

All transaction events stored in `Realtime.Replication` process, which means in RAM.

## What is the new behavior?

Each decoded db message stored in a file

## Additional context

I've implemented a feature for storing replication changes on the disc and read them as Elixir's Stream. Using stream allows easy integration into the current architecture.

Tell me what you think about that approach? Is it need to continue work on that?


kiwicopple:
Conceptually this seems like a massive win. One question-
if the server is rebooted while processing a file, will it pick up where it was before rebooting? (that would be amazing for CDC).

abc3:
Not yet. Currently, I am focused on memory saving and soft integration. But it can be easily done by storing the position of the last success sent record, for each exporter (ws/webhook). 

The backlog file format is sequences of two fields - payload size(32 bits) and payload, which is a binary representation of decoded db message through `term_to_binary`. Every new transaction creates a new backlog file.

abc3:
changed to #152

PR #154, opened 2021-05-21
fix: drop replication slot when db deletes wal segment

w3b6x9:

Bug fix

Database deletes a WAL segment that Realtime server still needs to replay. Realtime is caught restarting in a loop as it cannot find the WAL segment.

## What is the new behavior?

Database deletes a WAL segment that Realtime server still needs to replay. Realtime recognizes this, drops the replication slot, and on restart a new replication slot is created. All is normal again.

## Additional context

- Re-creating a standby (Realtime server) that falls behind is a suggested solution (see https://www.2ndquadrant.com/en/blog/postgresql-9-4-slots for more details)

- There's no good way to test this live so we'll just have to see if it resolves the issue after deploying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ A few reasons:
 2. Decoupling. For example, if you want to send a new slack message every time someone makes a new purchase you might build that functionality directly into your API. This allows you to decouple your async functionality from your API.
 3. This is built with Phoenix, an [extremely scalable Elixir framework](https://www.phoenixframework.org/blog/the-road-to-2-million-websocket-connections).
 
+### Does this server guarentee delivery of every data change?
+
+Not yet! Due to the following limitations:
+
+1. Postgres database runs out of disk space due to Write-Ahead Logging (WAL) buildup, which can crash the database and prevent Realtime server from streaming replication and broadcasting changes.
+2. Realtime server can crash due to a larger replication lag than available memory, forcing the creation of a new replication slot and resetting streaming replication to read from the latest WAL data.
+3. When Realtime server falls too far behind for any reason, for example disconnecting from database as WAL continues to build up, then database can delete WAL segments the server still needs to read from, for example after reconnecting.
 
 ## Quick start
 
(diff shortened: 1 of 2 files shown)


kiwicopple:
Yeah I agree with this approach for our current use-case. If we ever want to use this as an auditing tool however, we'll have to re-engineer this. Perhaps we can add this to the FAQ's in the readme. Something like:

- q: does this provide guaranteed delivery?
- a: not yet, due to XXX.

I think as part of our re-engineering or realtime/sockets/workflows we can come up with some solutions here

w3b6x9:
@kangmingtay

> If we ever want to use this as an auditing tool however, we'll have to re-engineer this.
> I think as part of our re-engineering or realtime/sockets/workflows we can come up with some solutions here

Yep, totally agree!

> add this to the FAQ's in the readme

Yep, good idea! Let me go add to readme and i'll ping you for feedback

README.md
```diff
@@ -62,6 +62,13 @@ A few reasons:
 2. Decoupling. For example, if you want to send a new slack message every time someone makes a new purchase you might build that functionality directly into your API. This allows you to decouple your async functionality from your API.
 3. This is built with Phoenix, an [extremely scalable Elixir framework](https://www.phoenixframework.org/blog/the-road-to-2-million-websocket-connections).
 
+### Does this server guarentee delivery of every data change?
+
+Not yet! Due to the following limitations:
+
+1. Postgres database runs out of disk space due to Write-Ahead Logging (WAL) buildup, which will crash the database and prevent Realtime server from streaming replication and broadcasting changes.
+2. Realtime server will crash due to a larger replication lag than available memory, forcing the creation of a new replication slot and resetting streaming replication to read from the latest WAL data.
+3. When Realtime server falls too far behind for any reason, for example disconnecting from database as WAL continues to build up, then database will delete WAL segments the server still needs to read from, for example after reconnecting.
```
kiwicopple, line 71: Looks great @w3b6x9 - I would just update some of the "will"s to "can"s. eg:

> which can crash the database

kiwicopple (approved):

README.md
w3b6x9: @kiwicopple yep, good call

kiwicopple: lgtm 👍 feel free to merge

PR #157, opened 2021-05-31
fix: drop replication slot when lag exceeds max replication lag value

w3b6x9:

Prod improvement

Running on a production instance, a big replication lag will use a lot of CPU and continuously crash Realtime due to instance memory limits.

## What is the new behavior?

Running on a production instance, when a big replication lag exceeds a set limit (`MAX_REPLICATION_SLOT_MB`), then replication slot will be dropped, Realtime will be restarted, and a new slot will be created. This will allow the instance to keep operating as normal.

## Additional context

**WARNING**: Setting `MAX_REPLICATION_SLOT_MB` could cause database changes to be lost when the replication slot is dropped.


kiwicopple:
This looks good @w3b6x9. we should also put this warning in the README yeah ?

w3b6x9:
> This looks good @w3b6x9. we should also put this warning in the README yeah ?

good idea! https://github.com/supabase/realtime/pull/157/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R201

PR #161, opened 2021-06-09
Feat: workflows v1

kiwicopple:
Picking up on Fran's work. The goal is to:

- validate all the code is working with the new changes
- create a simple transient workflow which will send to an endpoint
- build a simple UI which we can use in our Dashboard (will make this a Next.js app)

potential payload:

`POST http://localhost:4000/api/workflows/`
```json
{
    "name": "send HTTP req",
    "trigger": "public:users",
    "default_execution_type": "transient",
    "definition": {
      "StartAt": "StartHere",
      "States": {
        "StartHere": {
          "Type": "Task",
          "Resource": "http",
          "Parameters": {
              "url": "https://webhook.site/4666291b-d063-45f6-a32f-f2d10bdc548c",
              "method": "GET"
          },
          "End": true
        }
      }
    }
  }
```

Development:

```bash
make local:db # start database on localhost port 5432

cd server
mix deps.get
mix ecto.migrate # run the workflow migrations
mix do event_store.create, event_store.init # init the event store: https://github.com/commanded/eventstore/blob/master/guides/Getting%20Started.md
mix phx.server # will now be running on port 4000
```


server/lib/realtime/resource/http.ex
```diff
@@ -0,0 +1,33 @@
...
+  alias Realtime.Resource
+
+  @behaviour Resource
+
+  def can_handle(resource) do
+    resource == "http"
+  end
+
+  def handle(resource, _ctx, args) do
+    Logger.info("Calling http resource #{inspect resource} with args #{inspect args}")
```
kiwicopple, line 29: TODO: create this implementation

server/lib/realtime/workflows.ex
```diff
@@ -0,0 +1,238 @@
...
+
+    txn_as_map = Map.from_struct(txn)
+
+    attrs = %{
+      arguments: txn_as_map,
+    }
+
+    Enum.each(workflows, fn workflow ->
+      with {:ok, %{execution: execution, revision: revision}} <- create_workflow_execution(workflow.id, attrs) do
+	# TODO: should be based on default execution type
```
kiwicopple, line 189: Todo: we should which type of workflow this is (it's always running as a persistent workflow right now

server/lib/realtime/interpreter/transient.ex
```diff
@@ -0,0 +1,137 @@
...
+
+    Process.send_after(self(), {:finish_waiting, event}, duration)
+    :ok
+  end
+
+  defp execute_event_side_effect(%Event.TaskStarted{} = event) do
+    server_pid = self()
+
+    Task.start(fn ->
+      # TODO: execute task
```
kiwicopple, line 109: Todo: add a handle_event function for executing the resource

server/lib/realtime/resource/http.ex
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  # TODO: make this better
+  def handle(resource, ctx, args) do
+    Logger.info("Calling http resource #{inspect(resource)} with args #{inspect(args)}")
+
+    headers = [{"Content-Type", "application/json"}]
+    %{"method" => _method, "url" => url} = args
+    HTTPoison.get(url, headers)
+  end
```
kiwicopple, line 35: Super basic implementation here

server/lib/realtime/interpreter/transient.ex
```diff
@@ -0,0 +1,172 @@
...
+
+    if is_pid(state.reply_to) do
+      send(state.reply_to, {:succeed, result})
+    end
+
+    {:stop, :normal, state}
+  end
+
+  defp handle_event(%Event.TaskStarted{} = event, execution_id) do
+    # TODO: Do we need Oban backing for transient workflows? Might be good for tracking failures
```
kiwicopple, line 155: @fracek - need to check with you on the implementation you were thinking for transient workflows. Does it require Oban?

fracek: I think it's fine (maybe event better) without Oban. I was thinking to use Oban to gain observability of workflows executions, but I think logging the errors to AWS is a better choice in general.

kiwicopple: awesome thanks. I'm thinking for this early version we should try to touch the database as little as possible. once we know it's scaling we can add in a bit more observability.

kiwicopple:
This latest change starts a basic UI for us to play with State Machines. I will generate a few and seed the database so that I can hand it off to the frontend team


https://user-images.githubusercontent.com/10214025/121700066-1b1a1780-cb02-11eb-9d61-5321f3f626a7.mov

kiwicopple:
Possible workflows to use:

```json
{
    "name": "Send HTTP req",
    "trigger": "public:users",
    "default_execution_type": "transient",
    "definition": {
        "StartAt": "TriggerWebhook",
        "States": {
            "TriggerWebhook": {
                "Type": "Task",
                "Resource": "http",
                "Parameters": {
                    "url": "https://webhook.site/4666291b-d063-45f6-a32f-f2d10bdc548c",
                    "method": "POST",
                },
                "Next": "Complete"
            },
            "Complete": {
                "Type": "Succeed"
            }
        }
    }
}
```

With Map: 
```json
{
  "name": "Map through changes",
  "trigger": "public:users",
  "default_execution_type": "transient",
  "definition": {
    "StartAt": "MapOverChanges",
    "States": {
      "MapOverChanges": {
        "Type": "Map",
        "End": true,
        "InputPath": "$.changes",
        "Iterator": {
          "StartAt": "TriggerWebhook",
          "States": {
            "TriggerWebhook": {
              "Type": "Task",
              "Next": "Complete",
              "Resource": "http",
              "Parameters": {
                "url": "https://webhook.site/4666291b-d063-45f6-a32f-f2d10bdc548c",
                "method": "POST",
                "payload": {
                    "id": "$.id",
                    "name": "$.name"
                }
              }
            },
            "Complete": {
              "Type": "Succeed"
              
            }
          }
        }
      }
    }
  }
}

```

w3b6x9:
Working `Map` workflow:

```
{
  "name": "Map through changes",
  "trigger": "public:users",
  "default_execution_type": "transient",
  "definition": {
    "StartAt": "MapOverChanges",
    "States": {
      "MapOverChanges": {
        "Type": "Map",
        "End": true,
        "InputPath": "$.changes",
        "Iterator": {
          "StartAt": "TriggerWebhook",
          "States": {
            "TriggerWebhook": {
              "Type": "Task",
              "Next": "Complete",
              "Resource": "http",
              "Parameters": {
                "url": "https://webhook.site/421b6551-c0df-4302-9967-24150fb224a1",
                "method": "POST",
                "payload": {
                  "id.$": "$.record.id",
                  "name.$": "$.record.name"
                }
              }
            },
            "Complete": {
              "Type": "Succeed"
            }
          }
        }
      }
    }
  }
}
```

server/lib/realtime/resource/http.ex
```diff
@@ -0,0 +1,36 @@
...
+   * `url` - The request URL
+   * `method` - The request HTTP method
+   * `headers` - A map of key-values that will be added to the request headers
+   * `body` - The request body
+
+  ## Result
+
+   * `headers` - The response headers
+   * `status_code` - The response status code
+   * `body` - The response body
```
thesayyn, line 16: I know this is very very rough implementation yet but I wanted to point out some things that might be useful 

 - response body can contain non-serializable data. 
 - there might not be any data at all
 - data can be so big that leading to memory overflow 
 
 these cases could prevent the response body to be passed as input to the next state.

w3b6x9: thanks! will definitely keep these things in mind as we continue to develop this resource.

thesayyn: I am sorry I know that this PR is not a discussion place but I really want to get involved in the development.

I really wonder, how these flows will be executed. (as I can see there I there is a new thread for every new execution) and what will happen when there is an enormous back pressuring queue of tasks? 

I don’t think having many threads will solve the problem because after the limits of the host machine there will be no space to execute any more workflows.

there will be a need for horizontal scaling after some time imho.

w3b6x9: > really want to get involved in the development

appreciate that you want to get involved!

> I really wonder, how these flows will be executed.

I think for the initial version we just want to make sure it works with a couple of resources like webhooks and email so all workflows will be transient. I think you're exactly right when you identified a new process for every execution and an enormous queue of tasks on the host machine causing problems down the line. We've discussed using Postgres-backed queues (Oban) and a pool of processes for workflow executions but we'll have to investigate how well that scales and explore viable alternatives.

chasers:
Closing. Working on Workflows somewhere else now. To be announced!

PR #168, opened 2021-07-28
feat: add Prometheus metrics for active ws conn and topic sub totals

w3b6x9:

feature

## What is the new behavior?

Introduce Prometheus metrics exposed on `/metrics` endpoint. Currently reports:

- total number of active websocket connections.
- total number of active topic subscriptions.


server/lib/realtime/metrics/prom_ex.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Realtime.Metrics.PromEx do
+  use PromEx, otp_app: :realtime
+
+  @impl true
+  def plugins do
+    [Realtime.Metrics.PromEx.Plugins.Realtime]
```
abc3, line 6: It can be useful for debugging and optimizing adding metrics for VM `Plugins.Beam`

server/lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -3,6 +3,7 @@ defmodule RealtimeWeb.RealtimeChannel do
   require Logger, warn: false
 
   def join("realtime:" <> _topic, _payload, socket) do
+    send(self(), :after_join)
```
abc3, line 6: there is no need to make another message send, you can call the `track_channel` here

server/lib/realtime_web/channels/user_socket.ex
```diff
@@ -3,6 +3,14 @@ defmodule RealtimeWeb.UserSocket do
 
   alias RealtimeWeb.ChannelsAuthorization
 
+  defoverridable init: 1
+
+  def init(state) do
+    res = {:ok, {_, socket}} = super(state)
+    Realtime.Metrics.SocketMonitor.track_socket(socket)
```
abc3, line 10: Maybe better track socket after `authorize_conn` on 32 line?

server/lib/realtime/metrics/socket_monitor.ex
```diff
@@ -0,0 +1,81 @@
...
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  def init(_) do
+    {:ok, %{transport_pids: %{}, channel_pids: %{}}}
+  end
+
+  def track_socket(socket = %Phoenix.Socket{}) do
+    GenServer.call(__MODULE__, {:track_socket, socket})
```
abc3, line 17: I think it's better to use non-blocking callings `(handle_cast/handle_info)` because metrics can slow down accepting new sessions, the same for `track_channel`

abc3 (commented):
Looks good for me, but with few warnings :smile:

abc3 (approved):

server/lib/realtime_web/channels/realtime_channel.ex
w3b6x9: yep, good call!

server/lib/realtime/metrics/socket_monitor.ex
w3b6x9: yes, agreed!

server/lib/realtime/metrics/prom_ex.ex
w3b6x9: sounds good, i'll add it in

PR #175, opened 2021-09-09
feat: Add headers to the Webhook request

abc3:

There are use cases where it needs to add custom headers to Webhook, for example, for authentication. This code implements it through the environment variable `WEBHOOK_HEADERS`. Colon `:` used for split multiple headers.

The `key1=value1:key2=value2` will transform to `[{"key1", "value1"},{"key2", "value2"}]`

In the current behavior used the only one additional header `[{"content-type", "application/json"}]`

## What is the new behavior?
In new behavior, there is checking the value of `WEBHOOK_HEADERS` variable, and if it's not empty and valid, then it adds to the webhook request. 
If the application has started with `WEBHOOK_HEADERS=key1=value1:key2=value2`, the request on the server part looks like this:
Screenshot 2021-09-09 at 18 54 23

### DIFF
--- a/server/config/config.exs
+++ b/server/config/config.exs
@@ -58,6 +58,8 @@ db_ip_version =
   %{"ipv4" => :inet, "ipv6" => :inet6}
   |> Map.fetch(System.get_env("DB_IP_VERSION", "") |> String.downcase())
 
+webhook_headers = System.get_env("WEBHOOK_HEADERS")
+
 config :realtime,
   app_port: app_port,
   db_host: db_host,
@@ -73,7 +75,9 @@ config :realtime,
   secure_channels: secure_channels,
   jwt_secret: jwt_secret,
   jwt_claim_validators: jwt_claim_validators,
-  max_replication_lag_in_mb: max_replication_lag_in_mb
+  max_replication_lag_in_mb: max_replication_lag_in_mb,
+  webhook_default_headers: [{"Content-Type", "application/json"}],
+  webhook_headers: webhook_headers
 
 # Configures the endpoint
 config :realtime, RealtimeWeb.Endpoint,
--- a/server/config/config.exs
+++ b/server/config/config.exs
@@ -76,7 +76,7 @@ config :realtime,
   jwt_secret: jwt_secret,
   jwt_claim_validators: jwt_claim_validators,
   max_replication_lag_in_mb: max_replication_lag_in_mb,
-  webhook_default_headers: [{"Content-Type", "application/json"}],
+  webhook_default_headers: %{"content-type" => "application/json"},
   webhook_headers: webhook_headers
 
 # Configures the endpoint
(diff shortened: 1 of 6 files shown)


server/lib/realtime/helpers.ex
```diff
@@ -0,0 +1,25 @@
...
+      )
+      |> Enum.concat(def_list)
+
+    try do
+      # check if keywords is valid
+      Keyword.values(keywords)
+      {:ok, keywords}
+    rescue
+      _ -> :error
+    end
```
w3b6x9, line 21: @abc3 if `def_list` is always a keyword list of `[{"Content-Type", "application/json"}]`, since we're hardcoding it, is this check necessary?

```elixir
String.split(env_val, ":")
|> Enum.map(
  &(String.split(&1, "=")
    |> List.to_tuple())
)
```

will always return a keyword list. But if you think it is then I suggest `Keyword.keyword?/1`.

server/lib/realtime/helpers.ex
```diff
@@ -0,0 +1,25 @@
...
+  def env_kv_to_list("", _), do: :error
+
+  def env_kv_to_list(env_val, def_list) when is_binary(env_val) do
+    keywords =
+      String.split(env_val, ":")
+      |> Enum.map(
+        &(String.split(&1, "=")
+          |> List.to_tuple())
+      )
+      |> Enum.concat(def_list)
```
w3b6x9, line 13: @abc3 i think we should replace existing `"Content-Type"` with `"application/json"` instead of concat-ing  in case a user accidentally passes in another content type since the post body is always json in this post req: `HTTPoison.post(endpoint, serialized_txn, headers)`.

w3b6x9 (commented):
@abc3 looks good! just had a few comments for you.

server/lib/realtime/helpers.ex
abc3: Yes, I agree, replace is better

server/lib/realtime/helpers.ex
abc3: Not exactly, 
the string "key1=value1key2=value2" (missing separator ":" between value1 and key2) return `[{"key1", "value1key2", "value2"}]` which is not two-element tuple.

Anyway, I've already rewritten this function. So, check the update, please.

w3b6x9 (approved):
@abc3 nice one!

server/config/config.exs
```diff
@@ -58,6 +58,8 @@ db_ip_version =
   %{"ipv4" => :inet, "ipv6" => :inet6}
   |> Map.fetch(System.get_env("DB_IP_VERSION", "") |> String.downcase())
 
+webhook_headers = System.get_env("WEBHOOK_HEADERS")
```
darora, line 61: Can we start using a config file (yaml or toml) to supply our config options instead of env vars?
That'll also avoid things like this blowing up if a header has a colon (unusual, but not unheard of)

kiwicopple (approved):

PR #178, opened 2021-09-29
chore: add application_name to differentiate connections on the DB

darora:
Based on https://github.com/epgsql/epgsql#connect , but untested


w3b6x9:
@darora thanks, will confirm and then merge this in.

w3b6x9 (approved):

PR #195, opened 2021-11-23
fix: handle realtime channel join when sub and email are absent

w3b6x9:

Bug fix

## What is the new behavior?

Enable Realtime channel join when `user_token` JWT does not contain sub and email data.

## Additional context

This is possible when `supabase-js` attempts to subscribe to a channel and passes in anonymous key to `user_token` (see [ref](https://github.com/supabase/supabase-js/blob/134105f5f6c966f46fe77213a80de3e28e75441f/src/SupabaseClient.ts#L161) and [ref](https://github.com/supabase/supabase-js/blob/52d417c05aa84a9b9c7b769a2c40a6bcd4f07603/src/lib/SupabaseRealtimeClient.ts#L15)).


abc3 (approved):
looks good to me

PR #202, opened 2021-12-03
fix: ensure primary key compatibility with all types and improve filtering

w3b6x9:

Bug fixes

- can't listen to tables if primary key is non-numeric type
- cannot filter if column is numeric type

## What is the new behavior?

- can listen to all tables with primary key
- can filter column no matter the type


abc3 (approved):
lgtm

PR #203, opened 2021-12-08
fix: handle filters when column is any type

w3b6x9:

Bug fix

Column types, such as `json`/`jsonb`, are not properly handled by Realtime filters, resulting in raised errors.

## What is the new behavior?

Column types, such as `json`/`jsonb`, are properly handled.

## Additional context

When subscribing to a `json` column make sure to strip whitespace.

For example, if a `json` column named `details` on `public.test` table has a new insert:

`[{"a": 1}]`

then changes will be received when listening to inserts on topic `realtime:public:details=eq.[{"a":1}]`.

### DIFF
--- a/server/lib/realtime/subscribers_notification.ex
+++ b/server/lib/realtime/subscribers_notification.ex
@@ -65,10 +65,11 @@ defmodule Realtime.SubscribersNotification do
 
               is_map(record) &&
                 Enum.each(record, fn {k, v} ->
-                  should_notify_column = has_column(event_config, schema, table, k)
-
-                  if is_valid_notification_key(v) and should_notify_column do
-                    "#{table_topic}:#{k}=eq.#{v}"
+                  with true <- is_notification_key_valid(v),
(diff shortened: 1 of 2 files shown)


abc3 (commented):
all good and works well, but could we replace all `case` with sth like code below? Because the difference is only in the record key

```elixir
if type in ["INSERT", "UPDATE", "DELETE"] do
  record_key = if type == "DELETE", do: :old_record, else: :record

  record = Map.get(change, record_key)

  is_map(record) &&
    Enum.each(record, fn {k, v} ->
      with true <- is_notification_key_valid(v),
            {:ok, stringified_v} <- stringify_value(v),
            true <- is_notification_key_length_valid(stringified_v),
            true <- has_column(event_config, schema, table, k) do
        "#{table_topic}:#{k}=eq.#{stringified_v}"
        |> broadcast_change(change)
      end
    end)
end
```

abc3 (approved):

PR #206, opened 2021-12-15
Set the maximum number of active connections

abc3:

This PR allows setting the maximum number of active connections through the Ranch [max_connections](https://ninenines.eu/docs/en/ranch/1.7/manual/ranch/) parameter.

In current behavior, `Plug.Cowboy` uses the default [16_384](https://github.com/elixir-plug/plug_cowboy/blob/040c53f45bdbd3f1afe9f449fa1cb3b54294e7fb/lib/plug/cowboy.ex#L45) value.

## What is the new behavior?

In new behavior, it's possible to set any maximum number of active connections if start Realtime with environment variable MAX_CONNECTIONS. Soft limit.
By default, it uses 16_384 as earlier.


w3b6x9 (approved):
lgtm!

Issue #213, opened 2021-12-26
realtime only emitting `DELETE` events (no `INSERT` or `UPDATE`)


tconroy:
# Bug report
realtime only emitting `DELETE` events (no `INSERT` or `UPDATE`)

## Describe the bug
When I perform an `INSERT` or `UPDATE` to a table with realtime + RLS enabled, I do not receive any realtime event: 

```typescript
supabase
    .from('*')
    .on('*', (payload) => {
      console.log(`${payload.eventType} received: `, payload);
    })
    .subscribe();
```

`DELETE` events are correctly received.

The user has appropriate RLS permissions, because all data correctly loads on the client via supabase cllient `REST` calls. 

## To Reproduce
Unfortunately not able to share an example because it would be difficult to decouple the issue I'm seeing from domain-specific app code. However, here is my configuration:

**1. replication enabled:**
<img width="1408" alt="Screen Capture 2021-12-26 at 11 47 45@2x" src="https://user-images.githubusercontent.com/1609336/147414858-a97311d8-7e69-452d-a461-ad8b390b3421.png">

**2. replication enabled on tables:**
![Screen Capture 2021-12-26 at 11 48 29@2x](https://user-images.githubusercontent.com/1609336/147414908-ce5c8a82-bbca-47f6-8bcb-9f9c1ceb47a7.png)

**3. Subscription configuration:**
```typescript
supabase
    .from('*')
    .on('*', (payload) => {
      console.log(`${payload.eventType} received: `, payload);
    })
    .subscribe();
```

## Expected behavior
When I perform an `INSERT` or `UPDATE` on entities (such as `board_lists` table or `board_cards` table), I expect to see the following logged:

```
INSERT received: {payload data}
```

However, I **do not** see any realtime event **except** for `DELETE` operations, which are behaving as expected.

I am able to correctly perform `INSERT`, `UPDATE` and `READ` events on the client via the Supabase client/REST API, so I know that the RLS policies are not preventing the user from performing these operations.

## System / versions
* MacOS latest
* Node @ v17.2.0
* Yarn @ 1.22.17
* @supabase/supabase-js@1.29.0
* @supabase/realtime-js@^1.3.3
* @supabase/gotrue-js@^1.21.7
* @supabase/postgrest-js@^0.35.0

tconroy:
So it does seem to be something related to RLS policies. But I'm not sure exactly what the cause is.

My table `board_cards` has RLS Enabled with the following policy (`ALL`):
```
[...]

PR #214, opened 2021-12-28
fix: remove setting search path before replication

w3b6x9:

Bug fix

Postgres error when calling a Postgres function containing non-schema-qualified table name (e.g. referencing `test` instead of `public.test`) inside a RLS policy because `search_path` is set to `''`

## What is the new behavior?

Realtime RLS works when calling a Postgres function containing non-schema-qualified table name (e.g. referencing `test` instead of `public.test`) inside a RLS policy. This is resolved by removing setting `search_path` to `''`.

## Additional context

Related issues:

- [realtime only emitting DELETE events (no INSERT or UPDATE) #213](https://github.com/supabase/realtime/issues/213)
- [Realtime broken when policy use custom SQL functions #4683](https://github.com/supabase/supabase/issues/4683)

Related PR:

- https://github.com/supabase/walrus/pull/35

### DIFF
--- a/server/lib/realtime/rls/replication_poller.ex
+++ b/server/lib/realtime/rls/replication_poller.ex
@@ -49,7 +49,7 @@ defmodule Realtime.RLS.ReplicationPoller do
       :error, error -> {:error, error}
     end
     |> case do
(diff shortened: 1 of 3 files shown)


abc3 (approved):
lgtm!

PR #218, opened 2022-01-07
fix: update change timestamp to ISO 8601 Zulu format

w3b6x9:

Bug fix

timestamp is in the following example format:

```
2022-01-07T02:24:31.114082+00:00
```

## What is the new behavior?

timestamp is in iso 8601 zulu format. example format:

```
2022-01-06T05:49:57Z
```

## Additional context

Related issue: https://github.com/supabase/walrus/issues/37

### DIFF
--- a/server/priv/repo/migrations/20220107221237_update_change_timestamp_to_iso_8601_zulu_format.exs
+++ b/server/priv/repo/migrations/20220107221237_update_change_timestamp_to_iso_8601_zulu_format.exs
@@ -1,0 +1,242 @@
+defmodule Realtime.RLS.Repo.Migrations.UpdateChangeTimestampToIso8601ZuluFormat do
+  use Ecto.Migration
+
+  def change do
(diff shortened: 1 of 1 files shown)


server/priv/repo/migrations/20220107221237_update_change_timestamp_to_iso_8601_zulu_format.exs
```diff
@@ -0,0 +1,242 @@
...
+
+          else
+            output = jsonb_build_object(
+              'schema', wal ->> 'schema',
+              'table', wal ->> 'table',
+              'type', action,
+              'commit_timestamp', to_char(
+                (wal ->> 'timestamp')::timestamptz,
+                'YYYY-MM-DD\"T\"HH24:MI:SS\"Z\"'
+              ),
```
w3b6x9, line 144: @olirice here's the only change

olirice: looks good

olirice (approved):

PR #220, opened 2022-01-13
fix: default pg_logical_slot_get_changes upto_nchanges to 100

w3b6x9:

Bug fix

polling query waits to end of WAL to return response and can seconds, minutes, etc. to return response depending on frequency and size of database changes.

## What is the new behavior?

polling query returns response after grabbing 100 (default) changes.


abc3 (approved):
lgtm!

PR #229, opened 2022-02-08
feat: verify token of all roles coming from client

w3b6x9:

feature

only the `authenticated` role access token is being checked and updated when client sends it.

## What is the new behavior?

all roles' access tokens will be checked and updated.


abc3 (approved):
lgtm!

Issue #236, opened 2022-02-24
Realtime generates duplicate events on multiple computers when one is signed in and other is not.


GaryAustin1:
# Bug report
I'm not sure if this is realtime, or realtime.js....

I have two computers, on same router.  Both accessing similar code base on different hosts (one signs, in other skips signin).
On any change to a public non-RLS table two events will be generated for each change to both computers.
If both environments are logged in, or logged out together only one event is generated.

Edit:
I did an anon login from an iPad with LTE and it also gets double events, so it has nothing to do with common network.
Edit:
Adding a 3rd device with a 2nd different logged in user, still only generates double events.

## To Reproduce

Start up two computers with a simple app with a createClient using anon key.
On one computer have code run a sign In for authorized user.  On other skip signIn.
Subscribe to same table (event types do not matter).
Update or Insert on table.
Double payload events are generated.

## Expected behavior
Single event per device, or even browser.

## Screenshots
realtime subscription table.  Notice both were updated at same time, but the two computers connected at different times.

![realtimetable](https://user-images.githubusercontent.com/54564956/155439086-44008538-b2c7-484c-b4d1-9090e69df7fe.JPG)

Logged in computer.... Notice double insert.

![double-onelogged one not](https://user-images.githubusercontent.com/54564956/155439158-f699421e-6005-4a2c-b23b-9c2a26bd32b4.JPG)

Non logged in computer.  Notice null token.

![other browser](https://user-images.githubusercontent.com/54564956/155439393-85e85ba5-50ad-4315-a835-dc7e3f92e50d.JPG)

## System information

Current CDN Supabase.js (1.30.6)

## Additional context

This started off as an issue on a single computer with two different browsers, I moved to a second computer to see if localStorage was being shared by browsers somehow.  I believe the initial case was not even signed in, just had left over localStorage session from days ago signIn for another test...

w3b6x9:
@GaryAustin1 thanks for reporting this issue! i was able to find the root cause and i've put up a fix here: https://github.com/supabase/realtime/pull/239

PR #239, opened 2022-03-03
fix: prevent duplicate broadcast when multiple roles are subscribed

w3b6x9:

Bug fix

When RLS is disabled for a table and there are multiple database roles subscribed to Realtime (e.g. one `anon` user and `authenticated` user) then clients will receive duplicate messages for the same change.

## What is the new behavior?

Clients will receive a single message for the same change regardless of which database roles are subscribed.

## Additional context

Related issue: https://github.com/supabase/realtime/issues/236


abc3 (approved):
lgtm

PR #241, opened 2022-03-12
fix: update apply_rls function to convert timestamp to ISO 8601

w3b6x9:

Bug fix

Migration `20211228014915_add_wal_payload_on_errors_in_apply_rls_function` was applied after `20220107221237_update_change_timestamp_to_iso_8601_zulu_format`

## What is the new behavior?

New migration will update the `apply_rls` WALRUS function to the latest.


olirice (approved):
lgtm!

PR #248, opened 2022-04-04
Add runtime debug logging ability

Marviel:

Adds `UNSAFE_SHOW_DEBUG_LOGS` environment variable, which allows for easier debugging of the built Realtime container.

Currently, the default log-level is info -- but in certain circumstances it is helpful to be able to override the log-level with verbosity.

### DIFF
--- a/server/lib/realtime/application.ex
+++ b/server/lib/realtime/application.ex
@@ -35,16 +35,16 @@ defmodule Realtime.Application do
 
     replication_mode = Application.fetch_env!(:realtime, :replication_mode)
 
-    # When UNSAFE_ALL_DEBUG_LOGS is set, we show log-level of debug.
+    # When UNSAFE_SHOW_DEBUG_LOGS is set, we show log-level of debug.
     # This is marked unsafe, because it should not be used in production,
     # and is not generally guaranteed that DEBUG-level logs will be performant.
     unsafe_all_debug_logs =
       System.get_env(
-        "UNSAFE_ALL_DEBUG_LOGS",
+        "UNSAFE_SHOW_DEBUG_LOGS",
         nil
       )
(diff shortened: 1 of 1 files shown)


server/lib/realtime/application.ex
```diff
@@ -35,6 +35,20 @@ defmodule Realtime.Application do
 
     replication_mode = Application.fetch_env!(:realtime, :replication_mode)
 
+    # When UNSAFE_ALL_DEBUG_LOGS is set, we show log-level of debug.
+    # This is marked unsafe, because it should not be used in production,
+    # and is not generally guaranteed that DEBUG-level logs will be performant.
+    unsafe_all_debug_logs =
+      System.get_env(
+        "UNSAFE_ALL_DEBUG_LOGS",
```
kiwicopple, line 43:
```suggestion
        "UNSAFE_SHOW_DEBUG_LOGS",
```

Just a suggestion. I needed to read your comment to understand _what_ this does. Hopefully adding a verb makes it a bit clearer

Marviel: very fair, will update.

slaesh: just stumbled across this..
maybe keep `unsafe_all_debug_logs` and `UNSAFE_SHOW_DEBUG_LOGS` also in sync :)

w3b6x9:
@Marviel Hey there, just wanted to get a better sense of how useful this is.

As I recall during our debugging dialogue on your issue (https://github.com/supabase/supabase/issues/6290) you were trying to get Realtime RLS up and running.

I'm looking at the Realtime RLS-specific parts of the code and I don't see any `Logger.debug...`. Were you just adding them in in your local Realtime codebase?

PR #249, opened 2022-04-05
Config file name should be .env.local

johnoscott:
Point .3 in `examples/next-js/README.md` mentions creating `.env` file which does not work. Should be `.env.local`

Bug fix

API requests from nextjs client fail because environment variables are not set (because README says to create `.env` which does not get detected when running `npm run dev`)

## What is the new behavior?

Works (environment variables are read from correct file `.env.local`


w3b6x9:
@johnoscott awesome, thank you!

w3b6x9 (approved):
lgtm!

PR #251, opened 2022-04-17
Update ci.yml

J0:

Updates CI to have a parallel build which runs `credo` `mix format` and `dialyzer`

Commented linters

## What is the new behavior?

Linters run in a parallel build

## Additional context

Not sure if we want to add more versions for OTP/Elixir into the matrix


.github/workflows/ci.yml
```diff
@@ -48,14 +48,59 @@ jobs:
...
+
+    strategy:
+      matrix:
+        platform: [ubuntu-18.04]
+        otp: [23.x]
+        elixir: [1.11.x]
 
-      # - name: Run Credo
-      #   run: mix credo
+    runs-on: ubuntu-18.04
```
darora, line 66: Is there a reason to be on 18 and not the latest LTS?

J0: Not quite -- will modify to 20.04 instead. Thanks for the catch!

.github/workflows/ci.yml
```diff
@@ -48,14 +48,59 @@ jobs:
...
-      #   run: mix format --check-formatted
+    defaults:
+      run:
+        working-directory: ./server
 
-      # - name: Run Credo
-      #   run: mix credo
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
```
darora, line 62: I'm not qualified to comment on the elixir specifics, but given that we're only supporting a single target for each of these vars, getting rid of the matrix would simplify things. I don't think there's a reason to need multiple versions in the future either.

darora (approved):

PR #253, opened 2022-04-29
feat: enable IP version specification for realtime app

dudlf:

feature

Machine which does not support IPv6 can't run realtime due to there is no options for this.

## What is the new behavior?

User can specify IP version for realtime app.
Default behavior does not change.

## Additional context

I'm new to supabase, and does not have any experience on Elixir and Phoenix.
Is there any possible issue when realtime is binded to IPv4?


w3b6x9:
@dudlf are you currently having issues connecting Realtime server via IPv4? are you self-hosting?

dudlf:
> @dudlf are you currently having issues connecting Realtime server via IPv4? are you self-hosting?

I'm using self-hosting.

Silur:
self hosting doesn't work on systems where ipv6 is explicitly disabled (by sysctl or kernel flags), so an ip family flag/setting is indeed preferred

TeeWallz:
I experienced the issue this resolves today,

This PR would be really appreciated

w3b6x9 (approved):
@dudlf thanks for your contribution!

Issue #252, opened 2022-04-23
Error 413: Payload Too Large


barbalex:
# Bug report

## Describe the bug

I am subscribing to a table that contains files in a bytea field.
It seems that realtime does not like this as I get this error: `Error 413: Payload Too Large`

I am building an offline capable app that continually syncs data and files while online. So I need to know when a file was added.

I thought maybe I could get realtime to only send the id's of the changed datasets and then fetch the bytea columns with a regular query. But I have not found an option to choose what fields a payload contains.

What can I do?

## To Reproduce

Try to sync a table containing a bytea column that contains a file, in my case with 600KB.

## Expected behavior

Either: that syncing works for all fields, no matter what payload size is. 
Or: that I can choose what fields to include in the payload which enables fetching the large payload with a regular query.

Might be related to https://github.com/supabase/realtime/issues/119

w3b6x9:
> I have not found an option to choose what fields a payload contains.

@barbalex Realtime currently does not support choosing the columns you want to include in the payload.

Let me see if there's a good way to pass the primary key when the error is 413 so that you can make a query for the data instead. I'll report back here soon.

olirice:
@w3b6x9 here's the relevant source that filters out the `record` and `old_record`
https://github.com/supabase/walrus/blob/974798432bbb21d4838efa4ad3de2dd95f2b7edc/sql/walrus--0.1.sql#L455-L473

> Let me see if there's a good way to pass the primary key when the error is 413 

this risk in this case is that the primary key could be the one that includes the large amount of data e.g. a big string

What do you think about filtering out any columns that are > ~64 bytes.  That'll maintain guarantee we want about message size and will include the primary key columns unless the user sets the primary key as something crazy

w3b6x9:
@barbalex we have now addressed this by way of my colleague @olirice's suggestion:

> What do you think about filtering out any columns that are > ~64 bytes. That'll maintain guarantee we want about message size and will include the primary key columns unless the user sets the primary key as something crazy

please give it a try and let me know what you think.

PR #262, opened 2022-06-03
fix: add quoted regtype support and output when payload too large

w3b6x9:

Bug fix

1. Realtime WALRUS doesn't work with quoted regtypes. This has been a huge pain point for developers who use Prisma.
2. When payload size too large developers don't get enough information on the client.

## What is the new behavior?

1. Realtime WALRUS works with quoted regtypes ([WALRUS PR](https://github.com/supabase/walrus/pull/46))
2. When payload size too large developers receive columns whose payload size is less than or equal to 64 bytes ([WALRUS PR](https://github.com/supabase/walrus/pull/47))

## Additional context

Related issues:

1. https://github.com/supabase/supabase/issues/5685#issuecomment-1142714282
2. https://github.com/supabase/realtime/issues/252

### DIFF
--- a/server/priv/repo/migrations/20220603231003_add_quoted_regtypes_support.exs
+++ b/server/priv/repo/migrations/20220603231003_add_quoted_regtypes_support.exs
@@ -2,11 +2,17 @@ defmodule Realtime.RLS.Repo.Migrations.AddQuotedRegtypesSupport do
   use Ecto.Migration
 
   def change do
+    execute("drop type if exists realtime.wal_column cascade;")
+
     execute("
-      alter type realtime.wal_column
-      add attribute type_oid oid cascade,
-      drop attribute if exists type cascade,
-      add attribute type_name text cascade;
+      create type realtime.wal_column as (
+        name text,
+        type_name text,
+        type_oid oid,
+        value jsonb,
+        is_pkey boolean,
+        is_selectable boolean
+      );
     ")
 
     execute("
--- a/server/priv/repo/migrations/20220603231003_add_quoted_regtypes_support.exs
+++ b/server/priv/repo/migrations/20220603231003_add_quoted_regtypes_support.exs
@@ -1,0 +1,291 @@
+defmodule Realtime.RLS.Repo.Migrations.AddQuotedRegtypesSupport do
+  use Ecto.Migration
+
+  def change do
+    execute("
+      alter type realtime.wal_column
+      add attribute type_oid oid cascade,
+      drop attribute if exists type cascade,
+      add attribute type_name text cascade;
+    ")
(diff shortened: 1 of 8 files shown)


w3b6x9:
@olirice i'm getting this following error when running the test suite:

```elixir
%Postgrex.Error{connection_id: 248, message: nil, postgres: %{code: :cannot_coerce, detail: "Cannot cast type boolean to oid in column 5.", file: "parse_coerce.c", internal_position: "421", internal_query: "columns =\n              array_agg(\n                  (\n                      x->>'name',\n                      x->>'type',\n                      x->>'typeoid',\n                      realtime.cast(\n                          (x->'value') #>> '{}',\n                          (x->>'typeoid')::regtype\n                      ),\n                      (pks ->> 'name') is not null,\n                      true\n                  )::realtime.wal_column\n              )\n              from\n                  jsonb_array_elements(wal -> 'columns') x\n                  left join jsonb_array_elements(wal -> 'pk') pks\n                      on (x ->> 'name') = (pks ->> 'name')", line: "1104", message: "cannot cast type record to realtime.wal_column", pg_code: "42846", routine: "coerce_record_to_complex", severity: "ERROR", unknown: "ERROR", where: "PL/pgSQL function realtime.apply_rls(jsonb,integer) line 44 at assignment"}, query: nil}
```

Not sure if it has to do with my code or the newest `wal2json` version.

server/priv/repo/migrations/20220603231003_add_quoted_regtypes_support.exs
```diff
@@ -0,0 +1,291 @@
+defmodule Realtime.RLS.Repo.Migrations.AddQuotedRegtypesSupport do
+  use Ecto.Migration
+
+  def change do
+    execute("
+      alter type realtime.wal_column
+      add attribute type_oid oid cascade,
+      drop attribute if exists type cascade,
+      add attribute type_name text cascade;
```
olirice, line 9: This adjustment to the type changes the attribute order of the `realtime.wal_column` type

it is expected to be 
```sql
create type realtime.wal_column as (
    name text,
    type_name text,
    type_oid oid,
    value jsonb,
    is_pkey boolean,
    is_selectable boolean
);
```

but adding attributes to a type pushes them to the end, so we're getting
```sql
create type realtime.wal_column as (
    name text,
    value jsonb,
    is_pkey boolean,
    is_selectable boolean
    type_oid oid,
    type_name text
);
```

The error message you're receiving is on a line that casts a tuple to `realtime.wal_column` where the values are provided by position (vs name) so I think that will resolve the error

olirice (approved):
lgtm!

PR #280, opened 2022-08-18
fix: mirror docker image to ECR

sweatybridge:

simpler release workflow

sets before and after version as a workaround

## What is the new behavior?

- uses semantic releaser action instead
- publish to both ecr and docker hub with tags
- consistent with other repo https://github.com/supabase/postgres-meta/blob/master/.github/workflows/release.yml


w3b6x9 (approved):
@sweatybridge thanks!

PR #286, opened 2022-08-29
feat: support docker file-based secrets

burmecia:

This PR is to add an entry point file `docker-entrypoint.sh` to Docker image, which will enable use of Docker file-based secrets as mentioned in https://github.com/supabase/supabase/issues/6014.

Currently this Docker image only supports environment variable based secrets, file-based secrets `*_FILE` are not supported but is suggested by Docker in https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images.

## What is the new behavior?

File-based Docker secrets `*_FILE` will be supported and fall back to environment variable secrets.

## Additional context

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,5 +45,7 @@ COPY --from=app_builder /app/_build .
 # RUN chown -R app: ./prod
 # USER app
 
+ENTRYPOINT ["docker-entrypoint.sh"]
+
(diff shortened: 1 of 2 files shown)


w3b6x9 (approved):
@burmecia thank you!

PR #328, opened 2022-11-15
Caching a tenant's info

abc3:


lib/realtime/api.ex
```diff
@@ -202,4 +216,12 @@ defmodule Realtime.Api do
     |> Task.await_many()
     |> List.flatten()
   end
+
+  def drop_dist_cache(id) do
```
w3b6x9, line 220: @abc3 we should make this a private method.

lib/realtime/api.ex
```diff
@@ -202,4 +216,12 @@ defmodule Realtime.Api do
...
+    {_, bad_nodes} =
+      [node() | Node.list()]
+      |> :rpc.multicall(Cachex, :del, [:db_cache, id], 10_000)
+
+    if bad_nodes != [], do: Logger.error("Failed to drop cache on nodes: #{inspect(bad_nodes)}")
```
w3b6x9, line 225: @abc3 `bad_nodes` are a list of nodes that do not exist. It's equally, or more, important to log the badrpc reasons on nodes this call was able to reach.

lib/realtime/api.ex
```diff
@@ -77,6 +77,8 @@ defmodule Realtime.Api do
 
   """
   def update_tenant(%Tenant{} = tenant, attrs) do
+    drop_dist_cache(tenant.external_id)
```
w3b6x9, line 80: @abc3 did you confirm with Infra that this is alright with timeouts on their end? This adds 10 seconds to the database operation after this for a total of 25 seconds max.

Also, for this to take a max of 25 seconds seems way too long for this method call.

w3b6x9: @abc3 we should call `drop_dist_cache/1` in `create_tenant/1` as well to cover the edge case.

lib/realtime/api.ex
```diff
@@ -133,6 +137,16 @@ defmodule Realtime.Api do
     |> repo_replica.preload(:extensions)
   end
 
+  def get_cached_tenant(external_id) do
+    {status, value} =
+      Cachex.fetch(:db_cache, external_id, fn ->
+        get_tenant_by_external_id(external_id)
```
w3b6x9, line 143: @abc3 we have to check the result of this db operation and then conditionally `:commit` or `:ignore`. Fallback function should return a tuple. Let's return either `{:commit, %Tenant{}}` or `{:ignore, nil}`.

Also, we don't have to set this fallback function or expire (https://github.com/supabase/realtime/pull/328/files#diff-5adc67952ec1b4291e4fffde5db81f2b734aa61ba0e416cb244c953e54c2b25bR146) every time. We can just do it once on `start_link`.

See https://github.com/whitfin/cachex/blob/c3bf7d9cacd47db8cf7dae56a3a2c290b20d7218/lib/cachex.ex#L188 and https://github.com/whitfin/cachex/blob/c3bf7d9cacd47db8cf7dae56a3a2c290b20d7218/lib/cachex.ex#L163.

We should also think about adding in a cache limit too so we don't OOM: https://github.com/whitfin/cachex/blob/c3bf7d9cacd47db8cf7dae56a3a2c290b20d7218/lib/cachex.ex#L234

w3b6x9 (changes requested):

lib/realtime/api.ex
w3b6x9: @chasers did you set a cachex limit for logflare?

lib/realtime/api.ex
abc3: `bad_nodes` is a list of cluster members which didn't answer

w3b6x9:
```
It returns {ResL, BadNodes}, where BadNodes is a list of the nodes that do not exist, and ResL is a list of the return values, or {badrpc, Reason} for failing calls.
```
- https://www.erlang.org/doc/man/rpc.html#multicall-5

Let's log both.

lib/realtime/api.ex
abc3: this function is proxying before `get_tenant_by_external_id`, it can return `Tenant%{}` or `nil`

lib/realtime/api.ex
abc3: what edge cases do you mean?

abc3: How you calculated 25 seconds? 10 seconds is the maximum timeout to go over all cluster, infra doesn't need to know about this timeout anything

lib/extensions/postgres_cdc_rls/cdc_rls.ex
```diff
@@ -101,7 +101,8 @@ defmodule Extensions.PostgresCdcRls do
 
     args =
       Map.merge(args, %{
-        "db_socket_opts" => [addrtype]
+        "db_socket_opts" => [addrtype],
+        "subs_pool_size" => Map.get(args, "subscriptions_pool", 5)
```
abc3, line 105:
```suggestion
        "subs_pool_size" => Map.get(args, "subcriber_pool_size", 5)
```

lib/realtime/api.ex
w3b6x9: `10 seconds is the maximum timeout` + 15 seconds for database update max timeout.

Well there was an issue previously where we added something that extended the max timeout of a method call and that affected Infra.

lib/realtime/api.ex
w3b6x9: > If the Tuple is tagged :commit the value will be placed into the cache and then returned. If tagged :ignore the value will be returned without being written to the cache. If you return a value which does not fit this structure, it will be assumed that you are committing the value.

- https://hexdocs.pm/cachex/Cachex.html#fetch/4

Right but why are you trying to commit `nil` and then setting the new expiration. Just unnecessary work.

lib/realtime/api.ex
abc3: You can reproduce a bad node if you would try to exec a multicall function on timer.sleep which is bigger than the timeout
```erlang
:rpc.multicall(Node.list(), :timer, :sleep, [1000], 500)
{[], [:"node@127.0.0.1"]}
```
ReaL does not include a node name
```erlang
:rpc.multicall(Node.list(), Map, :new, [1], 500)
{[
   badrpc: {:EXIT,
    {%Protocol.UndefinedError{protocol: Enumerable, value: 1, description: ""},
     [
       {Enumerable, :impl_for!, 1, [file: 'lib/enum.ex', line: 1]},
       {Enumerable, :reduce, 3, [file: 'lib/enum.ex', line: 166]},
       {Enum, :reverse, 1, [file: 'lib/enum.ex', line: 4307]},
       {Enum, :to_list, 1, [file: 'lib/enum.ex', line: 3646]},
       {Map, :new_from_enum, 1, [file: 'lib/map.ex', line: 228]},
       {Map, :new, 1, []}
     ]}}
 ], []}
```

lib/realtime/api.ex
abc3: > If you return a value which does not fit this structure, it will be assumed that you are committing the value.

ohh, I missed that 🤦‍♂️

abc3: > Also, we don't have to set this fallback function or expire every time. We can just do it once on start_link.

I would like not to link the cache to the current process and to have a standalone cache service tree

w3b6x9: > standalone cache service tree

Right, so you can start_link in `Realtime.Application.start/2` no?

lib/realtime/api.ex
abc3: There was a blocking operation to inform tenant's users to stop. Currently, 15 seconds should be enough for everything.

PR #340, opened 2022-11-17
Custom resolver

abc3:


lib/extensions/postgres_cdc_rls/syn_handler.ex
```diff
@@ -0,0 +1,52 @@
...
+    {keep, stop} =
+      Enum.filter([pid1, pid2], fn pid ->
+        Enum.member?(fly_region_nodes, node(pid))
+      end)
+      |> case do
+        [pid] ->
+          {pid, if(pid != pid1, do: pid1, else: pid2)}
+
+        _ ->
+          if time1 < time2 do
```
w3b6x9, line 39: @abc3 why not keep the same logic as `syn`?

```suggestion
          if time1 > time2 do
```

See: https://github.com/ostinelli/syn/blob/4b5fbc17e8008af2780af61fbf6966c15036741e/src/syn_event_handler.erl#L270

lib/extensions/postgres_cdc_rls/syn_handler.ex
```diff
@@ -0,0 +1,52 @@
...
+            {pid1, pid2}
+          else
+            {pid2, pid1}
+          end
+      end
+
+    target = node(stop)
+    Logger.warn("Resolving #{name} conflict, target: #{inspect(target)}")
+    :rpc.call(target, DynamicSupervisor, :stop, [stop, :normal, 15_000])
```
w3b6x9, line 48: @abc3 we're not waiting on the return so better to use `async_call` so it's not blocking.

lib/extensions/postgres_cdc_rls/syn_handler.ex
```diff
@@ -0,0 +1,52 @@
...
+      ) do
+    fly_region = Realtime.PostgresCdc.aws_to_fly(region)
+
+    fly_region_nodes =
+      :syn.members(RegionNodes, fly_region)
+      |> Enum.map(fn {_, [node: node]} -> node end)
+
+    {keep, stop} =
+      Enum.filter([pid1, pid2], fn pid ->
+        Enum.member?(fly_region_nodes, node(pid))
```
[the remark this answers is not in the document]
w3b6x9: so this still won't guarantee that it'll be on the closest node.

lib/extensions/postgres_cdc_rls/syn_handler.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Extensions.PostgresCdcRls.SynHandler do
+  require Logger
+  alias RealtimeWeb.Endpoint
+
+  def on_process_unregistered(Extensions.PostgresCdcRls, name, _pid, _meta, reason) do
+    Logger.info("PostgresCdcRls terminated: #{inspect(name)} #{node()}")
+
+    broadcast_method =
+      if reason == :syn_conflict_resolution do
```
w3b6x9, line 9: `All of the callbacks, except for resolve_registry_conflict/4, are called on all the nodes of the cluster.`
- https://github.com/ostinelli/syn/blob/4b5fbc17e8008af2780af61fbf6966c15036741e/src/syn_event_handler.erl#L37

@abc3 does that mean every node's `on_process_unregistered` will be called with reason `:syn_conflict_resolution` so every node will broadcast to all the other nodes to shut down their postgres cdc process?

abc3: `resolve_registry_conflict` will be called only once, that is why it uses `broadcast/3`

lib/extensions/postgres_cdc_rls/syn_handler.ex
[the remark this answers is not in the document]
abc3: I think the resolver should only decide which process to keep alive. We can introduce another process that periodically checks some rules and transfers connection trees between nodes if needed. For example, by CPU / mem usage, ping time, Dijkstra's algorithm, etc

lib/extensions/postgres_cdc_rls/syn_handler.ex
abc3: Actually, this is doesn't matter to us, the main thing that we should keep only one process. I choose the oldest because it uses a slot first

PR #341, opened 2022-11-17
feat: enable Prometheus Ecto metrics

w3b6x9:

feature

## What is the new behavior?

Enable Prom Ecto metrics

### DIFF
--- a/lib/realtime/monitoring/prom_ex.ex
+++ b/lib/realtime/monitoring/prom_ex.ex
@@ -1,6 +1,4 @@
 defmodule Realtime.PromEx do
-  alias Realtime.PromEx.Plugins.{OsMon, Phoenix}
-
   @moduledoc """
   Be sure to add the following to finish setting up PromEx:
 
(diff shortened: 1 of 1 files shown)


abc3 (changes requested):
This plugin will provide info about all db hosts, including tenants' databases (migrations use Ecto), which significantly blows up metrics size. We should remove unnecessary fields from this plugin and expose only metrics of the internal db.

w3b6x9:
> This plugin will provide info about all db hosts, including tenants' databases (migrations use Ecto), which significantly blows up metrics size. We should remove unnecessary fields from this plugin and expose only metrics of the internal db.

@abc3 ah, good call. Tenant DB migrations use their own process but under Realtime.Repo. I can just take that out bc Replica.SIN is the same.

lib/realtime/monitoring/prom_ex.ex
```diff
@@ -56,21 +54,20 @@ defmodule Realtime.PromEx do
...
-      # Plugins.Application,
-      {Plugins.Beam, poll_rate: poll_rate, metric_prefix: [:beam]},
-      {Phoenix, router: RealtimeWeb.Router, poll_rate: poll_rate, metric_prefix: [:phoenix]},
-      # {Plugins.Ecto, poll_rate: poll_rate, metric_prefix: [:ecto]},
-      # Plugins.Oban,
-      # Plugins.PhoenixLiveView
-      {OsMon, poll_rate: poll_rate}
+      {Beam, poll_rate: poll_rate, metric_prefix: [:beam]},
+      # list `Realtime.Repo` in `repos` when decoupled from tenant database migrations
+      {Ecto, repos: [Replica.FRA, Replica.IAD, Replica.SIN], metric_prefix: [:ecto]},
```
w3b6x9, line 68: @abc3 `Realtime.Repo` no longer listed.

w3b6x9:
Closing as there appears to be an issue where `Realtime.Repo` metrics are reported even though it's not in the Plugin Ecto `repos` list.

PR #351, opened 2022-11-22
fix: change deployment strategy from bluegreen to rolling

w3b6x9:


w3b6x9:
Remember to update Fly's `max-per-region` appropriately once this is deployed.

abc3 (approved):

PR #359, opened 2022-11-30
fix: check incoming access token from client

w3b6x9:

Bug fix

Realtime channel will subscribe to Postgres changes on every incoming `access_token` message regardless of token.

## What is the new behavior?

Realtime channel will subscribe to Postgres changes on every incoming `access_token` message only when the token has changed, or in other words, refreshed.

## Additional context

- Same functionality already existed in single tenant Realtime (`main` branch): https://github.com/supabase/realtime/commit/4b73df8050739fccfcb75984a85d0a61e5a6ff2a#diff-42b9c485d7ea1a694807612fbc928f3b5d98db823edab2ec682214280e0e31b7R140

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -257,7 +257,6 @@ defmodule RealtimeWeb.RealtimeChannel do
             pg_change_params: pg_change_params,
             postgres_extension: postgres_extension,
             channel_name: channel_name,
-            tenant_topic: tenant_topic,
             postgres_cdc_module: module
           }
         } = socket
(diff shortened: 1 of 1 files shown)


chasers:
Can we get a test for this one?

abc3 (approved):

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -356,29 +355,39 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+          }
+        } = socket
       )
       when is_binary(refresh_token) do
     socket = assign_counter(socket) |> assign(:access_token, refresh_token)
 
     case confirm_token(socket) do
       {:ok, claims, confirm_token_ref} ->
-        cancel_timer(pg_sub_ref)
+        if refresh_token == access_token do
```
chasers, line 371: I think this would be much more clear if you put it in another function head in a guard.

Like: `when is_binary(access_token) and access_token == refresh_token`

Also this needs a `cancel_timer`.

chasers (commented):
Overall I think this is fine for now and will help, but auth is moving towards shorter and short expirations.  

We really should be going off of the `pg_change_params` or if the subscription query and the user are the same then don't resubscribe. 

If we only go off of a token exact match then we'll just be here again in 6 months.

PR #360, opened 2022-11-30
fix: handle Presence list call timeout without crashing channel

w3b6x9:

Bug fix

Phoenix Presence shard can be overwhelmed if many channels of the same topic try to call `Presence.list` at or around the same time. This leads to shard time out errors after 5 seconds which causes the channel to crash.

## What is the new behavior?

Channel will check to see if topic is currently tracking any presences by fetching from `ets` table directly. If there are presences then it reverts back to `Presence.list` call b/c only the shard state has access to down replicas.

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -7,7 +7,7 @@ defmodule RealtimeWeb.RealtimeChannel do
   require Logger
 
   alias DBConnection.Backoff
-  # alias Extensions.Postgres
+  alias Phoenix.Tracker.Shard
   alias RealtimeWeb.{ChannelsAuthorization, Endpoint, Presence}
   alias Realtime.{GenCounter, RateCounter, PostgresCdc}
 
@@ -226,7 +226,23 @@ defmodule RealtimeWeb.RealtimeChannel do
   @impl true
   def handle_info(:sync_presence, %{assigns: %{tenant_topic: topic}} = socket) do
     socket = assign_counter(socket)
-    push(socket, "presence_state", Presence.list(topic))
+
+    presence_list =
+      socket
+      |> presence_list_shard_bypass()
+      |> case do
+        {:ok, [_ | _]} ->
+          Presence.list(topic)
+
+        {:ok, []} ->
+          %{}
+
+        {:error, message} ->
+          Logger.error("Unable to fetch Presence list for topic #{topic}: #{message}")
+          Presence.list(topic)
+      end
+
+    push(socket, "presence_state", presence_list)
 
     {:noreply, socket}
   end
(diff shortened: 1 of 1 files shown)


chasers:
idk do we even know for sure this is an issue?  Presence has Shards for a reason.  I'd rather see some instrumentation around this call before we up and bypass a bunch of shit in the lib.  

Also, what if they change the underlying ETS structure and this breaks on a version update?

w3b6x9:
> idk do we even know for sure this is an issue? 
> I'd rather see some instrumentation around this call before we up and bypass a bunch of shit in the lib.

This is 100% an issue. See [logs from this month](https://logflare.app/sources/20275/search?querystring=%22exited+in%3A+GenServer.call%28RealtimeWeb.Presence_shard%22+t%3Athis%40month+c%3Acount%28%2A%29+c%3Agroup_by%28t%3A%3Aday%29&tailing%3F=false). The same developers keep reaching out about how they are having repeated Realtime issues and this has been the culprit for them.

> Presence has Shards for a reason. 

Right but if channels are all trying to sync/track Presence over a single topic then it's all over a single shard process. Additional shards won't help in that scenario. All channels listening to a topic will be routed to the same shard process on that node.

> Also, what if they change the underlying ETS structure and this breaks on a version update?

Well the Tracker state ets table hasn't changed in 6 years but I've updated this PR to account for that so reverts back to calling `Presence.list`. See https://github.com/supabase/realtime/pull/360/files#diff-2a3a48674821fef237a3451af5d33ca977ee5c7d2f6bc8ca18bd7f18b44a7a26R639 and https://github.com/supabase/realtime/pull/360/files#diff-2a3a48674821fef237a3451af5d33ca977ee5c7d2f6bc8ca18bd7f18b44a7a26R241-R242.

chasers:
Hmm, yeah happened when SIN had connection issues.

Did you see the `dirty_list`?

https://github.com/phoenixframework/phoenix_pubsub/blob/ca2b47c8cf31324b0bf96cea862058f783a3e7bd/lib/phoenix/tracker/shard.ex#L78

w3b6x9:
> Did you see the `dirty_list`?
> 
> https://github.com/phoenixframework/phoenix_pubsub/blob/ca2b47c8cf31324b0bf96cea862058f783a3e7bd/lib/phoenix/tracker/shard.ex#L78

No but I just swapped it in: https://github.com/supabase/realtime/pull/360/files#diff-2a3a48674821fef237a3451af5d33ca977ee5c7d2f6bc8ca18bd7f18b44a7a26R637.

chasers (approved):

chasers:
This is all the same stuff I did when trying to use Tracker to track rate limits across the cluster. Could never get it working.

PR #371, opened 2022-12-07
Feat: cluster latency matrix

chasers:
Problem: 
It's difficult to understand when our host is having subtle networking issues.

Solution:
We now have a GenServer to ping each node in the cluster from each node every 5 seconds. 

It logs:
- when the rpc fails
- when the latency of the rpc is over a certain value

We also have a LiveView to see realtime latencies between all nodes on a matrix:

Screen Shot 2022-12-07 at 12 42 10 PM


chasers:
@abc3 I'm having some really weird compilition issues in the `Realtime.Latency` module and could use your help on this one when you're back.

abc3 (approved):

PR #377, opened 2022-12-17
feat: streamline local Docker setup

w3b6x9:

chore

## What is the new behavior?

Local Docker setup for multi-tenant Realtime is a breeze.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ seed:
 	mix run priv/repo/seeds.exs
 
 prod:
-	MIX_ENV=prod FLY_APP_NAME=realtime-local API_KEY=dev SECURE_CHANNELS=true API_JWT_SECRET=dev METRICS_JWT_SECRET=dev FLY_REGION=fra FLY_ALLOC_ID=123e4567-e89b-12d3-a456-426614174000 DB_ENC_KEY="1234567890123456" SECRET_KEY_BASE=M+55t7f6L9VWyhH03R5N7cIhrdRlZaMDfTE6Udz0eZS7gCbnoLQ8PImxwhEyao6D DASHBOARD_USER=realtime_local DASHBOARD_PASSWORD=password ERL_AFLAGS="-kernel shell_history enabled" iex -S mix phx.server
+	MIX_ENV=prod FLY_APP_NAME=realtime-local SECURE_CHANNELS=true API_JWT_SECRET=dev METRICS_JWT_SECRET=dev FLY_REGION=fra FLY_ALLOC_ID=123e4567-e89b-12d3-a456-426614174000 DB_ENC_KEY="1234567890123456" SECRET_KEY_BASE=M+55t7f6L9VWyhH03R5N7cIhrdRlZaMDfTE6Udz0eZS7gCbnoLQ8PImxwhEyao6D DASHBOARD_USER=realtime_local DASHBOARD_PASSWORD=password ERL_AFLAGS="-kernel shell_history enabled" iex -S mix phx.server
 
 swagger:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ seed:
 	mix run priv/repo/seeds.exs
 
 prod:
-	MIX_ENV=prod FLY_APP_NAME=realtime-local API_KEY=dev SECURE_CHANNELS=true API_JWT_SECRET=dev METRICS_JWT_SECRET=dev FLY_REGION=fra FLY_ALLOC_ID=123e4567-e89b-12d3-a456-426614174000 DB_ENC_KEY="1234567890123456" SECRET_KEY_BASE=M+55t7f6L9VWyhH03R5N7cIhrdRlZaMDfTE6Udz0eZS7gCbnoLQ8PImxwhEyao6D DASHBOARD_USER=realtime_local DASHBOARD_PASSWORD=password ERL_AFLAGS="-kernel shell_history enabled" iex -S mix phx.server
```
w3b6x9, line 11: **REMINDER:** `API_KEY` is no longer used so remove it from staging and prod.

Dockerfile
```diff
@@ -112,7 +112,7 @@ RUN mkdir -p /var/run/tailscale /var/cache/tailscale /var/lib/tailscale
 COPY limits.sh /app/limits.sh
 ENTRYPOINT ["/app/limits.sh"]
 
-CMD ["/tailscale/wrapper.sh"]
+CMD ["sh", "-c", "/tailscale/wrapper.sh && /app/bin/migrate && /app/bin/server"]
```
abc3, line 115: We should check the Fly's behavior in the case when migration unsuccessful

docker-compose.dbs.yml
```diff
@@ -1,16 +1,14 @@
 version: '3'
+
 services:
   db:
-    image: supabase/postgres:14.1.0
+    image: supabase/postgres:14.1.0.105
```
abc3, line 5: Is there postgres v15 already available?

tailscale/wrapper.sh
```diff
@@ -9,6 +9,3 @@ if [ "${ENABLE_TAILSCALE}" = true ]; then
     /tailscale/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/var/run/tailscale/tailscaled.sock &
     /tailscale/tailscale up --authkey=${TAILSCALE_AUTHKEY} --hostname="${TAILSCALE_APP_NAME}" --accept-routes=true
 fi
-
-echo "Starting Realtime"
-sudo -E -u nobody /app/bin/server
```
abc3, line 14: Let's rename this file if it is not a wrapper anymore

w3b6x9: @abc3 it's still a wrapper for tailscale though and it's name-spaced appropriately: `tailscale/wrapper.sh`

docker-compose.dbs.yml
w3b6x9: @abc3 Yes, but most projects are on pg v14 so we should stick with it for the time being.

Dockerfile
w3b6x9: @abc3 looks like Fly's release command "perform database migrations" and "run in a temporary VM that is created just before your application is deployed and released" so it helps with "zero-downtime deploys" but what about local development or self-hosting that's not on Fly? How will we run db migrations then?

w3b6x9: @abc3 I added back the release command and will take care of migrations `docker-compose.yml`.

abc3 (approved):

PR #416, opened 2023-01-08
feat: stop connection to a tenant's db if there are no connected user…

abc3:
Stop connection to a tenant's db if there are no connected users for 10 minutes.


chasers (approved):

lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
```diff
@@ -79,16 +79,7 @@ defmodule Extensions.PostgresCdcRls.SubscriptionsChecker do
       Subscriptions.delete_multi(state.conn, ids)
```
w3b6x9, line 79: @abc3 we should only delete from `ets` (`:ets.delete(tid, pid)`) once `Subscriptions.delete_multi(state.conn, ids)` is successful.

lib/extensions/postgres_cdc_rls/subscription_manager.ex
```diff
@@ -169,6 +176,26 @@ defmodule Extensions.PostgresCdcRls.SubscriptionManager do
     {:noreply, %{state | delete_queue: %{ref: ref, queue: q1}}}
   end
 
+  def handle_info(:check_no_users, %{subscribers_tid: tid, no_users_ts: ts} = state) do
+    H.cancel_timer(state.no_users_ref)
+
+    ts2 =
```
w3b6x9, line 182: @abc3 i would rename this to `ts_new` or something. `ts2` is confusing.

lib/extensions/postgres_cdc_rls/subscription_manager.ex
```diff
@@ -169,6 +176,26 @@ defmodule Extensions.PostgresCdcRls.SubscriptionManager do
...
+      case {:ets.info(tid, :size), ts != nil && ts + @stop_after < now()} do
+        {0, true} ->
+          Logger.info("Stop tenant #{state.id} because of no connected users")
+          Rls.handle_stop(state.id, 15_000)
```
w3b6x9, line 186: @abc3 let's say this synchronous action does take 15 seconds can't this block and choke the process if the tenant suddenly has an influx of subscribers sending `:subscribed` messages? won't this potentially prevent the dynamic supervisor tree from shutting down?

lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
abc3: Good call! I've added this to tasks and will do it in another PR

lib/extensions/postgres_cdc_rls/subscription_manager.ex
abc3: If some users have tried to subscribe, they will receive the `postgres_cdc_down` when `handle_stop/2` is terminated and will try to resubscribe and make a new db connection.

The `handle_stop/2` function will stop the existing supervisor `PID` (or crush `SubscriptionManager` because timeout raises an error).

PR #420, opened 2023-01-09
fix: add http checks

chasers:
Adds http checks which will hopefully reduce error responses during rolling deploys.


abc3 (approved):

deploy/fly/prod.toml
```diff
@@ -40,3 +39,15 @@ processes = []
...
     restart_limit = 6
     timeout = "2s"
+  
+  [[services.http_checks]]
+    interval = 10000
+    grace_period = "5s"
+    method = "get"
+    path = "/"
+    protocol = "http"
+    restart_limit = 0
```
w3b6x9, line 49: @chasers wouldn't it be better to set this to a positive number so `n` consecutive failing health checks will trigger a restart?

deploy/fly/prod.toml
```diff
@@ -40,3 +39,15 @@ processes = []
     interval = "15s"
     restart_limit = 6
     timeout = "2s"
+  
+  [[services.http_checks]]
+    interval = 10000
+    grace_period = "5s"
+    method = "get"
+    path = "/"
+    protocol = "http"
```
w3b6x9, line 48: @chasers we should set this to `https` if we're going to set `tls_skip_verify` to `false`

deploy/fly/prod.toml
chasers: Eventually but I don't want to do that now so we change change any behavior of the cluster after a node is up and running.

deploy/fly/prod.toml
chasers: "When true (and using HTTPS protocol) skip verifying the certificates sent by the server."

Seems to me that this should be `true` when using `https`.

chasers: And I'm thinking these checks happen internally so unless we actually serve a cert from the instance we need to do http.  Although not sure if they're terminating https at the load balancer or what.

chasers: Gonna give it a go and see what happens.

PR #453, opened 2023-01-19
feat: high res commit_timestamp in Postgres change payload

chasers:
- `commit_timestamp` from walrus is now high res `2023-01-19T21:57:13.694Z` 
  - Previously was `2023-01-19T13:58:43Z`
- Latency for Postgres changes is now accurate in the Inspector
- Latency on Footer matches precision and style as Postgres change latency

Screen Shot 2023-01-19 at 3 02 19 PM

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -79,7 +79,7 @@ Hooks.payload = {
         <td class="py-4 px-6">${iso_ts.toISOString()}</td>
         <td class="py-4 px-6">
           <div class="pb-3">${JSON.stringify(payload)}</div>
-          <div class="pt-3 border-t hover:bg-gray-50">Latency: ${latency.toFixed(2)} ms</div>
+          <div class="pt-3 border-t hover:bg-gray-50">Latency: ${latency.toFixed(1)} ms</div>
         </td>
       </tr>`
     let list = document.querySelector("#plist")
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -69,16 +69,17 @@ Hooks.payload = {
 
   // Listen for all (`*`) `postgres_changes` events on tables in the `public` schema
   this.channel.on("postgres_changes", { event: "*", schema: schema, table: table }, payload => {
-    let ts = new Date();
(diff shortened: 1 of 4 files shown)


olirice (changes requested):
@chasers I'm on board with increasing precision but if we are going to edit the walrus code lets make that change at supabase/walrus first and then port the migration here afterward. If they get out-of-sync the next update to `realtime.apply_rls` in walrus will revert this change

Extra context:
Previous issues related to timestamp formatting 
https://github.com/supabase/walrus/issues/37
https://github.com/supabase/walrus/issues/14

w3b6x9 (changes requested):
This needs to be in ISO 8601 format.

chasers:
Screen Shot 2023-01-19 at 3 00 27 PM

chasers:
@olirice sure thing, but this repo and the migration here is what actually defines the `apply_rls` function. Does some other system try to define it?

w3b6x9:
>Does some other system try to define it?

@chasers these types of things originate in the `walrus` repo and after it's approved and merged there then we take the migration(s) and apply them to Realtime. All changes should originate from `walrus` and then copied over to `realtime` repo.

olirice:
once there's consensus on exactly what format/precision to use I'll make the walrus PR and get it under test. lmk!

chasers:
@w3b6x9 I'd like to use these hi-res timestamps to show a client how to monitor Realtime on Monday. 

Is there an issue you see with ISO 8601 with millis?

w3b6x9:
> Is there an issue you see with ISO 8601 with millis?

@chasers nope, we just need the migration over in walrus first.

w3b6x9:
@olirice can you please take a quick look at https://github.com/supabase/realtime/blob/0d177e1078c2efdb612b0316b5f9629f5543cac8/priv/repo/postgres/migrations/20230119133233_millisecond_precision_for_walrus.exs?

w3b6x9 (approved):

PR #458, opened 2023-01-26
fix: create ReplicationPoller tenant db conn in handle_continue

w3b6x9:
Move `ReplicationPoller` db connecting to handle_continue so it won't block the process from coming up which in turn won't block the supervisor from spinning up `SubscriptionManager` and `SubscriptionsChecker`.

### DIFF
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -29,16 +29,6 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
 
   @impl true
   def init(args) do
-    {:ok, conn} =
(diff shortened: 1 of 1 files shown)


w3b6x9:
> @w3b6x9 Are you sure that `Postgrex.start_link` is a blocking operation?

@abc3 I'll just let ChatGPT answer:

> Yes, the Postgrex.start_link/3 function in the Elixir programming language is a blocking function. It establishes a connection to a PostgreSQL database and starts a process to handle the connection, which can take some time to complete. This function blocks until the connection is established and the process is started.

abc3:
Lol :D

Anyway, I've checked with the db proxy by putting the sleep command in the connection and can confirm that `Postgres.start_link` is not blocking

PR #460, opened 2023-01-27
fix: make rlimit configurable by env var

sweatybridge:

rlimit is always set to 100k, causing problem when ran in docker rootless mode

## What is the new behavior?

- uses system default if `RLIMIT_NOFILE` is unset (empty string)
- naming follows linux convention https://man7.org/linux/man-pages/man2/getrlimit.2.html

## Additional context


w3b6x9 (approved):
LGTM, thank you!

PR #469, opened 2023-02-02
feat: use regex for 'in' filter parentheses value

w3b6x9:


chasers (changes requested):
Actually can you put this in a `defp` so the `with` statement is easier to read?

chasers (approved):

PR #471, opened 2023-02-03
fix: sort region nodes by node name

chasers:
Sorts region nodes by node name in case the node list order is unstable.

### DIFF
--- a/lib/realtime/postgres_cdc.ex
+++ b/lib/realtime/postgres_cdc.ex
@@ -1,6 +1,8 @@
 defmodule Realtime.PostgresCdc do
   @moduledoc false
 
+  require Logger
+
   @timeout 10_000
   @extensions Application.compile_env(:realtime, :extensions)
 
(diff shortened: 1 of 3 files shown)


lib/realtime/postgres_cdc.ex
```diff
@@ -78,9 +78,15 @@ defmodule Realtime.PostgresCdc do
...
-  @spec region_nodes(String.t()) :: [{pid, any}]
+  @doc """
+  Lists the nodes in a region. Sorts by node name in case the list order
+  is unstable.
+  """
+
+  @spec region_nodes(String.t()) :: [{pid(), [node: atom()]}]
   def region_nodes(region) when is_binary(region) do
     :syn.members(RegionNodes, region)
+    |> Enum.sort_by(fn {_pid, [node: node]} -> node end)
```
abc3, line 89: Should we update the current function to return only a list of nodes? Pid inside seems useless

chasers: Sure

abc3 (approved):
>should just error or start retrying until we have two launch nodes in a region to make this behavior more deterministic and/or until we support more replication slots

I don't think it's necessary, it's a pretty rare case

PR #473, opened 2023-02-04
feat: status page improvements

chasers:
- [x] adds regions to pings
- [x] uses node short name

## Future work
- dropdowns to select `from` and `to` regions
- show seconds from last ping
- color something orage > red depending on latency and maybe seconds from last ping
- dropdown for sort by

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -110,7 +110,7 @@ defmodule Realtime.Helpers do
 
       iex> node = Node.self()
       iex> Realtime.Helpers.short_node_id_from_name(node)
-      "localhost"
+      "nohost"
 
       iex> node = :"realtime-prod@fdaa:0:cc:a7b:b385:83c3:cfe3:2"
       iex> Realtime.Helpers.short_node_id_from_name(node)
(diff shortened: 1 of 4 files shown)


lib/realtime/helpers.ex
```diff
@@ -103,6 +103,54 @@ defmodule Realtime.Helpers do
...
+      iex> node = :"pink@10.0.1.1"
+      iex> Realtime.Helpers.short_node_id_from_name(node)
+      "10.0.1.1"
+  """
+
+  @spec short_node_id_from_name(atom()) :: String.t()
+  def short_node_id_from_name(name) when is_atom(name) do
+    [_, ip] = name |> Atom.to_string() |> String.split("@", parts: 2)
+
+    {:ok, ip} = ip |> handle_nohost() |> String.to_charlist() |> :inet.parse_address()
```
abc3, line 132: But this only works for IP addresses. the node with hostname, for example, node@localhost will produce an error

abc3 (approved):

lib/realtime/helpers.ex
chasers: Fixed!

PR #480, opened 2023-02-09
feat: cache tenants

chasers:
- `Tenants.Cache` module provides read-through cache functionality for any functions in the module
- `Realtime.ContextCache` to easily build caches for other contexts
- Uses `Tenants.Cache` in `Tenant` metrics poller
- Uses `Tenants.Cache` in `UserSocket`
- `Tenants.Cache` ttl is set to expire in 30 seconds

### DIFF
--- a/lib/realtime/context_cache.ex
+++ b/lib/realtime/context_cache.ex
@@ -13,11 +13,9 @@ defmodule Realtime.ContextCache do
            {:commit, {:cached, apply(context, fun, args)}}
          end) do
       {:commit, {:cached, value}} ->
-        IO.inspect("MISS")
         value
 
       {:ok, {:cached, value}} ->
-        IO.inspect("HIT")
         value
     end
   end
--- a/lib/realtime/context_cache.ex
+++ b/lib/realtime/context_cache.ex
@@ -1,0 +1,28 @@
+defmodule Realtime.ContextCache do
+  @moduledoc """
+    Read through cache for hot database paths.
+  """
+
+  require Logger
+
+  def apply_fun(context, {fun, arity}, args) do
+    cache = cache_name(context)
+    cache_key = {{fun, arity}, args}
+
+    case Cachex.fetch(cache, cache_key, fn {{_fun, _arity}, args} ->
+           {:commit, {:cached, apply(context, fun, args)}}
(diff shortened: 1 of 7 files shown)


lib/realtime/tenants/cache.ex
```diff
@@ -0,0 +1,23 @@
...
+
+  require Cachex.Spec
+
+  alias Realtime.Tenants
+
+  def child_spec(_) do
+    %{
+      id: __MODULE__,
+      start:
+        {Cachex, :start_link, [__MODULE__, [expiration: Cachex.Spec.expiration(default: 30_000)]]}
```
abc3, line 14: cache invalidation works on each node independently, and this can lead to inconsistency across the cluster

lib/realtime/context_cache.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Realtime.ContextCache do
+  @moduledoc """
+    Read through cache for hot database paths.
+  """
+
+  require Logger
+
+  def apply_fun(context, {fun, arity}, args) do
+    cache = cache_name(context)
+    cache_key = {{fun, arity}, args}
```
w3b6x9, line 10: @chasers can't we simplify the `cache_key` to just `{fun, args}`? why do we need arity if the `args` make the function unique in the key?

lib/realtime/tenants/cache.ex
```diff
@@ -0,0 +1,23 @@
+defmodule Realtime.Tenants.Cache do
+  @moduledoc """
+  Cache for users.
```
w3b6x9, line 3:
```suggestion
  Cache for tenants.
```

w3b6x9 (approved):

lib/realtime/tenants/cache.ex
chasers: v2 of this will listen to the wal, and invalidate the cache.

lib/realtime/context_cache.ex
chasers: Wasn't conscious really ... __ENV__.function returns `{fn_name, arity}`. 

I guess this makes it more obvious what the cache key actually is.

PR #495, opened 2023-02-21
feat: add cluster total connected users to metrics

chasers:
- [x] adds connected_cluster metric to prom endpoint
- [x] also logs this metric to Logflare

### DIFF
--- a/lib/realtime/monitoring/prom_ex/plugins/tenant.ex
+++ b/lib/realtime/monitoring/prom_ex/plugins/tenant.ex
@@ -33,8 +33,15 @@ defmodule Realtime.PromEx.Plugins.Tenant do
         last_value(
           [:realtime, :connections, :connected],
           event_name: [:realtime, :connections],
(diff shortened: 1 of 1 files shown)


lib/realtime/monitoring/prom_ex/plugins/tenant.ex
```diff
@@ -53,11 +60,12 @@ defmodule Realtime.PromEx.Plugins.Tenant do
 
     for t <- tenants do
       count = UsersCounter.tenant_users(Node.self(), t)
+      cluster_count = UsersCounter.tenant_users(t)
```
abc3, line 63: This value will be the same on all nodes and each node logs it. Is what you expect?

w3b6x9 (approved):

lib/realtime/monitoring/prom_ex/plugins/tenant.ex
chasers: Yep!

PR #531, opened 2023-03-02
feat: support rs256 signing

ecchochan:

- ~Added `API_JWT_SIGNING_METHOD` config (`api_jwt_signing_method`)~
- ~Added `METRICS_JWT_SIGNING_METHOD ` config (`metrics_jwt_signing_method`)~
- ~Added `jwt_signing_method` field in tenant~
- Increase `jwt_secret` field size in tenant (so that it can store long private long)
- Support `RS` sign algorithms in `lib/realtime_web/channels/auth/jwt_verification.ex`
- ~Added `jwt_verification_test_rs.exs` similar to `jwt_verification_test.exs` but for RS~

No Support of RS256

## What is the new behavior?

Support RS256

## Additional context

- https://github.com/supabase/supabase/discussions/12759


chasers:
@abc3 this looks pretty complete. Let's review tomorrow.

w3b6x9:
@ecchochan thanks for your contribution! this looks great.

can we keep it simpler for the initial version and just check for the `alg` from the JWT header?

so for this PR just keep:

- your changes in `lib/realtime_web/channels/auth/jwt_verification.ex` while adding another `generate_signer/2` that returns the specific error of missing alg:

```elixir
defp generate_signer(header, _secret) when not is_map_key(header, "alg") do
  {:error, :missing_alg_header}
end
```

- your migration but only modify `tenants` table `jwt_secret` column to type `text`:

```elixir
modify(:jwt_secret, :text, from: :string)
```

- https://github.com/supabase/realtime/pull/531/commits/09d81a887f59a9fbf6969bad85a43be175a8262d

w3b6x9 (changes requested):
see: https://github.com/supabase/realtime/pull/531#issuecomment-1579859025

ecchochan:
@w3b6x9 Thanks for reviewing! Did a cleanup and you can have a look!

Lan-Hekary:
@ecchochan you mean my comment ?

ecchochan:
@Lan-Hekary Yes, the current implementation in this PR is respecting the `alg` in the header, which according to your [comment](https://github.com/orgs/supabase/discussions/12759#discussioncomment-5995098) is a security risk!

This PR has to change to address the issue, thank you for raising that concern!

Lan-Hekary:
@ecchochan you mean my comment ?
First of all, Thank you for your effort in this PR.
I did some digging in the matter of Key Confusion attacks.
The problem rises from passing the key to the verification function along with the `alg` header without checking which key should be used.

I am not very good with Elixir Programming Language, but from what I see in the commit you pass the same `jwt_secret` as PEM to create a signer. I am afraid this is risky, because if an attacker passed the same `jwt_secret` for RS256 which is the public key and modify the `alg` to HS256 he can sign his own JWT with your public key and the validation will be ok.
The Correct approach is to store multiple keys in a Key set, The standard is [JWK ](https://datatracker.ietf.org/doc/html/rfc7517)

In PostgREST for example you set the keys in a JWKs object and the library choose the correct key based on the `alg` header.
and the JWKs can have multiple keys, symmetric for HS signing and asymmetric for RSA, EC, and Even EdDSA.
I tested this concept with Supabase and I can generate keys of various types and their respective JWTs and validate them successfully, but I can't confuse the "jose" library in PostgREST

we need a way to store the JWKs for realtime as well.
and choose which keys to use based on `alg` header

This also got me thinking that in PostgREST it's very easy to modify the JWKs by:
`alter role authenticator set pgrst.jwt_secret = {jwks_json_object}` 
and then issue and notification to `pgrst` channel.

how can we do that same in Realtime ? 
Also we should be able to change it without restarting the project or the container (Can't set it via env variable)
and preferably there should be one configuration the reflect in all the components ( PostgREST, Realtime, GoTrue, etc)

Lan-Hekary:
@ecchochan, can you check https://hexdocs.pm/joken/2.6.0/Joken.Signer.html#parse_config/1
I think it takes care of all the key separation, and it even supports multiple key representations.
All you have to do is pass it the keys and the algorithms, which is by defentions is what a JWKs (JSON Web Key Set) is.
take a look at the 'kty' key in JWK and JWS RFC.

All we need now is a way to configure a JWKs in Supabase globally and we can use the same JWKs across Realtime and PostgREST without any hassle.

I feel optimistic that this is an easy implantation that will not break exciting behavior nor be a security risk.

PR #538, opened 2023-03-06
feat: track db events

chasers:
- [x] updates metric names to be more like the limiter fields in the db

## Note
- Found the spot to start tracking db events, can do this whenever

### DIFF
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -121,9 +121,6 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
         end)
         |> Enum.reverse()
         |> Enum.each(fn change ->
-          # Remove when ready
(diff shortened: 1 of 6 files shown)


lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -120,6 +121,9 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
         end)
         |> Enum.reverse()
         |> Enum.each(fn change ->
+          # Remove when ready
+          # GenCounter.add(Tenants.events_per_second_key(tenant))
+
```
abc3, line 126: Do you want to log the count of messages received from db to Realtime or measure how many messages from db Realtime send to clients? For the last, the right place is here https://github.com/supabase/realtime/blob/337ad15a0f464328fe506ff6593422444eaf2c9a/lib/extensions/postgres_cdc_rls/message_dispatcher.ex#L53

abc3 (approved):

PR #547, opened 2023-03-13
feat: support 100k connections

chasers:
We're getting close to 16k connections on one node when a node in a region is restarted.

This will give us some room for growth.


abc3 (approved):

chasers:
Squashing and merging to see what sem release does with this.

PR #558, opened 2023-03-16
fix: dirty_list Presence

chasers:
Presence is eventually consistent anyways why run `list` through a GenServer?

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -636,7 +636,7 @@ defmodule RealtimeWeb.RealtimeChannel do
     })
   end
 
-  defp presence_dirty_list(topic) do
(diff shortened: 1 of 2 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -651,17 +636,13 @@ defmodule RealtimeWeb.RealtimeChannel do
     })
   end
 
-  defp presence_list_shard_bypass(%{assigns: %{tenant_topic: topic}} = _socket) do
-    try do
-      [{:pool_size, size}] = :ets.lookup(Presence, :pool_size)
+  defp presence_dirty_list(topic) do
```
abc3, line 639: it would be useful for debugging
```suggestion
  def presence_dirty_list(topic) do
```

PR #559, opened 2023-03-22
fix: only delete tenant from database on tenant delete endpoint

w3b6x9:

Bug fix

Tenant delete endpoint would return a `503` error, or take a long time, when it fails to stop all the processes related to `PostgresCdc`. 

## What is the new behavior?

Tenant delete endpoint deletes the tenant from Realtime database, and then creates an async process to stop all processes, so it'll take at most 15 seconds or so and won't return a `503`.

### DIFF
--- a/lib/realtime_web/controllers/tenant_controller.ex
+++ b/lib/realtime_web/controllers/tenant_controller.ex
@@ -5,7 +5,6 @@ defmodule RealtimeWeb.TenantController do
   require Logger
 
   alias Realtime.Api
-  alias Realtime.Repo
   alias Realtime.Api.Tenant
   alias Realtime.PostgresCdc
   alias RealtimeWeb.{Endpoint, UserSocket}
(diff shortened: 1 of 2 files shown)


lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -122,27 +121,12 @@ defmodule RealtimeWeb.TenantController do
...
-    )
-    |> case do
-      {:error, reason} ->
-        Logger.error("Can't remove tenant #{inspect(reason)}")
-        send_resp(conn, 503, "")
-
-      _ ->
-        send_resp(conn, 204, "")
+    case Api.delete_tenant_by_external_id(id) do
+      true -> id |> UserSocket.subscribers_id() |> Endpoint.broadcast("disconnect", %{})
```
abc3, line 125: tenant record can be cached and this allows users to subscribe again

abc3 (changes requested):

abc3:
Also, I think that if we try entirely deleting the tenant from the system, we should be sure that connected to the db also stopped.

abc3:
My comment above doesn't mean that it's necessary needs stop immediately cdc by the deleting request, we can store the tenant id in some queue and then tries to kill connections in a background process.

lib/realtime_web/controllers/tenant_controller.ex
w3b6x9: this was one of my concerns with this approach

abc3 (approved):

w3b6x9:
@chasers need you to unblock this so we can ship.

chasers (approved):

chasers:
@w3b6x9 sorry missed this one. Bug me about it next time!

PR #560, opened 2023-03-22
feat: fast lane outgoing broadcast messages

w3b6x9:

feat

## What is the new behavior?

Fast lane outgoing broadcast messages, and count them, so they don't need to be relayed through channel processes.


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -249,8 +275,6 @@ defmodule RealtimeWeb.RealtimeChannel do
 
   @impl true
   def handle_info(%{event: type, payload: payload}, socket) do
-    socket = count(socket)
```
w3b6x9, line 252: taking this out b/c we should count only when we are absolutely sure we want to.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -240,6 +259,13 @@ defmodule RealtimeWeb.RealtimeChannel do
     {:noreply, socket}
   end
 
+  @impl true
+  def handle_info(%Broadcast{event: "presence_diff", payload: payload}, socket) do
```
w3b6x9, line 263: intercepting `presence_diff` messages here. for some reason the `handle_out` wasn't triggering but this was.

lib/realtime_web/channels/realtime_channel.ex
chasers: I don't know, this is a catch all for a reason.

I would remove the whole function then and make sure we specifically handle all `event` types. Which, I think we are.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -453,7 +491,6 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   def handle_in(_, _, socket) do
-    socket = count(socket)
```
chasers, line 456: Again, make sure we're handling all messages and remove the whole function.

lib/realtime_web/channels/realtime_channel.ex
chasers: Nice. handle_out re-serializes everything anyways.

chasers:
Putting this here too for comms. It is handling outbound messages already:

https://user-images.githubusercontent.com/1019814/227007147-551b2279-426d-4c7f-8572-ec284ad29911.mp4

You did catch some something with having both Postgres extensions use the same dispatcher though. Need to do that.

And needs tests to make this obvious for the future if you want to help out with that?

chasers:
Maybe it's better because you're manually fastlane-ing them now? 

But doesn't Channels do that for us anyways with the default dispatcher? But maybe we're not getting that benefit because we're subscribing to `tenant <> : <> sub_topic` which is different than what the client is subscribing to.

That kind of always smelled odd to me actually. Why would clients not join `realtime:tenant:topic`? I think then we wouldn't have to subscribe to it and it would get fastlaned automatically. We'd still need a custom dispatcher to do the counting probably.

abc3:
> That kind of always smelled odd to me actually. Why would clients not join realtime:tenant:topic?

@chasers this was made for backward compatibility when we were transferring users to multiplayer

lib/realtime_web/channels/realtime_channel.ex
w3b6x9: yep, removing this.

lib/realtime_web/channels/realtime_channel.ex
w3b6x9: yep, removing

PR #592, opened 2023-05-11
feat: use the Inspector with a JWT signed with custom claims

chasers:
Screenshot 2023-05-11 at 4 32 49 PM


abc3 (approved):

lib/realtime_web/live/inspector_live/conn_component.html.heex
```diff
@@ -104,7 +104,21 @@
...
+                    my-1
+                    block
+                    w-full
+                    rounded-md
+                    border-gray-300
+                    shadow-sm
+                    focus:border-indigo-300 focus:ring focus:ring-indigo-200 focus:ring-opacity-50
+                  " %>
+            <%= error_tag f, :bearer %>
+            <p class="text-gray-600 text-xs italic">An Auth user JWT optionally signed with custom claims</p>
```
w3b6x9, line 121: @chasers if the only purpose is to pass in a JWT with custom claims then I would just be explicit and mention that

```suggestion
            <p class="text-gray-600 text-xs italic">A manually signed JWT with your custom claims</p>
```

or something to that effect.

assets/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "dependencies": {
-    "@supabase/realtime-js": "^2.1.0"
+    "@supabase/supabase-js": "^2.21.0"
```
w3b6x9, line 3: @chasers since you're already version bumping might as well version bump to the latest:

```suggestion
    "@supabase/supabase-js": "^2.24.0"
```

w3b6x9 (changes requested):
@chasers just making a note that the `Bearer` custom claims JWT token isn't working. It never shows up in the `realtime.subscription` table. Only the `Token` (Supabase API key like `anon`, `service_role`) shows up.

chasers:
✅ [docs updated](https://github.com/supabase/supabase/pull/15388)

PR #594, opened 2023-05-17
fix: token refresh

chasers:
## Problem
Clients can send over a new "access_token" which causes us to resubscribe to Postgres. Some clients seem to do this even if the access_token has not changed. With lots of clients connected this can cause extreme thrashing on the realtime.subscription table. 

## Solution
If the access_token in the socket is the same as the one they are trying to send over, don't do anything.


abc3 (approved):
🔥

PR #617, opened 2023-07-05
fix: Migration error for Cloud SQL

nkreiger:

Bug fix/Chore. 

The function realtime.list_changes is created, and sets the log_min_messages to 'fatal' for each session it runs.

## What is the new behavior?

The function realtime.list_changes is created, and if the your permissions are not a super use, it does not set the log_min_messages to 'fatal', but relies on the admin of the DB to set this at the appropriate level

## Additional context

This allows the migration to proceed for Cloud SQL setups that do not allow you to have a super user.

### DIFF
--- a/lib/extensions/postgres_cdc_rls/repo/migrations/20230328144023_create_list_changes_function.ex
+++ b/lib/extensions/postgres_cdc_rls/repo/migrations/20230328144023_create_list_changes_function.ex
@@ -4,68 +4,261 @@ defmodule Realtime.Extensions.Rls.Repo.Migrations.CreateListChangesFunction do
   use Ecto.Migration
 
   def change do
-    execute(
-      "create or replace function realtime.list_changes(publication name, slot_name name, max_changes int, max_record_bytes int)
-      returns setof realtime.wal_rls
-      language sql
-      set log_min_messages to 'fatal'
-    as $$
-      with pub as (
-        select
-          concat_ws(
-            ',',
-            case when bool_or(pubinsert) then 'insert' else null end,
-            case when bool_or(pubupdate) then 'update' else null end,
-            case when bool_or(pubdelete) then 'delete' else null end
-          ) as w2j_actions,
-          coalesce(
-            string_agg(
-              realtime.quote_wal2json(format('%I.%I', schemaname, tablename)::regclass),
-              ','
-            ) filter (where ppt.tablename is not null and ppt.tablename not like '% %'),
-            ''
-          ) w2j_add_tables
-        from
-          pg_publication pp
-          left join pg_publication_tables ppt
-            on pp.pubname = ppt.pubname
-        where
-          pp.pubname = publication
-        group by
-          pp.pubname
-        limit 1
-      ),
-      w2j as (
-        select
-          x.*, pub.w2j_add_tables
-        from
-          pub,
-          pg_logical_slot_get_changes(
-            slot_name, null, max_changes,
-            'include-pk', 'true',
-            'include-transaction', 'false',
-            'include-timestamp', 'true',
-            'include-type-oids', 'true',
-            'format-version', '2',
-            'actions', pub.w2j_actions,
-            'add-tables', pub.w2j_add_tables
-          ) x
-      )
-      select
-        xyz.wal,
-        xyz.is_rls_enabled,
-        xyz.subscription_ids,
-        xyz.errors
-      from
-        w2j,
-        realtime.apply_rls(
-          wal := w2j.data::jsonb,
-          max_record_bytes := max_record_bytes
-        ) xyz(wal, is_rls_enabled, subscription_ids, errors)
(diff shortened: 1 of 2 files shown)


lib/extensions/postgres_cdc_rls/repo/migrations/20230328144023_create_list_changes_function.ex
```diff
@@ -4,68 +4,132 @@ defmodule Realtime.Extensions.Rls.Repo.Migrations.CreateListChangesFunction do
...
-        and xyz.subscription_ids[1] is not null
-    $$;"
-    )
+    execute("DO $dyn$
+BEGIN
+    IF (SELECT rolsuper FROM pg_roles WHERE rolname = current_user) THEN
+        EXECUTE 'create or replace function realtime.list_changes(publication name, slot_name name, max_changes int, max_record_bytes int)
+          returns setof realtime.wal_rls
+          language sql
+          set log_min_messages to ''fatal''
```
olirice, line 13: if this is the only line of difference, can the if statement be moved into a separate posters query and add/remove this line using string interpolation?

olirice (changes requested):
would be best to make this a separate migration that `create or replaces` the function rather than editing an existing migration

w3b6x9:
@nkreiger thanks for your contribution!

can you also create a fresh migration file and move this in there?

w3b6x9 (changes requested):
see: https://github.com/supabase/realtime/pull/617#issuecomment-1624061380

nkreiger:
@w3b6x9 yep, I was going to actually comment that since its the only change maybe it could be broken out. I'll do this now

nkreiger:
@olirice @w3b6x9 I broke out the steps.

If I were to remove the `IF` statement entirely from the original migration, then the DB migration would still fail on that step with a permissions error, and I would get stuck there having to manually bump the migration passed it.

Maybe cleaner than the if statement if I just catch and ignore an exception? Let me know if thats preferable or you have alternative ideas.

olirice:
> If I were to remove the `IF` statement entirely from the original migration, then the DB migration would still fail on that step with a permissions error, and I would get stuck there having to manually bump the migration passed it.
> 
> Maybe cleaner than the if statement if I just catch and ignore an exception? Let me know if thats preferable or you have alternative ideas.

could you please provide the full error message you're seeing?

nkreiger:
@olirice sorry I had to pull it out because of the error and haven't had a chance to setup an example environment, will hopefully get back to you asap later this week or early next!

nkreiger:
@olirice apologies for the delay, this is the error:

```
03:36:35.796 project=realtime external_id=realtime [error] Error starting Postgres Extention: {:error,
 {:shutdown,
  {:failed_to_start_child, Extensions.PostgresCdcRls.Migrations,
   {%Postgrex.Error{
      message: nil,
      postgres: %{
        code: :insufficient_privilege,
        file: "guc.c",
        line: "8120",
        message: "permission denied to set parameter \"log_min_messages\"",
        pg_code: "42501",
        routine: "set_config_option_ext",
        severity: "ERROR",
        unknown: "ERROR"
      },
      connection_id: 254962,
      query: nil
    },
    [
      {Ecto.Adapters.SQL, :raise_sql_call_error, 1,
       [
         file: 'lib/ecto/adapters/sql.ex',
         line: 932,
         error_info: %{module: Exception}
       ]},
      {Enum, :"-map/2-lists^map/1-0-", 2, [file: 'lib/enum.ex', line: 1658]},
      {Ecto.Adapters.SQL, :execute_ddl, 4,
       [file: 'lib/ecto/adapters/sql.ex', line: 1024]},
      {Ecto.Migration.Runner, :log_and_execute_ddl, 3,
       [file: 'lib/ecto/migration/runner.ex', line: 352]},
      {Ecto.Migration.Runner, :"-flush/0-fun-1-", 6,
       [file: 'lib/ecto/migration/runner.ex', line: 117]},
      {Enum, :"-reduce/3-lists^foldl/2-0-", 3,
       [file: 'lib/enum.ex', line: 2468]},
      {Ecto.Migration.Runner, :flush, 0,
       [file: 'lib/ecto/migration/runner.ex', line: 116]},
      {Ecto.Migration.Runner, :perform_operation, 3,
       [file: 'lib/ecto/migration/runner.ex', line: 289]}
    ]}}}}
```

w3b6x9:
@nkreiger It's been awhile so just wanted to confirm we're on the same page:

1. need to update an existing migration file so it runs successfully
2. create another migration file replacing the function so that it doesn't set `log_min_messages` on every invocation

We're not going to do step 1 b/c it would change an existing migration file that's already been run on hosted and self-hosted Realtime projects.

We recommend that you fork this Realtime repo and add these necessary migrations to your version of Realtime so you're not blocked when using it against a Cloud SQL database.

PR #635, opened 2023-08-17
fix: Fixes flaky test on SubscriptionsCheckerTest

filipecabaco:

Tests for our SubscriptionChecker were flaky because of the way we checked the elements in the array.

Flaky test due to order of lists
Screenshot 2023-08-17 at 15 53 03


abc3:
@filipecabaco each PR requires increasing the version in ./mix.exs

abc3 (approved):
LGTM!

PR #638, opened 2023-08-18
fix: handle null access token

chasers:


abc3 (approved):

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -351,6 +351,15 @@ defmodule RealtimeWeb.RealtimeChannel do
     {:noreply, socket}
   end
 
+  def handle_in(
+        "access_token",
+        %{"access_token" => refresh_token},
+        %{assigns: %{access_token: _access_token}} = socket
+      )
+      when is_nil(refresh_token) do
+    {:noreply, socket}
```
w3b6x9, line 360: @chasers i highly recommend that we shut down the channel at this point like we did [below](https://github.com/supabase/realtime/pull/638/files#diff-2a3a48674821fef237a3451af5d33ca977ee5c7d2f6bc8ca18bd7f18b44a7a26R399). there shouldn't be any reason why any client lib should be sending updated nil access tokens but if they do we should err on the side of caution.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -452,14 +461,21 @@ defmodule RealtimeWeb.RealtimeChannel do
     {:reply, result, socket}
   end
 
-  def handle_in(_, _, socket) do
+  def handle_in(type, payload, socket) do
     socket = count(socket)
```
w3b6x9, line 465: @chasers how are we still confident in counting in this catchall given that it led to us counting incoming `nil` access tokens? we should seriously consider taking this out until we're fully confident in all the cases.

w3b6x9 (commented):
@chasers left you a few comments. cc @abc3

lib/realtime_web/channels/realtime_channel.ex
chasers: Pretty sure that would break everybody. @dshukertjr 's [example app](https://github.com/dshukertjr/realtime-debug/blob/main/src/pages/index.tsx) was sending nulls.

Can you check the js lib?

lib/realtime_web/channels/realtime_channel.ex
chasers: The counter is how we found this bug in the first place :D

Plus I added that explicit Logger message there now.

Ideally we have counters everywhere and one is billable and one is non-billable but we need to be able to support more metrics for that first.

lib/realtime_web/channels/realtime_channel.ex
chasers: https://github.com/supabase/realtime-js/blob/c9dc9c953cb6bb3e9440749d393e944b63a0a0f9/src/RealtimeClient.ts#L273

dshukertjr: I think I found why `accessToken` is null here. I opened a PR to fix it!
https://github.com/supabase/realtime-js/pull/245

chasers: @dshukertjr I assume you’re going to check the dart lib too?

dshukertjr: @chasers Yup! Once the PR on the JS library is merged, I will add the same implementation on the Dart library as well!

w3b6x9: @dshukertjr good find and appreciate the PR!

PR #639, opened 2023-08-19
feat: Add broadcast endpoint

filipecabaco:

Adds a batch broadcast endpoint so users can push multiple messages into multiple topics

## What is the new behavior?

New endpoint:
POST `/api` where a user can send:
```
{"messages": [ {"topic": "topic", "payload": "payload"} ]}
```

### DIFF
--- a/lib/realtime_web/controllers/broadcast_controller.ex
+++ b/lib/realtime_web/controllers/broadcast_controller.ex
@@ -1,42 +1,47 @@
 defmodule RealtimeWeb.BroadcastController do
   use RealtimeWeb, :controller
-  alias RealtimeWeb.Endpoint
+  use OpenApiSpex.ControllerSpecs
+
   alias Realtime.GenCounter
   alias Realtime.Tenants
+  alias Realtime.Tenants.BatchBroadcast
+  alias RealtimeWeb.Endpoint
+
+  alias RealtimeWeb.OpenApiSchemas.EmptyResponse
+  alias RealtimeWeb.OpenApiSchemas.TenantBatchParams
+  alias RealtimeWeb.OpenApiSchemas.UnprocessableEntityResponse
+
   action_fallback(RealtimeWeb.FallbackController)
 
-  defmodule Payload do
-    use Ecto.Schema
-    import Ecto.Changeset
-
-    embedded_schema do
-      embeds_many :messages, Message do
-        field(:topic, :string)
-        field(:payload, :map)
-      end
-    end
-
-    def changeset(payload, attrs) do
-      payload
-      |> cast(attrs, [])
-      |> cast_embed(:messages, required: true, with: &message_changeset/2)
-    end
-
-    def message_changeset(message, attrs) do
-      message
-      |> cast(attrs, [:topic, :payload])
-      |> validate_required([:topic, :payload])
-    end
-  end
+  operation(:broadcast,
+    summary: "Broadcasts a batch of messages",
+    parameters: [
+      token: [
+        in: :header,
+        name: "Authorization",
+        schema: %OpenApiSpex.Schema{type: :string},
+        required: true,
+        example:
+          "Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE2ODAxNjIxNTR9.U9orU6YYqXAtpF8uAiw6MS553tm4XxRzxOhz2IwDhpY"
+      ]
+    ],
+    request_body: TenantBatchParams.params(),
+    responses: %{
+      202 => EmptyResponse.response(),
+      403 => EmptyResponse.response(),
+      422 => UnprocessableEntityResponse.response()
+    }
+  )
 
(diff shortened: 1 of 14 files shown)


lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+      |> validate_required([:topic, :payload])
+    end
+  end
+
+  def broadcast(%{assigns: %{tenant: tenant}} = conn, attrs) do
+    with %Ecto.Changeset{valid?: true} = changeset <- Payload.changeset(%Payload{}, attrs),
+         %Ecto.Changeset{changes: %{messages: messages}} <- changeset,
+         requests_per_second_key <- Tenants.requests_per_second_key(tenant) do
+      for %{changes: %{topic: sub_topic, payload: payload}} <- messages do
+        tenant_topic = "#{tenant.external_id}:#{sub_topic}"
```
chasers, line 37: We should probably have an official helper for this somewhere consolidated.

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+        Endpoint.broadcast_from(self(), tenant_topic, "broadcast", payload)
+        GenCounter.add(requests_per_second_key)
```
chasers, line 39: This also should be handled in the rate limiter plug so we don't have to do this stuff in the controller.

test/realtime_web/controllers/broadcast_controller_test.exs
```diff
@@ -0,0 +1,128 @@
...
+      conn =
+        conn
+        |> put_req_header("accept", "application/json")
+        |> put_req_header("x-api-key", "potato")
+        |> then(&%{&1 | host: "potato.supabase.com"})
+
+      conn = post(conn, Routes.broadcast_path(conn, :broadcast), %{})
+      assert conn.status == 401
+    end
+  end
```
chasers, line 127: Can you add a test for the counter here also?

lib/realtime_web/plugs/auth_tenant.ex
```diff
@@ -0,0 +1,41 @@
...
+      [token] -> token
+      _ -> nil
+    end
+  end
+
+  defp fetch_api_key_param(conn) do
+    conn
+    |> Plug.Conn.fetch_query_params()
+    |> then(& &1.query_params)
+    |> Map.get("apikey")
```
chasers, line 37: Do we want to support this or only the `x-api-key` header?

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -0,0 +1,45 @@
+defmodule RealtimeWeb.BroadcastController do
+  use RealtimeWeb, :controller
+  alias RealtimeWeb.Endpoint
+  alias Realtime.GenCounter
+  alias Realtime.Tenants
+  action_fallback(RealtimeWeb.FallbackController)
+
+  defmodule Payload do
```
chasers, line 8: You may have modeled this from me including a schema in the LiveView for the form...

But I think this definitely warrants going into it's own module in `realtime` (not `realtime_web`).

The use of changesets here I think is v cool btw. An easy way to respond with a batch of errors.

test/realtime_web/controllers/broadcast_controller_test.exs
filipecabaco: test the counter is not incremented on bad requests?

lib/realtime_web/plugs/auth_tenant.ex
filipecabaco: will remove param support and focus on header support

lib/realtime_web/controllers/broadcast_controller.ex
filipecabaco: kept it here because I wasn't sure where to move it to 😅 

will move it to `realtime` and yes it simplified a lot the implementation of error handling

lib/realtime_web/controllers/broadcast_controller.ex
filipecabaco: will add one to `realtime/tenants.ex`

lib/realtime_web/plugs/assign_tenant.ex
```diff
@@ -23,8 +23,7 @@ defmodule RealtimeWeb.Plugs.AssignTenant do
          %Tenant{} = tenant <- Api.get_tenant_by_external_id(external_id) do
       tenant =
         tenant
-        |> tap(&GenCounter.new(Tenants.requests_per_second_key(&1)))
-        |> tap(&RateCounter.new(Tenants.requests_per_second_key(&1), idle_shutdown: :infinity))
+        |> tap(&initialize_counters/1)
```
chasers, line 26: 😍

chasers (approved):
🚀🚀🚀🚀

chasers (approved):
🚀🚀🚀🚀

PR #641, opened 2023-08-21
fix: Add rebalance code as an helper

filipecabaco:

Useful production code that runs ad-hoc added as an helper

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -217,11 +217,11 @@ defmodule Realtime.Helpers do
               # credo:disable-for-next-line
               IO.inspect({"Stopped", tenant, region})
             catch
-              kind, reason -> 
+              kind, reason ->
                 # credo:disable-for-next-line
                 IO.inspect({"Failed to stop", tenant, kind, reason})
-            end            
-            
+            end
+
             Process.sleep(1_500)
             acc + 1
           else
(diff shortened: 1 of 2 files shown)


lib/realtime/helpers.ex
```diff
@@ -204,6 +204,30 @@ defmodule Realtime.Helpers do
...
+            acc + 1
+          else
+            acc
+          end
+
+        _ ->
+          acc
+      end
+    end)
+  end
```
abc3, line 229: I've updated it a bit  

```suggestion
  def rebalance() do
    Enum.reduce(:syn.group_names(:users), 0, fn tenant, acc ->
      case :syn.lookup(Extensions.PostgresCdcRls, tenant) do
        {pid, %{region: region}} ->
          region = Realtime.PostgresCdc.aws_to_fly(region)
          launch_node = Realtime.PostgresCdc.launch_node(tenant, region, false)

          if launch_node && launch_node != node(pid) do
            try do
              Extensions.PostgresCdcRls.handle_stop(tenant, 5_000)
              IO.inspect({"Stopped", tenant, region})
            catch
              kind, reason -> IO.inspect({"Failed to stop", tenant, kind, reason})
            end            
            
            Process.sleep(1_500)
            acc + 1
          else
            acc
          end

        _ ->
          acc
      end
    end)
  end
```

abc3 (approved):
LGTM

PR #652, opened 2023-08-30
chore: Adapt script for AWS to use a IPV6 address

filipecabaco:

Due to the way we're setting up AWS deployment, we need to properly set the node long name.


chasers (approved):

chasers:
✅ assuming the test rerun check is okay and we bump the version.

Issue #618, opened 2023-07-12
Docker hub `latest` image tag is very old


liamkinne:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The `latest` tag is on a much older release on docker hub.

Causes errors using the newest docker-compose since it expects files to exists that don't.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. `docker pull supabase/realtime`
2. Using the docker-compose file in the repository.
3. `docker-compose up`
4. Errors due to missing files.

## Expected behavior

The `latest` tag points to the latest (stable?) release so when someone pulls they get the latest release.

## Screenshots

<img width="606" alt="Screenshot 2023-07-12 at 22 44 35" src="https://github.com/supabase/realtime/assets/10735142/4b708f7d-f549-43ab-bc08-171e5d4c960c">

## System information

N/A

ventsyv:
Just spent couple of hours scratching my head why the docker image is not matching the Dockerfile in the repo.

ventsyv:
BTW, the bitnami image seems to be more recent:
```
bitnami/supabase-realtime   latest    70d592e96c09   5 hours ago     115MB
supabase/realtime           latest    b6284e15dee3   12 months ago   195MB
```

filipecabaco:
Changed the CI cycle to properly release the latest version. Thank you for reporting 🙏

PR #657, opened 2023-09-01
Tag docker images with "latest"

ventsyv:
Tags the x86 and arm64 docker images with "latest" and "latest_arm" respectively.

Bug fix

Docker images are only tagged with the version number

## What is the new behavior?

In addition to the version tag, latest and latest_arm tags are added to the images.

## Additional context

I'm open to discussion as to what the tags should be. Maybe x86 and arm would be more appropriate?
Closes: https://github.com/supabase/realtime/issues/618

### DIFF
--- a/.github/workflows/prod_build.yml
+++ b/.github/workflows/prod_build.yml
@@ -85,11 +85,11 @@ jobs:
             network=host
 
       - id: build
-        uses: docker/build-push-action@v3, latest_arm
+        uses: docker/build-push-action@v3
         with:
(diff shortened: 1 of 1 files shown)


filipecabaco:
Hi @ventsyv 

First of all thank you for your contribution!

Just as a FYI tackled this has been tackled in a different PR ( https://github.com/supabase/realtime/pull/688 ) 

Ok to close this PR?

Kind regards,
Filipe

PR #662, opened 2023-09-07
fix: Bump phx to 1.7.7 to potentially reduce memory usage

filipecabaco:

https://github.com/phoenixframework/phoenix/issues/4933

We have some cowboy connections that seem to not be sweeped after a certain amount of time, this could be due to the issue found above which has a fix on phx 1.7.7

Other code changes are for warning removal.


.tool-versions
```diff
@@ -0,0 +1,3 @@
+elixir 1.14.3-otp-25
+nodejs 18.13.0
+erlang 25.3
```
abc3, line 3: why do we need this file?

filipecabaco: to use asdf to install the required versions to run realtime. makes it easier to onboard people to the project if need be.

with this and asdf we just need to do `asdf install` to have the dev environment setup

abc3: sounds good 👍

abc3 (approved):

PR #664, opened 2023-09-08
fix: Support AWS regions

filipecabaco:

Changes multiple elements to support AWS regions:
* Read replicas now point to the correct server and set the correct module name
* Replica repository is loaded dynamically and only when required to avoid extra modules to be loaded

### DIFF
--- a/lib/realtime/api.ex
+++ b/lib/realtime/api.ex
@@ -6,7 +6,13 @@ defmodule Realtime.Api do
 
   import Ecto.Query
 
-  alias Realtime.{Repo, Api.Tenant, Api.Extensions, RateCounter, GenCounter, Tenants}
+  alias Realtime.Repo
+  alias Realtime.Repo.Replica
+  alias Realtime.Api.Tenant
+  alias Realtime.Api.Extensions
+  alias Realtime.RateCounter
+  alias Realtime.GenCounter
+  alias Realtime.Tenants
 
   @doc """
   Returns the list of tenants.
@@ -18,7 +24,7 @@ defmodule Realtime.Api do
 
   """
   def list_tenants() do
-    repo_replica = Repo.replica()
+    repo_replica = Replica.replica()
 
     Tenant
     |> repo_replica.all()
@@ -33,7 +39,7 @@ defmodule Realtime.Api do
   * ordering (desc / asc)
   """
   def list_tenants(opts) when is_list(opts) do
-    repo_replica = Repo.replica()
+    repo_replica = Replica.replica()
 
     field = Keyword.get(opts, :order_by, "inserted_at") |> String.to_atom()
     external_id = Keyword.get(opts, :search)
@@ -71,7 +77,7 @@ defmodule Realtime.Api do
       ** (Ecto.NoResultsError)
 
   """
-  def get_tenant!(id), do: Repo.replica().get!(Tenant, id)
+  def get_tenant!(id), do: Replica.replica().get!(Tenant, id)
 
   @doc """
   Creates a tenant.
@@ -155,7 +161,7 @@ defmodule Realtime.Api do
 
   @spec get_tenant_by_external_id(String.t()) :: Tenant.t() | nil
   def get_tenant_by_external_id(external_id) do
-    repo_replica = Repo.replica()
+    repo_replica = Replica.replica()
 
     Tenant
     |> repo_replica.get_by(external_id: external_id)
@@ -167,7 +173,7 @@ defmodule Realtime.Api do
       where: e.type == ^type,
       select: e
     )
-    |> Repo.replica().all()
+    |> Replica.replica().all()
   end
 
   def rename_settings_field(from, to) do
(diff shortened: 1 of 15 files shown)


lib/realtime/api.ex
```diff
@@ -6,7 +6,13 @@ defmodule Realtime.Api do
...
   import Ecto.Query
 
-  alias Realtime.{Repo, Api.Tenant, Api.Extensions, RateCounter, GenCounter, Tenants}
+  alias Realtime.Repo
+  alias Realtime.Repo.Replica
+  alias Realtime.Api.Tenant
+  alias Realtime.Api.Extensions
+  alias Realtime.RateCounter
+  alias Realtime.GenCounter
+  alias Realtime.Tenants
```
abc3, line 15: I don't think this is necessary, realtime repo mainly uses a multi-alias style

lib/realtime/postgres_cdc.ex
```diff
@@ -60,23 +60,35 @@ defmodule Realtime.PostgresCdc do
...
+        "ap-southeast-2" -> "syd"
+        "ap-south-1" -> "syd"
+        "eu-west-1" -> "lhr"
+        "eu-west-2" -> "lhr"
+        "eu-west-3" -> "lhr"
+        "eu-central-1" -> "lhr"
+        _ -> nil
+      end
+    else
+      aws_region
```
abc3, line 91: Do we want to have RT nodes in all Supabase regions?

lib/realtime/repo_replica.ex
```diff
@@ -0,0 +1,75 @@
...
+      is_nil(replica_conf) ->
+        Logger.error("Replica config not found for #{region} region")
+        Realtime.Repo
+
+      true ->
+        # Check if module is present
+        case Code.ensure_compiled(replica) do
+          {:module, _} -> nil
+          _ -> {:module, _, _, _} = Module.create(replica, @ast, Macro.Env.location(__ENV__))
+        end
```
abc3, line 64: Is there a way to avoid dynamic creation of modules?

lib/realtime/api.ex
filipecabaco: This was the thing I mentioned during the sync that makes it trickier to find module usage within the code (e.g. with Realtime.Api.Tenant it becomes way easier to find aliases vs Api.Tenant in one module and Tenant in another module)

lib/realtime/postgres_cdc.ex
filipecabaco: doubt it and we could a check just to validate that we're in an allowed region

something along the lines of 
```elixir
aws_region in ["us-east-1", "eu-west-1", ...]
```

lib/realtime/repo_replica.ex
filipecabaco: trickier, we would need to define the region at compile time so we could properly load things up. with this change we're using runtime stuff which makes it trickier.

plus, with the current implementation, we are creating more repo modules as we are creating a Repo per region even if they are not used

lib/realtime/postgres_cdc.ex
abc3: `aws_region` is the region where the tenant's db resides. This function is responsible for initiating a database connection in the closest region. For example, if a tenant's database is located in aws's `us-east-1`, then we should establish a connection on a node in fly's `iad` region.
```elixir
"us-east-1" -> "iad"
```

lib/realtime/api.ex
abc3: many apps use a multi alias style, such as [phoenix](https://github.com/phoenixframework/phoenix/blob/1c1d50e36a9a293c9649c1649cae57dfa386e328/lib/phoenix/socket.ex#L201)

abc3: But, yeah, if it makes it harder to read, feel free to change that. However, I don't see any difference, just more lines

lib/realtime/postgres_cdc.ex
filipecabaco: Yeah and you're right for some reason I had the assumption that we were deploying in every Supabase tenant region and use said RDS conn.

Will check what can be done

lib/realtime/repo_replica.ex
```diff
@@ -0,0 +1,75 @@
...
+    "lhr" => Realtime.Repo.Replica.FRA,
+    "fra" => Realtime.Repo.Replica.FRA
+  }
+
+  @replicas_aws %{
+    "ap-southeast-1" => Realtime.Repo.Replica.Singapore,
+    "ap-southeast-2" => Realtime.Repo.Replica.Sydney,
+    "eu-west-1" => Realtime.Repo.Replica.Ireland,
+    "eu-west-2" => Realtime.Repo.Replica.London,
+    "us-east-1" => Realtime.Repo.Replica.NorthVirginia
```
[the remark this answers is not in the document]
filipecabaco: yes it was 😅

abc3 (approved):
LGTM

PR #676, opened 2023-09-14
fix: Set id based on host name instead of fly variable

filipecabaco:

Set id based on host name instead of fly variable

Currently we're using a variable set at the cluster level meaning that we will get repeated metrics for id ending up with the wrong metrics on metrics dashboards


abc3 (approved):
LGTM!

PR #677, opened 2023-09-18
feat: Connect to tenant database on channel join

filipecabaco:

To improve stability we'll connect to the Tenant database upon channel join with extra securities connection status of tenant database


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -691,4 +550,129 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        name: name,
+        user: user,
+        pass: password,
+        socket_opts: socket_opts,
+        pool: 1,
+        queue_target: 1000,
+        ssl_enforced: ssl_enforced
+      }
+
+      with {:ok, conn} <- Helpers.connect_db(opts),
```
chasers, line 654: We really can't do this here because every client is going to try and fire up it's own connection to the database when they attempt to join. 

We need a single process on the whole cluster which is started when the first client joins, and remains connected such that when subsequent clients join we know the database is healthy and can do a quick check of `:syn`.

This is basically the way the PostgresCdcRls and PostgresStream work. There is one database connection (tree) on the whole cluster which powers all the clients.

chasers: The reason we don't just use these is because a client may not want to stream changes from the database.

filipecabaco: ok will change things around to fit that, wanted to avoid a more complex solution but you are right an influx of 5k users trying to do this would kill everything 🤦 

will refactor this to fit that model with the following:
* One Genserver that on start checks the status of the tenant and keeps updating the status with `:syn`
* That Genserver should be started on the node nearest to the tenant database and identified by the tenant external id

lib/realtime_web/channels/user_socket.ex
```diff
@@ -25,10 +26,10 @@ defmodule RealtimeWeb.UserSocket do
       log_level =
         params
         |> Map.get("log_level", @default_log_level)
-        |> case do
+        |> then(fn
           "" -> @default_log_level
           level -> level
-        end
+        end)
```
abc3, line 32: Here's a simple comparison of different clauses, which is exactly what `case do` was designed for. Why would you want to wrap pattern matching in a fun?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -6,213 +6,97 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-  end
+  alias Realtime.GenCounter
+  alias Realtime.PostgresCdc
+  alias Realtime.RateCounter
+  alias Realtime.SignalHandler
+  alias Realtime.Tenants
+
+  alias RealtimeWeb.ChannelsAuthorization
+  alias RealtimeWeb.Endpoint
+  alias RealtimeWeb.Presence
```
abc3, line 23: Two lines of aliases have grown to eight. 

Hey, @supabase/dashbit, what's your recommendation on using [multi-aliases](https://elixir-lang.org/getting-started/alias-require-and-import.html#multi-aliasimportrequireuse)? Would you advise against aliasing multiple modules at once?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -513,15 +390,14 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        end
+      {:ok, %{avg: avg}} when avg < limits.max_joins_per_second ->
+        :ok
+
+      {:ok, %{avg: _}} ->
+        {:error, :too_many_joins}
 
       other ->
-        Logger.error("Unexpected error for #{tenant}: " <> inspect(other))
+        Logger.error("Unexpected error for " <> tenant <> ": " <> inspect(other))
```
abc3, line 400: Could it be possible that the tenant is not binary here?

lib/realtime_web/channels/realtime_channel.ex
abc3: Realtime initiates multiple connection pools to the tenant's db and I don't believe this check ensures that all of them will connect successfully. For example, the database could reach its maximum connection limit shortly after this check 🤷

Btw, it seems currently that `conn` is terminated only if the connection is successful. Also, it's better to use `GenServer.stop` for stopping, cos `conn` is an instance of the GenServer behavior

lib/realtime_web/channels/realtime_channel.ex
wojtekmach: We're definitely using multi-aliases on our projects. Nothing wrong with that. :)

The only concern is right now if you grep by `Realtime.GenCounter` you'll find this alias declaration but when it is changed to `alias Realtime.{GenCounter, ...}`, especially when it's spread across multiple lines, you no longer will be. FWIW, tools like ElixirLS, `mix xref`, "compilation tracers", etc are still able to see the "expanded" aliases.

I don't think it is a concern for most projects actually, I don't think it ever came up in projects I was directly involved in. There was an interesting twitter discussion about exactly this yesterday so you may want to check it out: https://twitter.com/ZacharyNDean/status/1705632083715019167.

As an aside and this is my personal opinion I often feel like we're using aliases too much. (My colleagues would sigh because I often would be the one that _did_ add a bunch of aliases.) It's fine to refer to modules by their full names. One easy thing people forget about is on apps (as opposed to libraries) the top-level name doesn't matter that much. If it's annoying (which I'd slightly agree with) to type `Realtime.Repo` and `RealtimeWeb.Endpoint`, I'd have called these `RT.Repo` and `RTWeb.Endpoint`. :)

abc3: @wojtekmach thank you for the explanation! We had a similar discussion to the one in the tweet above, so I decided to clarify this thing 🙏

lib/realtime_web/channels/realtime_channel.ex
chasers: lol this discussion will probably never end. I'm in the one-line-per-alias camp but at this point I don't particularly care. Should just setup a credo rule or something per project and just do what it says. @abc3 you can be the decider for Realtime :D

lib/realtime_web/channels/realtime_channel.ex
chasers: lol this discussion will probably never end. I'm in the one-line-per-alias camp but at this point I don't particularly care. Should just setup a credo rule or something per project and just do what it says. @abc3 you can be the decider for Realtime :D

filipecabaco: No strong opinions either. With other projects with thousands of files it became annoying to just ctrl+f an alias but we're not close to that 😆

> As an aside and this is my personal opinion I often feel like we're using aliases too much.

This!!! A lot of times when we're aliasing a module with two words it feels weird and not really necessary 

As for a Credo rule, there's one for the "option1" in that tweet as an optional https://hexdocs.pm/credo/Credo.Check.Readability.MultiAlias.html

lib/realtime_web/channels/user_socket.ex
filipecabaco: another one that would be a long discussion 😅

IMO this is a bit of an anti pattern as it leads to code to be "piled up" into the case while `then( fn ` will make it more "intuitive" to extract this to functions when the code requires. Plus it follow the idea of piping into functions

again, personal opinion which in this specific scenario I have no strong opinion about since it's a small amount of code.

lib/realtime/tenants/check.ex
```diff
@@ -0,0 +1,130 @@
...
+  def connection_status(tenant_id) do
+    case get_status(tenant_id) do
+      :undefined ->
+        :ok
+        node = Realtime.Nodes.get_node_for_tenant_id(tenant_id)
+
+        case :rpc.call(node, __MODULE__, :set_status, [tenant_id]) do
+          :ok ->
+            {_, res} = get_status(tenant_id)
+            res
```
abc3, line 17:
```suggestion
            case get_status(tenant_id) do
              {_, %{healthy?: true}} -> :ok
              {_, res} -> res
            end
```

lib/extensions/postgres_cdc_stream/cdc_stream.ex
```diff
@@ -43,21 +40,17 @@ defmodule Extensions.PostgresCdcStream do
...
   def get_manager_conn(id) do
-    Phoenix.Tracker.get_by_key(Stream.Tracker, "postgres_cdc_stream", id)
-    |> case do
-      [] ->
-        {:error, nil}
-
-      [{_, %{manager_pid: pid, conn: conn}}] ->
-        {:ok, pid, conn}
+    case Phoenix.Tracker.get_by_key(Stream.Tracker, "postgres_cdc_rls_stream", id) do
+      [] -> nil
```
chasers, line 46: I think you overwrote my last change from #682 here?

config/config.exs
```diff
@@ -29,7 +29,7 @@ config :realtime, :extensions,
   },
   postgres_cdc_stream: %{
     type: :postgres_cdc,
-    key: "postgres_cdc_stream",
+    key: "postgres_cdc_rls_stream",
```
chasers, line 32: `postgres_cdc_stream` is actually the NON RLS version (extension) so this should not change.

lib/extensions/postgres_cdc_stream/cdc_stream.ex
```diff
@@ -113,7 +102,7 @@ defmodule Extensions.PostgresCdcStream do
     Phoenix.Tracker.track(
       Stream.Tracker,
       self(),
-      "postgres_cdc_stream",
+      "postgres_cdc_rls_stream",
```
chasers, line 105: Same

lib/extensions/postgres_cdc_stream/tracker.ex
```diff
@@ -28,8 +28,8 @@ defmodule Extensions.PostgresCdcStream.Tracker do
     for {_topic, {_joins, leaves}} <- diff do
       for {id, _meta} <- leaves do
         Endpoint.local_broadcast(
-          "postgres_cdc:" <> id,
-          "postgres_cdc_down",
+          "postgres_cdc_rls:" <> id,
+          "postgres_cdc_rls_down",
```
chasers, line 32: Same

lib/realtime/tenants/check.ex
```diff
@@ -0,0 +1,130 @@
...
+
+  alias Realtime.Helpers
+
+  def connection_status(tenant_id) do
+    case get_status(tenant_id) do
+      :undefined ->
+        :ok
+        node = Realtime.Nodes.get_node_for_tenant_id(tenant_id)
+
+        case :rpc.call(node, __MODULE__, :set_status, [tenant_id]) do
```
chasers, line 14: So here the logic is?:
- `get_status` from syn
- if we don't have it yet - RPC to set set it via syn
- then `get_status` from syn again

If we don't have it already and you immediately look for it again you won't have it either because it's going to take some milliseconds to propagate to the caller node.

lib/realtime/tenants/check.ex
```diff
@@ -0,0 +1,130 @@
...
+        with {:ok, conn} <- Helpers.connect_db(opts) do
+          case Postgrex.query(conn, "SELECT 1", []) do
+            {:ok, _} -> {:ok, conn}
+            {:error, _} -> {:error, conn}
+          end
+        end
+      end)
+      # This makes the connection fail
+      |> tap(fn res ->
+        Enum.each(res, fn {_, conn} -> Process.exit(conn, :normal) end)
```
chasers, line 121: No real reason to exit here lets just keep it up this way there will be less traffic for syn to handle and less cases for race conditions.

lib/realtime/tenants/check.ex
```diff
@@ -0,0 +1,130 @@
...
+
+  def init(state), do: {:ok, state, {:continue, :setup_syn}}
+
+  def handle_continue(:setup_syn, state) do
+    :ok = :syn.add_node_to_scopes([__MODULE__])
+    {:noreply, state}
+  end
+
+  def handle_info({:set_status, tenant_id}, state) do
+    res = check_tenant_connection(tenant_id)
```
chasers, line 55: And you don't really actually have to do this either.

If this GenServer starts the Postgrex connection then we know it's alive if this GenServer is alive.

And if we register this GenServer with :syn then we know the connection is alive and this info is distributed for us.

lib/realtime/application.ex
```diff
@@ -42,6 +42,11 @@ defmodule Realtime.Application do
       name: Realtime.Registry.Unique
     )
 
+    Registry.start_link(
+      keys: :unique,
+      name: Realtime.Registry.Tenant
```
chasers, line 47: I don't think we need a separate local process registry for this because syn is our process registry.

lib/realtime/tenants/check.ex
```diff
@@ -0,0 +1,130 @@
...
+
+  require Logger
+
+  alias Realtime.Helpers
+
+  def connection_status(tenant_id) do
+    case get_status(tenant_id) do
+      :undefined ->
+        :ok
+        node = Realtime.Nodes.get_node_for_tenant_id(tenant_id)
```
chasers, line 12: This isn't going to work because `get_node_for_tenant_id` ends up calling `Tenants.get_tenant_by_external_id`.

Anywhere in the join path you need to use `Tenants.Cache.get_tenant_by_external_id` otherwise we hit our db too much when we get a flood of reconnects.

lib/realtime/tenants/check.ex
chasers: Also this is going to blow up when you have a bunch of tenants connect at the same time and all of them hit this handle info where this function could take a long time.

lib/realtime/application.ex
```diff
@@ -62,7 +67,8 @@ defmodule Realtime.Application do
         RealtimeWeb.Presence,
         {Task.Supervisor, name: Realtime.TaskSupervisor},
         Realtime.Latency,
-        Realtime.Telemetry.Logger
+        Realtime.Telemetry.Logger,
+        Realtime.Tenants.Check
```
chasers, line 71: This needs to be like a DynamicSupervisor where you start a Check on the correct node for a tenant.

And you can start is with a `name` `via` such that it registers with :syn.

Then if Check is alive for a tenant and we have it in syn then we know the db is okay.

lib/extensions/postgres_cdc_stream/cdc_stream.ex
filipecabaco: reverting to what is in main

lib/realtime/application.ex
filipecabaco: you are right, I think this are leftovers of other experiments, deleting

lib/realtime/application.ex
filipecabaco: Yes, will add the clustering later, just getting the childspec make sense

lib/realtime/tenants/check.ex
filipecabaco: my plan would be to retry with a backoff (logic is underneath) until we have the information in :syn since we're working with eventually consistency 😓

lib/realtime/tenants/check.ex
filipecabaco: On this Check.exs Genserver we connect the node to the database and won't stop the connection correct?

Should we keep track of the health status and pid of the postgrex conn in the metadata then?

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: from what I gather it seems it's possible, will add a safeguard

filipecabaco: actually will remove the tenant from there since we have that in the metadata of the logs

lib/realtime/tenants/check.ex
chasers: >we connect the node to the database and won't stop the connection correct?

Correct.

>Should we keep track of the health status and pid of the postgrex conn in the metadata then?

You can add metadata to a process in syn but if we register it with syn the fact that it's in syn should mean it's alive and healthy.

lib/realtime/tenants/check.ex
chasers: Yeah that should work. Didn't fully look at how your backoff stuff works yet honestly.

lib/realtime/tenants/check.ex
filipecabaco: Will change 👍

lib/realtime_web/channels/user_socket.ex
abc3: I don't want to start a holy war, I'm just curious because you replaced the existing code. I thought you had a strong opinion on this. From my perspective, doing pattern matching inside a lambda is actually an anti-pattern when Elixir already provides specific constructs for this purpose

josevalim: For what is worth, and not discussing style, the `case do` is guaranteed to always be faster as it doesn't require creating a function and then invoking it. We may be able to optimize the function one, but it is not guaranteed. :)

josevalim: That said, my comment above is explicitly about `then + fn` vs `case`. It is also completely fine to pattern match inside functions. :)

filipecabaco: Ny opinion is more on the side that pipes handle functions and as such then+fn makes more "sense" with that idea in mind

Plus it makes it easier to guide the person to the discussion of "should I extract it to a private function?" 

No strong opinions either 😁 whatever makes the team happier I'm in but this is my rationale

filipecabaco:
FYI: Clustering of the Genserver to be done in a later stagein this PR, just making sure the childspec makes sense

lib/realtime/syn_handler.ex
```diff
@@ -0,0 +1,78 @@
...
+  """
+  require Logger
+  alias RealtimeWeb.Endpoint
+
+  def on_process_unregistered(mod, name, _pid, _meta, reason) do
+    Logger.warn("#{mod} terminated: #{inspect(name)} #{node()}")
+    topic = topic(mod)
+
+    if reason != :syn_conflict_resolution do
+      Endpoint.local_broadcast(topic <> ":" <> name, topic <> "_down", nil)
```
abc3, line 13: The `realtime_channel` reconnects after receiving the `postgres_cdc_down` event. However, after this change, the event will be different: `postgres_cdc_rls_down` for RLS and "postgres_cdc_stream_down" for stream

lib/realtime/tenants/connect.ex
```diff
@@ -0,0 +1,121 @@
+defmodule Realtime.Tenants.Connect do
+  use GenServer
```
abc3, line 2: a friendly reminder to add @moduledoc, even if it's set to false

lib/realtime/tenants/connect.ex
```diff
@@ -0,0 +1,121 @@
...
+    {:noreply, state}
+  end
+
+  defp call_external_node(tenant_id) do
+    with tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_id),
+         {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant),
+         :ok <- :rpc.call(node, __MODULE__, :set_status, [tenant_id]) do
+      get_status(tenant_id)
+    end
+  end
```
abc3, line 55: if a tenant's region is missing in `Realtime.Nodes.region_mapping/2`, different nodes may keep the status for the same tenant

lib/realtime/tenants/connect.ex
```diff
@@ -0,0 +1,121 @@
...
+      _ ->
+        :ok
+    end
+  end
+
+  defp update_syn_with_conn_check(res, tenant_id) do
+    case res do
+      {:ok, conn} -> :syn.register(__MODULE__, tenant_id, self(), %{conn: conn})
+      {:error, _} -> :syn.register(__MODULE__, tenant_id, self(), %{conn: nil})
+    end
```
abc3, line 75:
```suggestion
    conn = case res do
      {:ok, conn} -> conn
      {:error, _} -> nil
    end
    
    :syn.register(__MODULE__, tenant_id, self(), %{conn: conn})
```

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -6,213 +6,97 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-    socket = socket |> assign_access_token(params) |> assign_counter()
+    socket =
+      socket
+      |> assign_access_token(params)
+      |> assign_counter()
 
     start_db_rate_counter(tenant)
 
     with false <- SignalHandler.shutdown_in_progress?(),
+         {:ok, _} <- Connect.connection_status(tenant),
```
abc3, line 48: Am I correct in understanding that db availability is checked only once when `:syn.lookup(__MODULE__, tenant_id)` returns undefined?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -236,15 +120,15 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+    socket = maybe_log_handle_info(socket, msg)
 
     push(socket, "presence_state", presence_dirty_list(topic))
 
     {:noreply, socket}
   end
 
   @impl true
-  def handle_info(%{event: "postgres_cdc_down"}, socket) do
+  def handle_info(%{event: "postgres_cdc_rls_down"}, socket) do
```
abc3, line 131: as I mentioned above, the stream version also can send `postgres_cdc_down`

lib/realtime_web/channels/user_socket.ex
abc3: @josevalim thanks for the additional comments. I apologize if that irrelevant discussion distracted you :D

lib/realtime/syn_handler.ex
abc3: btw, `Realtime.Tenants.Connect` should also handle tenant termination, because it needs to unregister the syn record and close the connection to db

filipecabaco: will change 👍

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: yes, we will check once that we are able to connect. this is actually a problem since I'm setting syn also on connect error and that shouldn't happen because that way if one user, on join, is unable to connect a second user will retry it and check again 👀 

this in a way could be the retry policy "retry if syn is undefined" and "success if syn is set". this will simplify the full logic and was something Chasers mentioned so I will go with that 😄

lib/realtime_web/channels/user_socket.ex
josevalim: Nah, no worries. It is still good language feedback. :)

lib/realtime/syn_handler.ex
filipecabaco: posted a question in slack for clarification regarding this change

filipecabaco: clarification made, will handle here as initially planned

lib/realtime/nodes.ex
```diff
@@ -0,0 +1,130 @@
...
+  Handles common needs for :syn module operations
+  """
+  require Logger
+  alias Realtime.Tenants
+  alias Realtime.Api.Tenant
+
+  @doc """
+  Gets the node to launch the Postgres connection on for a tenant id.
+  """
+  @spec get_node_for_tenant_id(String.t()) :: node()
```
josevalim, line 12: This spec does not match the implementation fully. It may return `{:error, ...}` or even `nil` (which is a valid node typing wise but never in practice).

lib/realtime/nodes.ex
```diff
@@ -0,0 +1,130 @@
...
+
+      [] ->
+        Logger.warning("Zero region nodes for #{region} using #{inspect(default)}")
+        default
+
+      regions_nodes ->
+        member_count = Enum.count(regions_nodes)
+        index = :erlang.phash2(tenant_id, member_count)
+
+        Enum.at(regions_nodes, index)
```
josevalim, line 127:
```suggestion
        Enum.fetch!(regions_nodes, index)
```

lib/realtime/tenants/connect.ex
```diff
@@ -0,0 +1,90 @@
...
+      case res do
+        {:ok, conn} ->
+          :syn.update_registry(__MODULE__, tenant_id, fn _pid, meta -> %{meta | conn: conn} end)
+          state = %{state | db_conn_reference: Process.monitor(conn)}
+
+          {:ok, state}
+
+        {:error, error} ->
+          Logger.error("Error connecting to tenant database: #{inspect(error)}")
+          {:stop, :normal, state}
```
josevalim, line 61: I don't believe this is a valid return type. You can use `{:stop, :normal}` or `:ignore`. Although you should check what happens based on your supervisor and restart strategies. I would also manually double-check your choice of supervisor strategy + genserver restart is fine with this return type (some options means it could potentially retry or keep the child specification in the supervisor).

josevalim (commented):
LGTM! Some comments added. :)

lib/realtime/tenants/connect.ex
filipecabaco: this was a wrong copy paste I believe 😓 we just need `{:stop, :normal}` in this scenario so will use that 👍

chasers (approved):

PR #678, opened 2023-09-19
fix: Take version into consideration in Postgres libcluster Strategy

filipecabaco:

Currently we have nodes connecting from different versions which could have a negative impact on the overall cluster operations. To avoid it, we're taking into consideration the version when connecting nodes via libcluster Postgres strategy.


lib/realtime/cluster_strategy/postgres.ex
```diff
@@ -31,10 +31,15 @@ defmodule Realtime.Cluster.Strategy.Postgres do
       parameters: Keyword.fetch!(state.config, :parameters)
     ]
 
+    version =
+      Application.spec(:realtime)[:vsn]
+      |> to_string()
```
abc3, line 36: nitpick:
```suggestion
      "#{Application.spec(:realtime)[:vsn]}"
```

abc3 (approved):

PR #681, opened 2023-09-22
fix: Refactor Replication Poller to improve readability

filipecabaco:

Refactor Replication Poller to improve readability

### DIFF
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -8,24 +8,16 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
 
   require Logger
 
-  import Realtime.Helpers,
-    only: [cancel_timer: 1, decrypt_creds: 5, default_ssl_param: 1, maybe_enforce_ssl_config: 2]
+  import Realtime.Helpers, only: [cancel_timer: 1, default_ssl_param: 1, connect_db: 10]
 
   alias DBConnection.Backoff
-
-  alias Extensions.PostgresCdcRls.MessageDispatcher
-  alias Extensions.PostgresCdcRls.Replications
-
-  alias Realtime.Adapters.Changes.DeletedRecord
-  alias Realtime.Adapters.Changes.NewRecord
-  alias Realtime.Adapters.Changes.UpdatedRecord
+  alias Extensions.PostgresCdcRls.{Replications, MessageDispatcher}
+  alias Realtime.Adapters.Changes.{DeletedRecord, NewRecord, UpdatedRecord}
   alias Realtime.PubSub
 
   @queue_target 5_000
 
-  def start_link(opts) do
-    GenServer.start_link(__MODULE__, opts)
-  end
+  def start_link(opts), do: GenServer.start_link(__MODULE__, opts)
 
   @impl true
   def init(args) do
@@ -39,7 +31,10 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
         args["db_user"],
         args["db_password"],
         args["db_socket_opts"],
-        ssl_enforced
+        1,
+        @queue_target,
+        ssl_enforced,
+        "realtime_rls"
       )
 
     tenant = args["id"]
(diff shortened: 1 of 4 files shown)


lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -99,76 +94,27 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
-          publication,
-          max_changes,
-          max_record_bytes
-        ])
-
-      Realtime.Telemetry.execute(
-        [:realtime, :replication, :poller, :query, :stop],
-        %{duration: time},
-        %{tenant: tenant}
+    broadcast_count =
```
chasers, line 97: I feel like this whole thing including `list_changes_with_telemetry` could be a with statement with each step making it very obvious what's going on.

chasers: Like break out what's going in `handle_list_changes_result`. Responding with the count of changes handled seems odd.

filipecabaco: will do 👍 that is the main priority for the code cleanup: be more clear on the flow of actions happening and that is a great suggestion

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -99,79 +94,21 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
-          else
-            Process.send_after(self(), :poll, poll_interval_ms)
-          end
-
-        {:noreply, %{state | backoff: backoff, poll_ref: poll_ref}}
+    with args <- [conn, slot_name, publication, max_changes, max_record_bytes],
+         {time, list_changes} <- :timer.tc(Replications, :list_changes, args),
+         _ <- record_list_changes_telemetry(time, tenant),
+         {:ok, row_count} <- handle_list_changes_result(list_changes, tenant),
+         backoff <- Backoff.reset(backoff) do
```
josevalim, line 101: This `with` merges clauses can fail with the ones that can never fail, making it harder to spot the failure cases. My suggestion would be to try something like:

```elixir
    args = [conn, slot_name, publication, max_changes, max_record_bytes]
    {time, list_changes} <- :timer.tc(Replications, :list_changes, args)
    record_list_changes_telemetry(time, tenant)

    case handle_list_changes_result(list_changes, tenant) do
        {:ok, row_count} ->
          backoff = Backoff.reset(backoff)
```

Making it clear where the actual logic is. :)

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -220,6 +152,93 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
+    |> Enum.reduce([], fn row, acc ->
+      columns
+      |> Enum.zip(row)
+      |> generate_record()
+      |> then(fn
+        nil -> acc
+        record_struct -> [record_struct | acc]
+      end)
+    end)
+    |> Enum.reverse()
```
josevalim, line 230: A `flat_map` is very hand where you want to map something but skip certain entries. To skip, you can return `[]`:

```suggestion
    |> Enum.flat_map(fn row ->
      columns
      |> Enum.zip(row)
      |> generate_record()
      |> List.wrap()
    end)
```

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -220,6 +152,93 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
+    |> Enum.each(fn change ->
+      topic = "realtime:postgres:" <> tenant
+      Phoenix.PubSub.broadcast_from(PubSub, self(), topic, change, MessageDispatcher)
+    end)
```
josevalim, line 234: You can even wrap the whole thing a comprehension if you want:

```suggestion
    for row <- rows,
        change <- columns |> Enum.zip(row) |> generate_record() |> List.wrap() do
      topic = "realtime:postgres:" <> tenant
      Phoenix.PubSub.broadcast_from(PubSub, self(), topic, change, MessageDispatcher)
    end)
```

josevalim (commented):
Hi @filipecabaco! Looking good, I have dropped some additional suggestions. :)

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -45,12 +45,7 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
     tenant = args["id"]
 
     state = %{
-      backoff:
-        Backoff.new(
-          backoff_min: 100,
-          backoff_max: 5_000,
-          backoff_type: :rand_exp
-        ),
+      backoff: Backoff.new(backoff_min: 100, backoff_max: 5_000, backoff_type: :rand_exp),
```
abc3, line 48: Since we started refactoring `ReplicationPooler`, let's replace `Backoff` with just `Enum.random/1`. Backoff is used here only to get a random timeout value

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -99,79 +94,21 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
+         _ <- record_list_changes_telemetry(time, tenant),
+         {:ok, row_count} <- handle_list_changes_result(list_changes, tenant),
+         backoff <- Backoff.reset(backoff) do
+      pool_ref =
+        if row_count == 0 do
+          send(self(), :poll)
+          nil
+        else
+          Process.send_after(self(), :poll, poll_interval_ms)
+        end
```
abc3, line 108: Here, if the rows are not empty, we should poll immediately to retrieve the rest of the changes. If there are no rows, we can wait

```suggestion
        if row_count > 0 do
          send(self(), :poll)
          nil
        else
          Process.send_after(self(), :poll, poll_interval_ms)
        end
```

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -220,6 +152,93 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
+      database: name,
+      password: pass,
+      username: user,
+      queue_target: @queue_target,
+      parameters: [application_name: "realtime_rls"],
+      socket_options: socket_opts
+    ]
+    |> maybe_enforce_ssl_config(ssl_enforced)
+    |> Postgrex.start_link()
+  end
```
abc3, line 185: I think it can be removed in favor of `Realtime.Helpers.connect_db`

lib/extensions/postgres_cdc_rls/replication_poller.ex
filipecabaco: will tackle this in another round, already a lot of changes happening

abc3 (approved):
LGTM 🔥

PR #684, opened 2023-09-26
fix: Change payload from Broadcast endpoint

filipecabaco:

Current format was not compliant with what was expected by our channel clients


abc3 (approved):
LGTM

PR #685, opened 2023-09-27
fix: Add Rate limit to Broadcast endpoint

filipecabaco:

To prevent over usage by customers, we will take into account the Rate Limits.

We handle two situations:
* The existing batch will trigger a rate limit so we advice to reduce the batch size
* Rate limit has been hit so we can't process the request

### DIFF
--- a/lib/realtime_web/controllers/broadcast_controller.ex
+++ b/lib/realtime_web/controllers/broadcast_controller.ex
@@ -3,7 +3,9 @@ defmodule RealtimeWeb.BroadcastController do
   use RealtimeWeb, :controller
   use OpenApiSpex.ControllerSpecs
 
+  alias Realtime.Api.Tenant
   alias Realtime.GenCounter
+  alias Realtime.RateCounter
   alias Realtime.Tenants
   alias Realtime.Tenants.BatchBroadcast
   alias RealtimeWeb.Endpoint
@@ -11,6 +13,7 @@ defmodule RealtimeWeb.BroadcastController do
   alias RealtimeWeb.OpenApiSchemas.EmptyResponse
   alias RealtimeWeb.OpenApiSchemas.TenantBatchParams
   alias RealtimeWeb.OpenApiSchemas.UnprocessableEntityResponse
+  alias RealtimeWeb.OpenApiSchemas.TooManyRequestsResponse
 
   action_fallback(RealtimeWeb.FallbackController)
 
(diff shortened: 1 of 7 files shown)


abc3 (approved):
LGTM, just want to say that there are failed tests

chasers:
Mmmm ... we're already have this logic in the rate limiter plug `RealtimeWeb.Plugs.RateLimiter`.

We should be doing this there.

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -30,26 +33,45 @@ defmodule RealtimeWeb.BroadcastController do
...
+         :ok <- check_rate_limit(events_per_second_key, tenant, length(messages)) do
       for %{changes: %{topic: sub_topic, payload: payload, event: event}} <- messages do
         tenant_topic = Tenants.tenant_topic(tenant, sub_topic)
         payload = %{"payload" => Map.merge(payload, %{"event" => event})}
 
         Endpoint.broadcast_from(self(), tenant_topic, "broadcast", payload)
 
         GenCounter.add(events_per_second_key)
-        Logger.info("Broadcasted message to using HTTP API: " <> tenant_topic)
+        Logger.info("Broadcasted message using HTTP API to topic " <> tenant_topic)
```
chasers, line 54: Thought you were only adding this for testing. We need to pull it out we don't want every Broadcast request being logged here.

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -30,26 +33,45 @@ defmodule RealtimeWeb.BroadcastController do
...
       end
 
       send_resp(conn, :accepted, "")
     end
   end
+
+  defp check_rate_limit(events_per_second_key, %Tenant{} = tenant, total_messages_to_broadcast) do
```
chasers, line 61: Can we consolidate this logic in `RealtimeWeb.Plugs.RateLimiter`?

filipecabaco:
TIL we had `RealtimeWeb.Plugs.RateLimiter` 

That will make it trickier to check the batch one 🤔 will refactor now

chasers:
If it comes after the JSON parser we should have the JSON list in the conn to count like you were doing.

filipecabaco:
Sent a message to Slack regarding this discussion since it might require more checks.

Issue is that we might want to use a different rate limit key

filipecabaco:
For public record sake 😄 

The keys we're using to check rate limiting on broadcast isn't the same as the rate limiter plug as such we can't properly merge the logic as expected.

We will revise the Plug logic and this endpoints logic to better abstract the rate limit keys at a later stage

PR #701, opened 2023-10-17
fix: Disconnect when tenant has no users

filipecabaco:

To ensure that we limit the amount of connections to a tenant database, we will check if that tenant has any users and if not, we kill the database connection.

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -2,23 +2,33 @@ defmodule Realtime.Tenants.Connect do
   @moduledoc """
   This module is responsible for attempting to connect to a tenant's database and store the DBConnection in a Syn registry.
   """
-  use GenServer
+  use GenServer, restart: :transient
 
   require Logger
 
   alias Realtime.Helpers
   alias Realtime.Tenants
+  alias Realtime.UsersCounter
 
-  defstruct tenant_id: nil, db_conn_reference: nil
+  @erpc_timeout_default 5000
+  @check_connected_user_interval_default 1000
+  @connected_users_bucket_shutdown [0, 0, 0, 0, 0, 0]
+
+  defstruct tenant_id: nil,
+            db_conn_reference: nil,
+            db_conn_pid: nil,
+            check_connected_user_interval: nil,
+            connected_users_bucket: [1]
 
   @doc """
   Returns the database connection for a tenant. If the tenant is not connected, it will attempt to connect to the tenant's database.
   """
-  @spec lookup_or_start_connection(binary()) :: {:ok, DBConnection.t()} | {:error, term()}
-  def lookup_or_start_connection(tenant_id) do
+  @spec lookup_or_start_connection(binary(), keyword()) ::
+          {:ok, DBConnection.t()} | {:error, term()}
+  def lookup_or_start_connection(tenant_id, opts \\ []) do
     case get_status(tenant_id) do
       {:ok, conn} -> {:ok, conn}
-      {:error, :tenant_database_unavailable} -> call_external_node(tenant_id)
+      {:error, :tenant_database_unavailable} -> call_external_node(tenant_id, opts)
       {:error, :initializing} -> {:error, :tenant_database_unavailable}
     end
   end
@@ -26,7 +36,6 @@ defmodule Realtime.Tenants.Connect do
   @doc """
   Returns the database connection pid from :syn if it exists.
   """
-
   @spec get_status(binary()) ::
           {:ok, DBConnection.t()} | {:error, :tenant_database_unavailable | :initializing}
   def get_status(tenant_id) do
@@ -40,10 +49,10 @@ defmodule Realtime.Tenants.Connect do
   @doc """
   Connects to a tenant's database and stores the DBConnection in the process :syn metadata
   """
-  @spec connect(binary()) :: {:ok, DBConnection.t()} | {:error, term()}
-  def connect(tenant_id) do
+  @spec connect(binary(), keyword()) :: {:ok, DBConnection.t()} | {:error, term()}
+  def connect(tenant_id, opts \\ []) do
     supervisor = {:via, PartitionSupervisor, {Realtime.Tenants.Connect.DynamicSupervisor, self()}}
-    spec = {__MODULE__, tenant_id: tenant_id}
+    spec = {__MODULE__, [tenant_id: tenant_id] ++ opts}
 
     case DynamicSupervisor.start_child(supervisor, spec) do
       {:ok, _} -> get_status(tenant_id)
@@ -52,9 +61,20 @@ defmodule Realtime.Tenants.Connect do
     end
   end
 
-  def start_link(tenant_id: tenant_id) do
+  def start_link(opts) do
+    tenant_id = Keyword.get(opts, :tenant_id)
+
+    check_connected_user_interval =
+      Keyword.get(opts, :check_connected_user_interval, @check_connected_user_interval_default)
+
     name = {__MODULE__, tenant_id, %{conn: nil}}
-    GenServer.start_link(__MODULE__, %__MODULE__{tenant_id: tenant_id}, name: {:via, :syn, name})
+
+    state = %__MODULE__{
+      tenant_id: tenant_id,
+      check_connected_user_interval: check_connected_user_interval
+    }
+
+    GenServer.start_link(__MODULE__, state, name: {:via, :syn, name})
   end
 
   ## GenServer callbacks
@@ -66,15 +86,50 @@ defmodule Realtime.Tenants.Connect do
       case res do
         {:ok, conn} ->
           :syn.update_registry(__MODULE__, tenant_id, fn _pid, meta -> %{meta | conn: conn} end)
-          state = %{state | db_conn_reference: Process.monitor(conn)}
+          state = %{state | db_conn_reference: Process.monitor(conn), db_conn_pid: conn}
 
-          {:ok, state}
+          {:ok, state, {:continue, :setup_connected_users}}
 
         {:error, error} ->
           Logger.error("Error connecting to tenant database: #{inspect(error)}")
           {:stop, :normal}
       end
     end
+  end
+
+  @impl GenServer
+  def handle_continue(
+        :setup_connected_users,
+        %{
+          check_connected_user_interval: check_connected_user_interval,
+          connected_users_bucket: connected_users_bucket
+        } = state
+      ) do
+    send_connected_user_check_message(connected_users_bucket, check_connected_user_interval)
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_info(
+        :check_connected_users,
+        %{
+          tenant_id: tenant_id,
+          check_connected_user_interval: check_connected_user_interval,
+          connected_users_bucket: connected_users_bucket
+        } = state
+      ) do
+    connected_users_bucket =
+      tenant_id
+      |> update_connected_users_bucket(connected_users_bucket)
+      |> send_connected_user_check_message(check_connected_user_interval)
+
+    {:noreply, %{state | connected_users_bucket: connected_users_bucket}}
+  end
+
+  def handle_info(:shutdown, %{db_conn_pid: db_conn_pid} = state) do
+    Logger.info("Tenant has no connected users, database connection will be terminated")
+    :ok = GenServer.stop(db_conn_pid)
+    {:stop, :normal, state}
   end
 
   @impl GenServer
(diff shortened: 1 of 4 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -78,6 +87,21 @@ defmodule Realtime.Tenants.Connect do
...
 
   @impl GenServer
+  def handle_continue(:setup_connected_users, state) do
+    %{check_connected_user_interval: check_connected_user_interval} = state
+    Process.send_after(self(), :check_connected_users, check_connected_user_interval)
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_info(:check_connected_users, %{tenant_id: tenant_id} = state) do
```
chasers, line 97: Needs to continue checking on the interval.

lib/realtime/tenants/connect.ex
```diff
@@ -78,6 +87,21 @@ defmodule Realtime.Tenants.Connect do
...
+    if UsersCounter.tenant_users(tenant_id) > 0 do
```
chasers, line 98: Here.

chasers: And also only if it's been zero for greater than one minute.

lib/realtime/tenants/connect.ex
filipecabaco: 🤦‍♂️ ...

filipecabaco: fixed 👍

lib/realtime/tenants/connect.ex
filipecabaco: that is done since we're using `send_after` instead of `send` so the first check will only happen after `check_connected_user_interval`

filipecabaco: will add a test to reflect said use case

filipecabaco: changed test:
```

    test "if no users are connected to a tenant channel, stop the connection", %{
      tenant: %{external_id: tenant_id}
    } do
      Connect.lookup_or_start_connection(tenant_id, check_connected_user_interval: 500)
      assert {_, %{conn: _}} = :syn.lookup(Connect, tenant_id)

      # Not enough time has passed, connection still alive
      :timer.sleep(100)
      assert {_, %{conn: _}} = :syn.lookup(Connect, tenant_id)

      # Enough time has passed, connection stopped
      :timer.sleep(500)
      assert :undefined = :syn.lookup(Connect, tenant_id)
    end
```

lib/realtime/tenants/connect.ex
josevalim: There are two race conditions in this code:

1. What happens if someone attempts to connect at the time you are disconnecting? The process will die and the user will likely get a connection error.

2. You may also run this check at the exact moment someone is reconnecting, so you have zero clients, but then you will have to immediately start the process again. This may not be a major issue though. It just means you will shut down a process to immediately start again.

I am not well versed to fully validate a solution but here are ideas for each.

## Problem 1

1. Instead of immediately returning {:stop, :normal}, instead, remove yourself from the registry

2. Then, in 5 seconds, if someone is connected directly to you, you add yourself back. This may be tricky because you use `syn` to measure how many connections you have, and you may need other mechanisms (like monitoring clients)

## Problem 2

1. Instead of immediately disconnecting, ping every 30 seconds. In the first 30 seconds, if there is nobody connected, you will set a  "will_disconnect" to true for the next ping. Otherwise the "will_disconnect" is false. If in the second ping "will_disconnect" is true, then you stop. This can be trivially done by changing the message itself to `{:check_connected_users, will_disconnect}`. But, this is not a major issue, it is free to choose code simplicity.

filipecabaco: Made the change to have that double check. 

Also took the opportunity to kill the db_conn which will trigger the killing of the process but it feels that this should be the way right? 

But at the same time feels like another race condition 😅

josevalim: Sorry, my comment was not the clearest. You need to implement both solutions to solve both problems. The first solution (for Problem 1) is described in 2 steps.

josevalim: Altogether, you need to implement 3 steps:

1. Instead of immediately disconnecting, ping every 30 seconds. In the first 30 seconds, if there is nobody connected, you will set a "will_disconnect" to true for the next ping. Otherwise the "will_disconnect" is false. If in the second ping "will_disconnect" is true, then you stop. This can be trivially done by changing the message itself to {:check_connected_users, will_disconnect}.

2. Then, instead of immediately returning {:stop, :normal}, instead, remove yourself from the registry

3. Then, in 5 seconds, if someone is connected directly to you, you add yourself back. This may be tricky because you use syn to measure how many connections you have, and you may need other mechanisms (like monitoring clients)

filipecabaco: Oh got it!

I wonder if I will be able to do it with https://hexdocs.pm/syn/syn.html#unregister/2 🤔 

As for tracking connected clients, do you think the state of this process could be a good spot? It's done during the join process and potentially a good spot

chasers: An idea...

Check the count in :syn every 10 seconds, keep a list of those counts with the latest 6 (one minute), if all of those 6 are zero then kill the db connection. 

Note:

Clients are not connected directly to anything really. They get their messages to deliver via PubSub.

Just want to make sure we're not duplicating (or hand rolling) registries where we don't need to.

A "tenant" is registered via :syn and we use that as the authority for who is connected to which tenant. We use this for for our custom PubSub dispatcher for the database records. I guess we're already kind of duplicating registries in this case. Even more reason not to create another probably.

chasers: Also note:

There is the `Extensions.PostgresCdcRls.SubscriptionManager` which replicates this data to a subscription table in the tenant database.

It also does the "stop this connection after a period of no connected users" thing:

https://github.com/supabase/realtime/blob/7bda695d927546762b14bf15185030da9117683a/lib/extensions/postgres_cdc_rls/subscription_manager.ex#L190

Which we can pull out if we get the `Connect` module managing all of these procs (todo later).

chasers: Need to handle processes in state there because we need to serialize adding and removing subscriptions to the database.  

But in this "health check" process we don't really need that complexity at the moment.

chasers: Sorry (rambling) I guess yeah you could just call the `Connect` process to register a Channel process with in and in that you would monitor the channel process for when it goes down.

Only thing here is that we need to ensure it can handle lots of connects/disconnects (assume one million).

filipecabaco: I think that you are right in thinking that at the moment it might be too much complexity as this process should only care about DB connection and shouldn't manage channels and if we add state that reacts to channels lifecycles we would open that Pandora box

To recap, for this module what we have:
- Check if no one is using channels associated with this tenant (counter already exists)
- If no one is using it, we disconnect the DB Connection
- If no one is using and db connection is dead, removes the Process from the Registry by killing it
- We won't be able to be very graceful when new users join since we're using syn which needs time to replicate across the network

In my eyes we have two options that can avoid full channel management and cover concerns mentioned:

## potential solution 1
after checking X times that no one is connected which trigger the unregister from syn (process still alive). Instead of controlling all connected users we can use the state of this process to track if "someone" is trying to reconnect by setting a single flag. If "is_connecting" is true register again. If flag is false kill the db conn process and Connect process after X seconds

- less sync (aka more eventually consistent)
- more processes being killed / started up
- syn might lead to conflicts but we can use conflict resolution already implemented


## potential solution 2
This last step IMO is the tricky one since it can die at the wrong time so what we can do is take advantage of the fact that we can associate metadata with syn and instead of killing the Connection process we can change its metadata to db_conn nil. With this change, upon new join, if we have db_conn metadata nil we know that we need to reconnect and that flow will become more "sync" since we are using only this process mailbox to receice and handle it.

- more sync
- less process changes in the system
- this process will become more of a bottleneck

lib/realtime/tenants/connect.ex
```diff
@@ -66,9 +84,9 @@ defmodule Realtime.Tenants.Connect do
       case res do
         {:ok, conn} ->
           :syn.update_registry(__MODULE__, tenant_id, fn _pid, meta -> %{meta | conn: conn} end)
-          state = %{state | db_conn_reference: Process.monitor(conn)}
+          state = %{state | db_conn_reference: Process.monitor(conn), db_conn_pid: conn}
 
-          {:ok, state}
+          {:ok, state, {:continue, :setup_connected_users}}
```
abc3, line 89: I would either send a `:check_connected_users` message or wrap `Process.send_after` in a function and call it. Currently, a similar code exists three times.

lib/realtime/tenants/connect.ex
```diff
@@ -88,10 +148,12 @@ defmodule Realtime.Tenants.Connect do
...
-  defp call_external_node(tenant_id) do
+  defp call_external_node(tenant_id, opts) do
     with tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_id),
          {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant) do
-      :erpc.call(node, __MODULE__, :connect, [tenant_id], 5000)
+      :erpc.call(node, __MODULE__, :connect, [tenant_id, opts], @erpc_timeout_default)
     end
   end
+
+  defp disconnect?(tenant_id), do: UsersCounter.tenant_users(tenant_id) <= 0
```
abc3, line 158: the `tenant_users/1` function returns only non-negative values https://github.com/supabase/realtime/blob/4ec674e52a551e0e7a449fef6d6e8dea916e8e38/lib/realtime/user_counter.ex#L20

```suggestion
  defp disconnect?(tenant_id), do: UsersCounter.tenant_users(tenant_id) == 0
```

lib/realtime/tenants/connect.ex
filipecabaco: opted to go with a simpler solution for now. ended up using the idea of buckets so we can better control the shutdown of the process (can also be useful in the future for more complex behaviours) but the process unregister is done on process shutdown so that way we keep the behaviour consistent with other parts of the code.

later we can revisit this code as this module will need more elements as we bring new features

https://github.com/supabase/realtime/pull/701/commits/486a8bc8c53b1de1fb66ef09ce994bb656ba038f

lib/realtime/tenants/connect.ex
```diff
@@ -77,6 +97,41 @@ defmodule Realtime.Tenants.Connect do
...
+      tenant_id
+      |> update_connected_users_bucket(connected_users_bucket)
+      |> send_connected_user_check_message(check_connected_user_interval)
+
+    {:noreply, %{state | connected_users_bucket: connected_users_bucket}}
+  end
+
+  def handle_info(:shutdown, %{db_conn_pid: db_conn_pid} = state) do
+    Logger.info("Tenant has no connected users, database connection will be terminated")
+    :ok = GenServer.stop(db_conn_pid)
```
chasers, line 131: What happens after this? Is the process going to keep checking and trying to kill a db_conn_pid that doesn't exist?

filipecabaco: Since we're monitoring the reference of this connection it will be caught by the handle info for that effect and kill itself

chasers: Roger. Okay LGTM.

chasers (approved):

lib/realtime/tenants/connect.ex
```diff
@@ -2,31 +2,40 @@ defmodule Realtime.Tenants.Connect do
...
-  defstruct tenant_id: nil, db_conn_reference: nil
+  @erpc_timeout_default 5000
+  @check_connected_user_interval_default 1000
+  @connected_users_bucket_shutdown [0, 0, 0, 0, 0, 0]
+
+  defstruct tenant_id: nil,
+            db_conn_reference: nil,
+            db_conn_pid: nil,
+            check_connected_user_interval: nil,
+            connected_users_bucket: [1, 1, 1, 1, 1, 1]
```
chasers, line 21: Technically this should be `[1]` because it's starting up.

lib/realtime/tenants/connect.ex
```diff
@@ -88,10 +143,28 @@ defmodule Realtime.Tenants.Connect do
...
          {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant) do
-      :erpc.call(node, __MODULE__, :connect, [tenant_id], 5000)
+      :erpc.call(node, __MODULE__, :connect, [tenant_id, opts], @erpc_timeout_default)
     end
   end
+
+  defp update_connected_users_bucket(tenant_id, connected_users_bucket) do
+    connected_users_bucket
+    |> Enum.drop(1)
+    |> then(&(&1 ++ [UsersCounter.tenant_users(tenant_id)]))
```
chasers, line 156:
```suggestion
    connected_users_bucket
    |> then(&(&1 ++ [UsersCounter.tenant_users(tenant_id)]))
    |> Enum.take(-6)
```

Double check this :D

lib/realtime/tenants/connect.ex
filipecabaco: Went with this approach to make it simpler and to provide all "the slots" needed. This way we also know that we have 6 buckets instead of checking the code

chasers: How about: `[1, 0, 0, 0, 0, 0]`

lib/realtime/tenants/connect.ex
chasers: And then here:

```
[UsersCounter.tenant_users(tenant_id) | connected_users_bucket]
|> Enum.take(6)
```

chasers:
@filipecabaco take the nits if you want and feel free to merge whenever.

PR #707, opened 2023-10-20
fix: only log :syn conflict resolutions

chasers:
@filipecabaco wdyt?

### DIFF
--- a/lib/realtime/syn_handler.ex
+++ b/lib/realtime/syn_handler.ex
@@ -5,12 +5,27 @@ defmodule Realtime.SynHandler do
   require Logger
   alias RealtimeWeb.Endpoint
 
-  def on_process_unregistered(mod, name, _pid, _meta, reason) do
-    Logger.warn("#{mod} terminated: #{inspect(name)} #{node()}")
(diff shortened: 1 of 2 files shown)


lib/realtime/syn_handler.ex
```diff
@@ -5,12 +5,27 @@ defmodule Realtime.SynHandler do
...
+  @spec on_process_unregistered(module(), binary(), pid(), map(), atom()) :: :ok
   def on_process_unregistered(mod, name, _pid, _meta, reason) do
-    Logger.warn("#{mod} terminated: #{inspect(name)} #{node()}")
-    topic = topic(mod)
+    case reason do
+      :syn_conflict_resolution ->
+        Logger.warn("#{mod} terminated: #{inspect(name)} #{node()}")
+
+        topic = topic(mod)
+        Endpoint.local_broadcast(topic <> ":" <> name, topic <> "_down", nil)
```
filipecabaco, line 25: we were doing this if it wasn't `:syn_conflict_resolution` so this changes the behaviour

PR #709, opened 2023-10-23
fix - Set slot_name_suffix to avoid conflicts on deploys

filipecabaco:

Sets a default value for slot_name_suffix so we avoid issues during deployments

Plus sets the SynHandler as conflict resolution for :syn on the Application.ex level

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -59,8 +59,7 @@ if config_env() != :test do
     region: System.get_env("FLY_REGION") || System.get_env("REGION"),
     fly_alloc_id: System.get_env("FLY_ALLOC_ID", ""),
     prom_poll_rate: System.get_env("PROM_POLL_RATE", "5000") |> String.to_integer(),
-    platform: platform,
-    slot_name_suffix: slot_name_suffix
+    platform: platform
 
   queue_target = System.get_env("DB_QUEUE_TARGET", "5000") |> String.to_integer()
   queue_interval = System.get_env("DB_QUEUE_INTERVAL", "5000") |> String.to_integer()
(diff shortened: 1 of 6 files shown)


config/runtime.exs
```diff
@@ -54,7 +59,8 @@ if config_env() != :test do
     region: System.get_env("FLY_REGION") || System.get_env("REGION"),
     fly_alloc_id: System.get_env("FLY_ALLOC_ID", ""),
     prom_poll_rate: System.get_env("PROM_POLL_RATE", "5000") |> String.to_integer(),
-    platform: platform
+    platform: platform,
+    slot_name_suffix: slot_name_suffix
```
abc3, line 63: second time the same?

lib/realtime/application.ex
```diff
@@ -42,6 +42,7 @@ defmodule Realtime.Application do
       name: Realtime.Registry.Unique
     )
 
+    :syn.set_event_handler(Realtime.SynHandler)
```
abc3, line 45: did you remove the setting for the syn handler from the old place?

test/realtime/extensions/cdc_rls/replication_poller_test.exs
```diff
@@ -287,14 +287,21 @@ defmodule ReplicationPollerTest do
   end
 
   describe "slot_name_suffix" do
-    test "when no SLOT_NAME_SUFFIX" do
-      System.delete_env("SLOT_NAME_SUFFIX")
-      assert Poller.slot_name_suffix() == ""
+    setup do
+      slot_name_suffix = Application.get_env(:realtime, :slot_name_suffix)
+
+      on_exit(fn -> Application.put_env(:realtime, :slot_name_suffix, slot_name_suffix) end)
```
abc3, line 293: why does it need to put this key on the exit?

filipecabaco: resetting the env to be equal to what it was in the start since we're changing application setup

abc3 (approved):
LGTM!

PR #714, opened 2023-10-24
fix: Add suspend flag to block new connections to tenant db

filipecabaco:

Add suspend flag to block new connections to tenant db

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -6,6 +6,7 @@ defmodule Realtime.Tenants.Connect do
 
   require Logger
 
+  alias Realtime.Api.Tenant
   alias Realtime.Helpers
   alias Realtime.Tenants
   alias Realtime.UsersCounter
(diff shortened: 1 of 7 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -147,6 +148,7 @@ defmodule Realtime.Tenants.Connect do
 
   defp call_external_node(tenant_id, opts) do
     with tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_id),
+         :ok <- tenant_suspended?(tenant),
          {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant) do
       :erpc.call(node, __MODULE__, :connect, [tenant_id, opts], @erpc_timeout_default)
     end
```
chasers, line 154: This should fail and buble up `:tenant_suspended` in the `RealtimeChannel` so we can log a nice error for people (and respond with it on the join) but with it like this it will explode on a no match error.

chasers:
We need to stop all the procs associated with a tenant when suspending. 

Do you want to do this here or in another PR?

filipecabaco:
yes, for now let's just control the entry point and then handle propagation (aka on suspend, pubsub a kill command to be handled by someone TBD as we're getting closer to the supervision tree we talked about 😅)

chasers (approved):

chasers:
Sounds GTM!!

PR #722, opened 2023-10-26
fix: Add Channels to Tenants

filipecabaco:

Adds the concept of Channels into the Tenant database

Also adds some helper methods to handle SQL queries and results in a more "Elixir" way

### DIFF
--- a/lib/realtime/api/channel.ex
+++ b/lib/realtime/api/channel.ex
@@ -1,0 +1,34 @@
+defmodule Realtime.Api.Channel do
+  @moduledoc """
+  Defines the Channel schema
+  """
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @schema_prefix "realtime"
+  schema "channels" do
+    field(:name, :string)
+    timestamps()
+  end
+
+  def changeset(channel, attrs) do
+    channel
+    |> cast(attrs, [:name, :inserted_at, :updated_at])
+    |> put_timestamp(:updated_at)
+    |> maybe_put_timestamp(:inserted_at)
+    |> validate_required([:name])
+  end
+
+  defp put_timestamp(changeset, field) do
+    put_change(changeset, field, DateTime.utc_now() |> DateTime.to_naive())
+  end
+
+  defp maybe_put_timestamp(changeset, field) do
+    case Map.get(changeset.data, field, nil) do
+      nil -> put_timestamp(changeset, field)
+      _ -> changeset
+    end
+  end
+end
--- a/lib/realtime/channels.ex
+++ b/lib/realtime/channels.ex
@@ -1,0 +1,74 @@
+defmodule Realtime.Channels do
+  @moduledoc """
+  Handles Channel related operations
+  """
+
+  alias Realtime.Api.Channel
+  alias Realtime.Repo
+
+  import Ecto.Query
+
+  @doc """
+  Creates a channel in the tenant database using a given DBConnection
+  """
+  @spec create_channel(map(), DBConnection.t()) :: :ok
+  def create_channel(attrs, db_conection) do
+    channel = Channel.changeset(%Channel{}, attrs)
+    {query, args} = insert_query(channel)
+
+    Postgrex.query!(db_conection, query, args)
+    :ok
+  end
+
+  @doc """
+  Fetches a channel by name from the tenant database using a given DBConnection
+  """
+  @spec get_channel_by_name(DBConnection.t(), String.t()) :: Channel.t() | nil
+  def get_channel_by_name(db_conection, name) do
+    query = from c in Channel, where: c.name == ^name
+    {query, args} = Repo.to_sql(:all, query)
+
+    with res <- Postgrex.query!(db_conection, query, args),
+         [channel] <- pg_result_to_struct(res, Channel) do
+      channel
+    else
+      [] -> nil
+      _ -> raise "Multiple channels with the same name"
+    end
+  end
+
+  # TODO Maybe move this to Repo module
+  defp pg_result_to_struct(%Postgrex.Result{rows: rows, columns: columns}, struct) do
+    Enum.map(rows, fn row ->
+      columns
+      |> Enum.zip(row)
+      |> Enum.map(fn {k, v} -> {String.to_existing_atom(k), v} end)
+      |> Map.new()
+      |> then(&struct(struct, &1))
+    end)
+  end
+
+  # TODO Maybe move this to Repo module
+  defp insert_query(changeset) do
+    schema = changeset.data.__struct__
+    source = schema.__schema__(:source)
+    prefix = schema.__schema__(:prefix)
+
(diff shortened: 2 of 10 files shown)


lib/realtime/channels.ex
```diff
@@ -0,0 +1,39 @@
...
+  def get_channel_by_name(db_conection, name) do
+    query = from c in Channel, where: c.name == ^name
+    {query, args} = Repo.to_sql(:all, query)
+
+    with res <- Postgrex.query!(db_conection, query, args),
+         [channel] <- Repo.pg_result_to_struct(res, Channel) do
+      channel
+    else
+      [] -> nil
+      _ -> raise "Multiple channels with the same name"
```
chasers, line 36: I assume we will have a constraint for this too in the db?

lib/realtime/repo.ex
```diff
@@ -15,4 +15,48 @@ defmodule Realtime.Repo do
       Supervisor.stop(repo)
     end
   end
+
+  @doc """
+  Converts a Postgrex.Result into a given struct
+  """
+  @spec pg_result_to_struct(Postgrex.Result.t(), module()) :: [struct()]
+  def pg_result_to_struct(%Postgrex.Result{rows: rows, columns: columns}, struct) do
+    Enum.map(rows, fn row ->
```
chasers, line 24: Is this how Ecto does it?

lib/realtime/tenants/repo/migrations/20231018144023_create_channels.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Realtime.Tenants.Migrations.CreateChannels do
+  @moduledoc false
+
+  use Ecto.Migration
+
+  def change do
+    create table(:channels, prefix: "realtime") do
```
chasers, line 7: Need to check but I think with self-hosted we have the "_realtime" schema too. Might need to make this dynamic.

lib/realtime/tenants/repo/migrations/20231018144023_create_channels.ex
```diff
@@ -0,0 +1,14 @@
...
+      add(:name, :string, null: false)
+      timestamps()
+    end
+
+    create unique_index(:channels, [:name], prefix: "realtime")
```
chasers, line 12: Ah yep.

chasers (approved):
LGTM

lib/realtime/api/channel.ex
```diff
@@ -0,0 +1,34 @@
...
+  schema "channels" do
+    field(:name, :string)
+    timestamps()
+  end
+
+  def changeset(channel, attrs) do
+    channel
+    |> cast(attrs, [:name, :inserted_at, :updated_at])
+    |> put_timestamp(:updated_at)
+    |> maybe_put_timestamp(:inserted_at)
```
abc3, line 20: why is that necessary?

lib/realtime/tenants/repo/migrations/20231018144023_create_channels.ex
chasers: Ah yeah[ our metadata db uses "_realtime"](https://github.com/supabase/realtime/blob/5c7bdf802592c97f3846585673ffcd16c3579618/README.md?plain=1#L67) but we create it with self-hosted anyways.

lib/realtime/channels.ex
```diff
@@ -0,0 +1,39 @@
...
+    :ok
+  end
+
+  @doc """
+  Fetches a channel by name from the tenant database using a given DBConnection
+  """
+  @spec get_channel_by_name(DBConnection.t(), String.t()) :: Channel.t() | nil
+  def get_channel_by_name(db_conection, name) do
+    query = from c in Channel, where: c.name == ^name
+    {query, args} = Repo.to_sql(:all, query)
```
abc3, line 29: Why is it necessary to write a query in a DSL and then convert it into standard SQL?

abc3: That is the same `select * from realtime.channels where name = $1?`

lib/realtime/channels.ex
filipecabaco: yes 👍 that is set with an unique index

lib/realtime/channels.ex
chasers: I think it's cool 😎

lib/realtime/repo.ex
filipecabaco: it's a bit tricky to find but I think so with this spot in the code
https://github.com/elixir-ecto/ecto/blob/8fac7068ed28cbb63c9844bb4cfba09fb48b9a09/lib/ecto/repo/schema.ex#L595

just needs to do this with a ton of code in the middle 😅

lib/realtime/tenants/repo/migrations/20231018144023_create_channels.ex
filipecabaco: will change 👍

lib/realtime/api/channel.ex
filipecabaco: Since we're not using Ecto but Postgrex directly this isn't done automatically as Ecto is the one handling timestamps

lib/realtime/repo.ex
filipecabaco: will check in the elixir slack where does this happen to see if I can reuse it

lib/realtime/channels.ex
filipecabaco: Using the DSL ends up being less error prone and avoids plain sql in the code which improves readability.

abc3: Sure, but we will need to call a bunch of functions and perform several lookups from ETS, which seems too much to me for a simple SQL select

lib/realtime/api/channel.ex
abc3: Can the channel records be edited in the future? Also, the `realtime.subscription` table uses the `created_at` column to indicate the creation time. So it would be nice to keep consistency

abc3: I think the `created_at` in `realtime.subscription` is filled by default

filipecabaco: can change but this does deviate from the ecto standard (timestamps() uses inserted_at and updated_at) 

I'm unsure if we will be able to change names or any new info we add it here, for now I will assume so.

subscriptions does it with a `created_at = now()` in the insert query.

lib/realtime/channels.ex
filipecabaco: this is the base to then have other more complex queries. I'm also not that concerned for performance in this query at the moment since this and other information for the Authz effort will use the Contextual Cache module if we need more reads

lib/realtime/tenants/repo/migrations/20231018144023_create_channels.ex
chasers: No no we use the `realtime` schema on tenants not `_realtime`.

PR #723, opened 2023-10-27
fix: Handle suspend / unsuspend tenants properly in Connect module

filipecabaco:

Handles events of suspending and unsuspending tenants by:
* Broadcasting a operation message to suspend / unsuspend a tenant
* Connect receives said messages and handles them appropriately
* Invalidates Cachex for that tenant

### DIFF
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -203,7 +203,11 @@ defmodule Realtime.Tenants do
     external_id
     |> Cache.get_tenant_by_external_id()
     |> Tenant.changeset(%{suspend: true})
-    |> Repo.update()
+    |> Repo.update!()
+    |> tap(fn _ ->
+      Cache.invalidate_tenant_cache(external_id)
+      Phoenix.PubSub.broadcast(Realtime.PubSub, "tenant:operations:#{external_id}", :suspend)
+    end)
   end
 
   @doc """
@@ -214,6 +218,10 @@ defmodule Realtime.Tenants do
     external_id
     |> Cache.get_tenant_by_external_id()
     |> Tenant.changeset(%{suspend: false})
-    |> Repo.update()
+    |> Repo.update!()
+    |> tap(fn _ ->
+      Cache.invalidate_tenant_cache(external_id)
+      Phoenix.PubSub.broadcast(Realtime.PubSub, "tenant:operations:#{external_id}", :unsuspend)
+    end)
   end
 end
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -204,10 +204,7 @@ defmodule Realtime.Tenants do
     |> Cache.get_tenant_by_external_id()
     |> Tenant.changeset(%{suspend: true})
     |> Repo.update!()
-    |> tap(fn _ ->
-      Cache.invalidate_tenant_cache(external_id)
-      Phoenix.PubSub.broadcast(Realtime.PubSub, "tenant:operations:#{external_id}", :suspend)
-    end)
+    |> tap(fn _ -> broadcast_operation_event(:suspend, external_id) end)
   end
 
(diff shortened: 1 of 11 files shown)


lib/realtime/tenants.ex
```diff
@@ -203,7 +203,11 @@ defmodule Realtime.Tenants do
     external_id
     |> Cache.get_tenant_by_external_id()
     |> Tenant.changeset(%{suspend: true})
-    |> Repo.update()
+    |> Repo.update!()
+    |> tap(fn _ ->
+      Cache.invalidate_tenant_cache(external_id)
```
abc3, line 208: what happens if there is a cache on other nodes?

lib/realtime/tenants/connect.ex
```diff
@@ -131,11 +132,21 @@ defmodule Realtime.Tenants.Connect do
 
   def handle_info(:shutdown, %{db_conn_pid: db_conn_pid} = state) do
     Logger.info("Tenant has no connected users, database connection will be terminated")
-    :ok = GenServer.stop(db_conn_pid)
+    :ok = GenServer.stop(db_conn_pid, :normal)
```
abc3, line 135: let's add some timeout, to exclude zombies cases

abc3, line 135: let's add some timeout, to exclude zombies cases

lib/realtime/tenants/connect.ex
```diff
@@ -131,11 +132,21 @@ defmodule Realtime.Tenants.Connect do
...
     {:stop, :normal, state}
   end
 
-  @impl GenServer
+  def handle_info(:suspend, %{db_conn_pid: db_conn_pid} = state) do
+    Logger.warn("Tenant was suspended, database connection will be terminated")
```
abc3, line 140: nitpick :D 
```suggestion
    Logger.warning("Tenant was suspended, database connection will be terminated")
```

lib/realtime/tenants/connect.ex
```diff
@@ -131,11 +132,21 @@ defmodule Realtime.Tenants.Connect do
...
+    :ok = GenServer.stop(db_conn_pid, :normal)
```
abc3, line 141: it would also be nice to add a timeout here

lib/realtime/tenants.ex
filipecabaco: it would not be invalidated which is an issue, we would need to trigger this action from the right node which would increase time to ticket resolution

will find another approach

filipecabaco: pushed a potential approach

test/realtime/tenants/cache_supervisor_test.exs
```diff
@@ -0,0 +1,27 @@
...
+  setup do
+    %{tenant: tenant_fixture()}
+  end
+
+  test "invalidates cache on PubSub message", %{tenant: tenant} do
+    external_id = tenant.external_id
+    assert %Tenant{suspend: false} = Cache.get_tenant_by_external_id(external_id)
+
+    # Update a tenant
+    tenant |> Realtime.Api.Tenant.changeset(%{suspend: true}) |> Realtime.Repo.update!()
```
abc3, line 16:
```suggestion
    tenant |> Tenant.changeset(%{suspend: true}) |> Realtime.Repo.update!()
```

lib/realtime/tenants/cache_supervisor.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Realtime.Tenants.CacheSupervisor do
```
abc3, line 1: add, please, @moduledoc and remove PubSub alias

abc3 (approved):

filipecabaco:
made some changes to make the events more clear on what they are doing

PR #724, opened 2023-10-30
fix: Add logs to connection killing

filipecabaco:

Add logs to connection killing

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -311,7 +311,7 @@ defmodule Realtime.Helpers do
   @doc """
   Kills all connections to a tenant database in all connected nodes
   """
-  @spec kill_connections_to_tenant_id(String.t(), atom()) :: :ok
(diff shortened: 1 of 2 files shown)


lib/realtime/helpers.ex
```diff
@@ -321,7 +323,10 @@ defmodule Realtime.Helpers do
           case elem(state, 11) do
             {pid, _, Postgrex.Protocol, opts} ->
               if opts[:hostname] == "db.#{tenant_id}.supabase.co" do
-                IO.puts("Killing #{inspect(pid)}")
+                Logger.warn(
+                  "Killing connection " <> inspect(pid) <> "with reason " <> inspect(reason)
+                )
+
```
abc3, line 329: - there will not be any output if you invoke this function from the REPL
- `warm` is deprecated

filipecabaco: I keep forgetting it ... 

and I just want to check it in logflare since we will need to trigger this to all nodes (should actually do that erpc call here 🤔 )

abc3 (approved):

PR #725, opened 2023-10-30
fix: Async connection killing

filipecabaco:

Async connection killing

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -314,7 +314,7 @@ defmodule Realtime.Helpers do
   @spec kill_connections_to_tenant_id_in_all_nodes(String.t(), atom()) :: :ok
   def kill_connections_to_tenant_id_in_all_nodes(tenant_id, reason) do
     Task.async_stream(
-      Node.list(),
+      Node.list() ++ [node()],
       fn node ->
         :erpc.call(node, __MODULE__, :kill_connections_to_tenant_id, [tenant_id, reason], 5000)
       end,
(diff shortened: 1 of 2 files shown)


lib/realtime/helpers.ex
```diff
@@ -313,9 +313,13 @@ defmodule Realtime.Helpers do
   """
   @spec kill_connections_to_tenant_id_in_all_nodes(String.t(), atom()) :: :ok
   def kill_connections_to_tenant_id_in_all_nodes(tenant_id, reason) do
-    Enum.each(Node.list(), fn node ->
-      :erpc.call(node, __MODULE__, :kill_connections_to_tenant_id, [tenant_id, reason], 5000)
-    end)
+    Task.async_stream(
+      Node.list(),
```
abc3, line 317: I guess it will be invoked on all other nodes, but not on the current one, correct?

filipecabaco: you are right, I need to add node() to this list, great catch thank you

lib/realtime/helpers.ex
```diff
@@ -314,7 +314,7 @@ defmodule Realtime.Helpers do
   @spec kill_connections_to_tenant_id_in_all_nodes(String.t(), atom()) :: :ok
   def kill_connections_to_tenant_id_in_all_nodes(tenant_id, reason) do
     Task.async_stream(
-      Node.list(),
+      Node.list() ++ [node()],
```
abc3, line 317: nitpick:
```suggestion
      [node() | Node.list()],
```
but list concatenation would also be fine for that case

abc3 (approved):

PR #726, opened 2023-10-31
fix: Broaden the errors that stop the Connect module

filipecabaco:

Broaden the errors that stop the Connect module


chasers (approved):

chasers (changes requested):
Actually I don't think this will do anything Postgrex.query should return :ok or :error tuple.

Maybe the errors you search for are coming from the connection pool processes themselves.

filipecabaco:
Yeah... Not worth the effort will close and will try to change the approach

PR #727, opened 2023-10-31
fix: Avoid exponential connection creation

filipecabaco:

Due to the way we were creating connections we kept a lot of Db Connections working that would keep retrying increasing the number of connections per tenant whenever one call would fail.

We now kill the Postgrex pid if we fail to query the database.

### DIFF
--- a/test/realtime/tenants/connect_test.exs
+++ b/test/realtime/tenants/connect_test.exs
@@ -146,7 +146,7 @@ defmodule Realtime.Tenants.ConnectTest do
           ]
         })
 
-      Enum.each(1..1000, fn _ ->
+      Enum.each(1..10, fn _ ->
         Task.start(fn -> Connect.lookup_or_start_connection(tenant.external_id) end)
       end)
 
--- a/test/realtime/tenants/connect_test.exs
+++ b/test/realtime/tenants/connect_test.exs
@@ -123,5 +123,65 @@ defmodule Realtime.Tenants.ConnectTest do
 
       assert {:ok, _} = Connect.lookup_or_start_connection(tenant.external_id)
     end
+
+    test "properly handles of failing calls by avoid creating too many connections" do
+      tenant =
+        tenant_fixture(%{
+          extensions: [
+            %{
+              "type" => "postgres_cdc_rls",
+              "settings" => %{
+                "db_host" => "localhost",
+                "db_name" => "postgres",
+                "db_user" => "postgres",
+                "db_password" => "postgres",
+                "db_port" => "5432",
+                "poll_interval" => 100,
+                "poll_max_changes" => 100,
+                "poll_max_record_bytes" => 1_048_576,
+                "region" => "us-east-1",
+                "ssl_enforced" => true
+              }
+            }
+          ]
+        })
+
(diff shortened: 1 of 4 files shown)


chasers (approved):

test/realtime/tenants/connect_test.exs
```diff
@@ -123,5 +123,67 @@ defmodule Realtime.Tenants.ConnectTest do
...
+          Process.list()
+          |> Enum.map(&Process.info/1)
+          |> Enum.reject(fn info -> is_nil(info) end)
+          |> Enum.map(fn info -> Keyword.get(info, :dictionary, []) end)
+          |> Enum.map(fn dict ->
+            {Keyword.get(dict, :"$logger_metadata$", %{}),
+             Keyword.get(dict, :"$initial_call", {})}
+          end)
+          |> Enum.reject(fn {metadata, _} -> Map.get(metadata, :external_id) != tenant_id end)
+          |> Enum.filter(fn {_, init} -> init == {DBConnection.Connection, :init, 1} end)
```
philss, line 174: It's fine the way you did, and it's totally not a problem since this is a test helper. But I would suggest to use the [`Stream`](https://hexdocs.pm/elixir/Stream.html) module instead of the `Enum` module if multiple operations like these are used together. This can lower the times you traverse the list, while keep the same idea - the main difference is that you need to call `Enum.to_list/1` before the `Enum.count/1`. An alternative is to use a [`for`](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#for/1) comprehension with filters and assignments, that can be shorter, but not so easy to read. WDYT?

philss (commented):
Hey @filipecabaco :wave: 
I'm late for this one, but it looks good to me!

test/realtime/tenants/connect_test.exs
filipecabaco: makes a lot of sense! could even be an interesting Credo rule 🤔 

I will probably focus on readability in this scenario and opt to use Streams

thank you for the suggestion 🙏 will refactor in another PR

wojtekmach: whenever you see an Enum.map,filter,reject consider using a for. It is often clearer than a composition of Stream functions. Here's an (untested) stab at refactoring this piece:

```elixir
for pid <- Process.list(),
    info = Process.info(pid),
    dict = Keyword.get(info, :dictionary, []),
    dict[:"$logger_metadata$"][:external_id] == tenant_id,
    match?({DBConnection.Connection, :init, 1}, dict[:"$initial_call"]) do
  dbg(pid)
  dbg
end
```

Notice each clause is an automatic _filter_, if it is truthy it passes otherwise it's skipped. Each clause is a _map_ too, we can massage the data however we want. I also did `dict[...][...]` given this notation is "nil safe", i.e. `nil[:whatever] #=> nil`.

As @philss said this is of course totally fine as is, especially for a test utility!

filipecabaco: That's way clearer! Will follow this example thank you 🙏


By the way, what could be triggering this issue? It seems that Postgrex pid won't link with the Connect pid. Is this due to the fact that we're trying to do it on the init/1?

PR #731, opened 2023-11-07
fix: Reduce code for ecto schema loading

filipecabaco:

Using the Repo loader to reduce the code loaded and to properly set the metadata of the created struct


w3b6x9 (approved):
lgtm!

PR #732, opened 2023-11-09
fix: Add Channels listing endpoint

filipecabaco:

Adds the listing endpoints and implements more changes to the way we're loading data

### DIFF
--- a/lib/realtime/channels.ex
+++ b/lib/realtime/channels.ex
@@ -9,31 +9,33 @@ defmodule Realtime.Channels do
   import Ecto.Query
 
   @doc """
+  Lists all channels in the tenant database using a given DBConnection
+  """
+  @spec list_channels(DBConnection.conn()) :: {:error, any()} | {:ok, [struct()]}
+  def list_channels(conn) do
+    Repo.all(conn, Channel, Channel)
+  end
+
+  @doc """
   Creates a channel in the tenant database using a given DBConnection
   """
-  @spec create_channel(map(), DBConnection.t()) :: :ok
-  def create_channel(attrs, db_conection) do
+  @spec create_channel(map(), DBConnection.conn()) :: {:ok, Channel.t()} | {:error, any()}
+  def create_channel(attrs, conn) do
     channel = Channel.changeset(%Channel{}, attrs)
     {query, args} = Repo.insert_query_from_changeset(channel)
 
-    Postgrex.query!(db_conection, query, args)
-    :ok
+    conn
+    |> Postgrex.query(query, args)
+    |> Repo.result_to_single_struct(Channel)
   end
 
   @doc """
   Fetches a channel by name from the tenant database using a given DBConnection
   """
-  @spec get_channel_by_name(DBConnection.t(), String.t()) :: Channel.t() | nil
-  def get_channel_by_name(db_conection, name) do
+  @spec get_channel_by_name(String.t(), DBConnection.conn()) ::
+          {:ok, Channel.t()} | {:ok, nil} | {:error, any()}
+  def get_channel_by_name(name, conn) do
     query = from c in Channel, where: c.name == ^name
-    {query, args} = Repo.to_sql(:all, query)
-
-    with res <- Postgrex.query!(db_conection, query, args),
-         [channel] <- Repo.pg_result_to_struct(res, Channel) do
-      channel
-    else
-      [] -> nil
-      _ -> raise "Multiple channels with the same name"
-    end
+    Repo.one(conn, query, Channel)
   end
 end
--- a/lib/realtime/repo.ex
+++ b/lib/realtime/repo.ex
@@ -8,21 +8,63 @@ defmodule Realtime.Repo do
     {:ok, repo} = [name: nil, pool_size: 2] |> Keyword.merge(config) |> Realtime.Repo.start_link()
 
     try do
-      Realtime.Repo.put_dynamic_repo(repo)
+      put_dynamic_repo(repo)
       callback.(repo)
     after
-      Realtime.Repo.put_dynamic_repo(default_dynamic_repo)
+      put_dynamic_repo(default_dynamic_repo)
       Supervisor.stop(repo)
     end
   end
 
   @doc """
+  Lists all records for a given query and converts them into a given struct
+  """
+  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module()) ::
+          {:ok, list(struct())} | {:error, any()}
+  def all(conn, query, result_struct) do
+    conn
+    |> run_all_query(query)
+    |> result_to_structs(result_struct)
+  end
+
+  @doc """
+  Fetches one record for a given query and converts it into a given struct
+  """
+  @spec one(DBConnection.conn(), Ecto.Query.t(), module()) ::
+          {:ok, struct()} | {:ok, nil} | {:error, any()}
+  def one(conn, query, result_struct) do
+    conn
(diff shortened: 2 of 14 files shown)


lib/realtime/channels.ex
```diff
@@ -8,32 +8,42 @@ defmodule Realtime.Channels do
 
   import Ecto.Query
 
+  def list_channels(db_conection) do
+    query = from(c in Channel)
+    {query, args} = Repo.to_sql(:all, query)
+
+    db_conection
+    |> Postgrex.query(query, args)
+    |> Repo.result_to_structs(Channel)
```
abc3, line 17: perhaps this "DSL to encoded result" is worth wrapping in a function, as this is the second time similar code has been used

lib/realtime/channels.ex
```diff
@@ -8,32 +8,42 @@ defmodule Realtime.Channels do
...
+  end
+
   @doc """
   Creates a channel in the tenant database using a given DBConnection
   """
-  @spec create_channel(map(), DBConnection.t()) :: :ok
+  @spec create_channel(map(), DBConnection.t()) :: {:ok, Channel.t()} | {:error, any()}
```
abc3, line 23:
```suggestion
  @spec create_channel(map(), DBConnection.conn()) :: {:ok, Channel.t()} | {:error, any()}
```

lib/realtime/channels.ex
```diff
@@ -8,32 +8,42 @@ defmodule Realtime.Channels do
 
   import Ecto.Query
 
+  def list_channels(db_conection) do
```
abc3, line 11: nitpick, inside Postgrex everywhere used the name "conn", not "db_connection"

lib/realtime/repo.ex
```diff
@@ -19,11 +19,31 @@ defmodule Realtime.Repo do
...
+          {:ok, struct()} | {:ok, nil} | {:error, any()}
+  def result_to_single_struct({:ok, %Postgrex.Result{rows: [row], columns: columns}}, struct) do
+    {:ok, Realtime.Repo.load(struct, Enum.zip(columns, row))}
   end
 
+  def result_to_single_struct({:ok, %Postgrex.Result{rows: []}}, _),
+    do: {:ok, nil}
+
+  def result_to_single_struct({:ok, %Postgrex.Result{rows: rows}}, _),
+    do: raise("expected at most one result but got #{length(rows)} in result")
```
abc3, line 32: double checking if it's okay to raise an error here or it would be better to return a response like `{:error, msg}`

lib/realtime_web/controllers/channels_controller.ex
```diff
@@ -0,0 +1,33 @@
...
+      ]
+    ],
+    responses: %{
+      200 => ChannelResponseList.response()
+    }
+  )
+
+  def index(%{assigns: %{tenant: tenant}} = conn, _params) do
+    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id),
+         {:ok, channels} <- Channels.list_channels(db_conn) do
```
abc3, line 29: Is it possible for a tenant to have channels if there are no connected users?

lib/realtime/channels.ex
filipecabaco: good call will do 👍

lib/realtime/channels.ex
filipecabaco: good call, we should keep that pattern to make it easier to think about what a conn is

lib/realtime_web/controllers/channels_controller.ex
filipecabaco: yes, channels will be created before actually having any connections

abc3 (approved):
feel free to merge it when it's ready

lib/realtime/repo.ex
```diff
@@ -8,22 +8,62 @@ defmodule Realtime.Repo do
     {:ok, repo} = [name: nil, pool_size: 2] |> Keyword.merge(config) |> Realtime.Repo.start_link()
 
     try do
-      Realtime.Repo.put_dynamic_repo(repo)
+      __MODULE__.put_dynamic_repo(repo)
       callback.(repo)
     after
-      Realtime.Repo.put_dynamic_repo(default_dynamic_repo)
+      __MODULE__.put_dynamic_repo(default_dynamic_repo)
```
josevalim, line 14: Question, maybe we can skip the namespacing altogether? Same for `start_link` above.

```suggestion
      put_dynamic_repo(default_dynamic_repo)
```

test/realtime/repo_test.exs
```diff
@@ -146,40 +146,82 @@ defmodule Realtime.RepoTest do
...
+              [
+                %Channel{
+                  __meta__: ^metadata,
+                  id: 1,
+                  name: "foo",
+                  updated_at: ^timestamp,
+                  inserted_at: ^timestamp
+                },
+                %Channel{
+                  __meta__: ^metadata,
```
josevalim, line 218: Instead of pattern matching on the metadata internals, consider assigning `%Channel{}` to a variable and then using `Ecto.get_meta` to access a value of a specific metadata field.

test/realtime_web/controllers/channels_controller_test.exs
```diff
@@ -0,0 +1,55 @@
...
+    start_supervised!({Tenants.Migrations, settings})
+    {:ok, db_conn} = Tenants.Connect.lookup_or_start_connection(tenant.external_id)
+
+    conn =
+      conn
+      |> put_req_header("accept", "application/json")
+      |> put_req_header("authorization", "Bearer #{@token}")
+      |> then(&%{&1 | host: "#{tenant.external_id}.supabase.com"})
+
+    on_exit(fn -> truncate_table(db_conn, "realtime.channels") end)
```
josevalim, line 34: Why is the table truncated? The sandbox is not available?

In any case, one tip is to truncate when the test starts, instead of `on_exit`, for two reasons:

1. If something fails spectacularly and you can't truncate, the next time you run the suite things will fail. If you truncate on setup, then it is always good
2. Sometimes you may want to inspect the table for debugging, but if you always truncate it after, you can't  :)

lib/realtime/channels.ex
```diff
@@ -7,33 +7,35 @@ defmodule Realtime.Channels do
   alias Realtime.Repo
 
   import Ecto.Query
+  @doc """
+  Lists all channels in the tenant database using a given DBConnection
+  """
+  @spec list_channels(DBConnection.conn()) :: {:error, any()} | {:ok, [struct()]}
+  def list_channels(conn) do
+    query = from(c in Channel)
+    Repo.all(conn, query, Channel)
```
josevalim, line 16:
```suggestion
    Repo.all(conn, Channel, Channel)
```

josevalim (commented):
LGTM, I have only added some minor nits!

test/realtime_web/controllers/channels_controller_test.exs
filipecabaco: unfortunately no since it's using Postgrex connections. I might be able to checkout a conn from Ecto but I had no clue on how to do it 😓

test/realtime/repo_test.exs
filipecabaco: will do 👍

lib/realtime/repo.ex
filipecabaco: makes sense 👍 will change

PR #734, opened 2023-11-14
fix: Send crash dump to S3 bucket

filipecabaco:

Currently we're not storing the Erlang Crash dumps. This changes allows to trap exist signals, fetch the crash dump and send it to an S3 bucket.

Can be emulated locally by running:
1. `docker compose up`
2. `docker exec -it realtime-server bin/realtime remote`
3. `Process.whereis(:application_controller) |> Process.exit(:kill)`

## Additional context
Use the Minio Console to fetch the crash logs when testing the docker environment. The address is on minio container startup

Screenshot 2023-11-14 at 00 23 23

### DIFF
--- a/tailscale/wrapper.sh
+++ b/tailscale/wrapper.sh
@@ -10,6 +10,52 @@ if [ "${ENABLE_TAILSCALE-}" = true ]; then
     /tailscale/tailscale up --authkey=${TAILSCALE_AUTHKEY} --hostname="${TAILSCALE_APP_NAME}" --accept-routes=true
 fi
 
+export ERL_CRASH_DUMP=/tmp/erl_crash.dump
+
+function upload_crash_dump_to_s3 {
+    EXIT_CODE=${?:-0}
+
+    bucket=$ERL_CRASH_DUMP_S3_BUCKET
+    s3Key=$ERL_CRASH_DUMP_S3_KEY
+    s3Secret=$ERL_CRASH_DUMP_S3_SECRET
+    s3Host=$ERL_CRASH_DUMP_S3_HOST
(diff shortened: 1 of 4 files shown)


abc3 (approved):
lgtm

tailscale/wrapper.sh
```diff
@@ -10,6 +10,51 @@ if [ "${ENABLE_TAILSCALE-}" = true ]; then
...
+
+    bucket=$ERL_CRASH_DUMP_S3_BUCKET
+    s3Key=$ERL_CRASH_DUMP_S3_KEY
+    s3Secret=$ERL_CRASH_DUMP_S3_SECRET
+    s3Host=$ERL_CRASH_DUMP_S3_HOST
+    s3Port=$ERL_CRASH_DUMP_S3_PORT
+
+    filePath=${ERL_CRASH_DUMP_FOLDER:-tmp}/$(date +%s)_${ERL_CRASH_DUMP_FILE_NAME:-erl_crash.dump}
+    if [ -f "${ERL_CRASH_DUMP_FOLDER:-tmp}/${ERL_CRASH_DUMP_FILE_NAME:-erl_crash.dump}" ]; then
+        cat ${ERL_CRASH_DUMP_FOLDER:-tmp}/${ERL_CRASH_DUMP_FILE_NAME:-erl_crash.dump} >${filePath}
```
tomashley, line 26: Is there a reason to cat the dump file rather than just copy it?
Do you even need to copy it before the upload?

filipecabaco: it's just to add the timestamp to the filename 😅

PR #735, opened 2023-11-14
fix: Add observer_cli to better debug vm status

filipecabaco:

Add observer_cli to better debug vm status


abc3 (approved):
🔥

PR #741, opened 2023-11-17
fix: terminate zombie migration runners

abc3:
While applying migrations to a tenant's database, I observed that `Ecto.Migration.Runner` does not stop when encountering errors. Upon brief inspection, I noticed that the `Ecto.Migrator` only [stops](https://github.com/elixir-ecto/ecto_sql/blob/30c2cbabef78d3fd5647475b6e6c840252571495/lib/ecto/migration/runner.ex#L28) the Agent if the migration succeeds. In the case of an error, the Migrator [catches](https://github.com/elixir-ecto/ecto_sql/blob/30c2cbabef78d3fd5647475b6e6c840252571495/lib/ecto/migrator.ex#L358-L361) it and exits with a `normal` reason, which does not lead to the termination of the [linked](https://github.com/elixir-ecto/ecto_sql/blob/30c2cbabef78d3fd5647475b6e6c840252571495/lib/ecto/migration/runner.ex#L44) Runner process. As a result, this leads to memory leaks when Realtime continuously attempts to run migrations that encounter errors

As a quick solution from our side, we could stop all runners that are linked only to `MigratorSupervisor` after a failure. Normally, they also have a link to the Migrator that initiated them

https://github.com/supabase/realtime/assets/1172600/c6e3c202-a9f7-42a2-b2e7-05e9cc4c3bf6

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -350,6 +350,26 @@ defmodule Realtime.Helpers do
     Enum.each(pids_to_kill, &Process.exit(&1, reason))
   end
 
+  @doc """
+  Kills all Ecto.Migration.Runner processes that are linked only to Ecto.MigratorSupervisor
+  """
+  @spec dirty_terminate_runners :: list()
+  def dirty_terminate_runners() do
+    Ecto.MigratorSupervisor
+    |> DynamicSupervisor.which_children()
+    |> Enum.reduce([], fn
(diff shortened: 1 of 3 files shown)


abc3:
@supabase/dashbit Could you take a look to see if this makes sense?

filipecabaco (approved):
🤩

abc3:
@philss thanks for the feedback! Please let me know if you need any additional info to reproduce such behavior

PR #743, opened 2023-11-17
fix: Add create Channel endpoint

filipecabaco:

Add create Channel endpoint

### DIFF
--- a/lib/realtime/repo.ex
+++ b/lib/realtime/repo.ex
@@ -39,12 +39,25 @@ defmodule Realtime.Repo do
   end
 
   @doc """
+  Inserts a given changeset into the database and converts the result into a given struct
+  """
+  @spec insert(DBConnection.conn(), Ecto.Changeset.t(), module()) ::
+          {:ok, struct()} | {:error, any()} | Ecto.Changeset.t()
+  def insert(conn, changeset, result_struct) do
+    with {query, args} <- insert_query_from_changeset(changeset) do
+      conn
+      |> Postgrex.query(query, args)
+      |> result_to_single_struct(result_struct)
+    end
+  end
+
+  @doc """
   Converts a Postgrex.Result into a given struct
   """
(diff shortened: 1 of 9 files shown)


lib/realtime/repo.ex
```diff
@@ -38,13 +38,26 @@ defmodule Realtime.Repo do
     |> result_to_single_struct(result_struct)
   end
 
+  @doc """
+  Inserts a given changeset into the database and converts the result into a given struct
+  """
+  @spec insert(DBConnection.conn(), Ecto.Changeset.t(), module()) ::
+          {:ok, struct()} | {:error, any()} | Ecto.Changeset.t()
+  def insert(conn, changeset, result_struct) do
+    with {query, args} <- insert_query_from_changeset(changeset) do
```
wojtekmach, line 47: w`insert_query_from_changeset` at the moment when invalid returns `changeset`, it would be better (and matching this type spec) that it returns `{:error, changeset}`!

test/realtime/repo_test.exs
```diff
@@ -114,36 +125,57 @@ defmodule Realtime.RepoTest do
...
+      query = from c in Channel, where: c.name == "potato"
+      assert {:ok, nil} = Repo.one(conn, query, Channel)
+    end
+  end
+
+  describe "insert/3" do
+    test "inserts a new entry with a given changeset and returns struct", %{conn: conn} do
+      changeset = Channel.changeset(%Channel{}, %{name: "foo"})
+      assert {:ok, channel} = Repo.insert(conn, changeset, Channel)
+      assert is_struct(channel, Channel)
```
wojtekmach, line 164: nitpick: you can skip one assertion with this pattern match:

```suggestion
      assert {:ok, %Channel{}} = Repo.insert(conn, changeset, Channel)
```

wojtekmach (commented):
Looks good to me, just one comment about returning `{:error, changeset}` on errors.

abc3 (approved):

PR #744, opened 2023-11-18
fix: Add delete endpoint

filipecabaco:

Adds delete endpoint along with other delete code required for the feature

To be merged after https://github.com/supabase/realtime/pull/743

### DIFF
--- a/lib/realtime/repo.ex
+++ b/lib/realtime/repo.ex
@@ -52,32 +52,35 @@ defmodule Realtime.Repo do
   end
 
   @doc """
-  Converts a Postgrex.Result into a given struct
+  Deletes all records for a given query and converts them into a given struct
   """
-  @spec result_to_single_struct({:ok, Postgrex.Result.t()} | {:error, any()}, module()) ::
-          {:ok, struct()} | {:ok, nil} | {:error, any()}
-  def result_to_single_struct({:ok, %Postgrex.Result{rows: [row], columns: columns}}, struct) do
+  @spec del(DBConnection.conn(), Ecto.Queryable.t()) ::
+          {:ok, non_neg_integer()} | {:error, any()}
+  def del(conn, query) do
+    with {:ok, %Postgrex.Result{num_rows: rows}} <- run_delete_query(conn, query) do
+      {:ok, rows}
+    end
+  rescue
+    error -> {:error, error}
+  end
+
+  defp result_to_single_struct({:ok, %Postgrex.Result{rows: [row], columns: columns}}, struct) do
     {:ok, load(struct, Enum.zip(columns, row))}
   end
 
-  def result_to_single_struct({:ok, %Postgrex.Result{rows: []}}, _),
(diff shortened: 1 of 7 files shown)


lib/realtime/repo.ex
```diff
@@ -52,41 +52,39 @@ defmodule Realtime.Repo do
   end
 
   @doc """
-  Converts a Postgrex.Result into a given struct
+  Deletes all records for a given query and converts them into a given struct
```
wojtekmach, line 55: the description doesn't seem to match the implementation, we're returning a number of deleted rows, not a struct.

lib/realtime/repo.ex
```diff
@@ -52,41 +52,39 @@ defmodule Realtime.Repo do
...
   """
-  @spec result_to_single_struct({:ok, Postgrex.Result.t()} | {:error, any()}, module()) ::
-          {:ok, struct()} | {:ok, nil} | {:error, any()}
-  def result_to_single_struct({:ok, %Postgrex.Result{rows: [row], columns: columns}}, struct) do
+  @spec del(DBConnection.conn(), Ecto.Queryable.t()) ::
+          {:ok, non_neg_integer()} | {:error, any()}
+  def del(conn, query) do
+    with {:ok, %Postgrex.Result{num_rows: rows}} <- run_delete_query(conn, query) do
+      {:ok, rows}
```
wojtekmach, line 61: nitpick: consider using the more descriptive variable name.

```suggestion
    with {:ok, %Postgrex.Result{num_rows: num_rows}} <- run_delete_query(conn, query) do
      {:ok, num_rows}
```

lib/realtime/repo.ex
```diff
@@ -52,41 +52,39 @@ defmodule Realtime.Repo do
...
+    end
+  rescue
+    error -> {:error, error}
```
wojtekmach, line 64: Below we have the following test:

```elixir
   test "handles error on wrong delete", %{conn: conn} do
       # wrong id type
       query = from c in Channel, where: c.id == "potato"
       assert {:error, error} = Repo.del(conn, query)
       assert is_struct(error, Ecto.QueryError)
```

I'd carefully consider using `rescue _ ->` here. It may catch errors that arguably shouldn't. In this particular case I'd say the query is _wrong_ and so it's a programmers bug and del/1 should raise. In other words the fact it's invalid data should have been caught sooner by for example an `Integer.parse(id)` check. If this doesn't sound practical, I'd consider a narrower handling e.g. `rescue Ecto.QueryError -> `

What do you think?

wojtekmach (commented):
Just one comment around error handling.

lib/realtime/repo.ex
filipecabaco: I think you are right! will revert to not catch the exception as it could be hiding some really bad interactions with such a destructive operation 👍

abc3 (approved):

PR #755, opened 2023-12-05
fix: version bump postcss to 8.4.32 to fix vulnerability

w3b6x9:

fix demo dep vulnerability

## Additional context

resolves https://github.com/supabase/realtime/security/dependabot/25


w3b6x9:
we don't have to version bump in mix given that this has to do with `demo` dir.

filipecabaco (approved):

PR #757, opened 2023-12-06
fix: Set connect context to check RLS policies

filipecabaco:

Set connect context to check RLS policies


filipecabaco:
https://github.com/supabase/realtime/pull/793

PR #758, opened 2023-12-08
fix: Apply RLS rules to Channel controller

filipecabaco:

Apply RLS rules to Channel controller


lib/realtime/tenants/authorization.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  def get_authorizations(%Conn{} = conn, db_conn, params) do
+    with {:ok, {:ok, permissions}} <- get_permissions_for_connection(db_conn, params) do
+      {:ok, Conn.assign(conn, :permissions, permissions)}
+    else
+      {:error, _} -> {:error, :unauthorized}
+    end
+  end
+
+  defp get_permissions_for_connection(conn, params) do
```
abc3, line 23:
```suggestion
```suggestion
  @spec get_permissions_for_connection(DBConnection.conn(), map()) ::
          {:ok, map()} | {:error, any()}
  defp get_permissions_for_connection(conn, params) do
```
  defp get_permissions_for_connection(conn, params) do
```

File: lib/realtime/tenants/authorization.ex [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+  alias Phoenix.Socket
+
+  require Logger
+
+  def get_authorizations(%Socket{} = socket, db_conn, params) do
+    with {:ok, {:ok, permissions}} <- get_permissions_for_connection(db_conn, params) do
+      {:ok, Socket.assign(socket, :permissions, permissions)}
+    else
+      {:error, _} -> {:error, :unauthorized}
+    end
```
abc3, line 12: nitpick and a small fix of the nested `:ok` tuple
```suggestion
    case get_permissions_for_connection(conn, params) do
      {:ok, permissions} -> {:ok, assign(socket, :permissions, permissions)}
      _ -> {:error, :unauthorized}
    end
```

lib/realtime/tenants/authorization.ex
abc3, line 23:
```suggestion
  @spec get_permissions_for_connection(DBConnection.conn(), map()) ::
          {:ok, map()} | {:error, any()}
  defp get_permissions_for_connection(conn, params) do
```

lib/realtime/tenants/authorization.ex
```diff
@@ -0,0 +1,68 @@
...
+    Postgrex.transaction(conn, fn conn ->
+      Postgrex.query!(
+        conn,
+        """
+        SELECT
+         set_config('role', $1, true),
+         set_config('realtime.channel_name', $2, true),
+         set_config('request.jwt.claim.role', $3, true),
+         set_config('request.jwt', $4, true),
+         set_config('request.jwt.claim.sub', $5, true),
```
abc3, line 45: you should add "sub" for jwt validation
https://github.com/supabase/realtime/blob/d3549c4587e0357ae1f414485d89bc77dd89b82e/lib/realtime_web/channels/auth/channels_authorization.ex#L22-L29

dev/postgres/00-supabase-schema.sql
```diff
@@ -0,0 +1,40 @@
...
+
+-- Supabase super admin
+create user supabase_admin password 'postgres';
+alter user  supabase_admin with superuser createdb createrole replication bypassrls;
+
+
+-- Extension namespacing
+create schema extensions;
+create extension if not exists "uuid-ossp"      with schema extensions;
+create extension if not exists pgcrypto         with schema extensions;
```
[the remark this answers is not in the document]
filipecabaco: removed all of this, reverted to the old setup script

dev/postgres/01-auth-schema.sql
```diff
@@ -0,0 +1,101 @@
+
+CREATE SCHEMA IF NOT EXISTS auth AUTHORIZATION supabase_admin;
+-- auth.users definition
+
+CREATE TABLE auth.users (
```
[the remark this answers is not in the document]
filipecabaco: removed all of this, reverted to the old setup script

lib/realtime/tenants/repo/migrations/20231204144023_create_config.ex
```diff
@@ -0,0 +1,33 @@
...
+    GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA realtime TO postgres, anon, authenticated, service_role
+    """)
+
+    execute("""
+    GRANT USAGE ON ALL SEQUENCES IN SCHEMA realtime TO postgres, anon, authenticated, service_role
+    """)
+
+    execute("""
+    ALTER TABLE realtime.channels ENABLE row level security;
+    """)
```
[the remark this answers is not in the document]
filipecabaco: we do have this as an hard dependency already due to previous migration files:
* `CreateRealtimeCheckFiltersTrigger`
* `CreateRealtimeApplyRlsFunction`
* `GrantRealtimeUsageToAuthenticatedRole`
* `EnableRealtimeApplyRlsFunctionPostgrest9Compatibility`
* `UpdateRealtimeSubscriptionCheckFiltersFunctionSecurity`
* `EnableGenericSubscriptionClaims`

lib/realtime/tenants/authorization.ex
filipecabaco: `sub` can be optional right? for example `anon` might not need a `sub` in their JWT

test/realtime_web/plugs/rls_authorization_test.exs
```diff
@@ -0,0 +1,96 @@
...
+      %{name: random_string()},
+      [{"id", "1"}],
+      %{read: false}
+    }
+  ]
+
+  for {role, rls, rls_setup_params, conn_params, %{read: expected_read}} <-
+        @select_authenticated_role_tests ++ @select_authenticated_role_on_channel_name_tests do
+    @tag role: role, rls: rls, rls_setup_params: rls_setup_params
+    test "assigns the permissions correctly to #{role} and #{rls} rule", %{
```
josevalim, line 58: Consider an alternative where you create a small helper in your test and share it. Something like:

```
test "assigns the permission correctly to anon and authenticated role", context do
  assert_foo_bar_baz(context, %{name: random_string()}, params, ...)
end
```

You should have a bit of repetition but you should be able to get rid of most meta-programming, which may or may not make them more maintainable. I thought I would mention it as an option.

test/realtime/tenants/authorization_test.exs
```diff
@@ -0,0 +1,145 @@
+defmodule Realtime.Tenants.AuthorizationTest do
+  require Phoenix.ChannelTest
+  use RealtimeWeb.ConnCase
```
josevalim, line 3: For what is worth, I like to explicitly add `async: true | false` to all of my tests and explicitly leave a comment when it has to be `async: false` with the reasons why.

lib/realtime_web/plugs/rls_authorization.ex
```diff
@@ -0,0 +1,45 @@
...
+  def call(conn, _opts), do: unauthorized(conn) |> halt()
+
+  defp set_channel_name(%{path_params: %{"id" => id}}, db_conn, params) do
+    case Channels.get_channel_by_id(id, db_conn) do
+      {:ok, channel} -> Map.put(params, :channel_name, channel.name)
+      _ -> Map.put(params, :channel_name, nil)
+    end
+  end
+
+  defp set_channel_name(_, _, params), do: Map.put(params, :channel_name, nil)
```
josevalim, line 43: All branches call `Map.put`, so you can probably refactor this to:

    Map.put(params, :channel_name, get_chanel_name(conn, db_conn))

josevalim (commented):
Hi @filipecabaco, comments added! 🎄

filipecabaco:
Awesome! thank you will tackle them 🙏

lib/realtime/tenants/repo/migrations/20231204144024_create_rls_helper_functions.ex
```diff
@@ -0,0 +1,37 @@
...
+    $$ language sql stable;
+    """)
+
+    execute("create schema if not exists auth")
+
+    execute("""
+    do $func_exists$
+    begin if not exists(select * from pg_proc where proname = 'role')
+    then
+      create or replace function auth.uid() returns uuid as $function_def$
```
chasers, line 19: Why are we managing auth functions exactly again? These are not our fns.

lib/realtime/tenants/authorization.ex
```diff
@@ -0,0 +1,82 @@
...
+         set_config('request.jwt.claims', $6, true),
+         set_config('request.headers', $7, true)
+        """,
+        [role, channel_name, role, jwt, sub, claims, headers]
+      )
+
+      case Postgrex.query(conn, "SELECT name from realtime.channels", [], mode: :savepoint) do
```
chasers, line 66: I think we need to start handling exceptions from Postgrex.query and turn them into an error tuple.

chasers: And what is `mode: :savepoint` doing for us here?

lib/realtime_web/controllers/channels_controller.ex
```diff
@@ -57,11 +61,12 @@ defmodule RealtimeWeb.ChannelsController do
     ],
     responses: %{
       200 => ChannelResponse.response(),
+      401 => UnauthorizedResponse.response(),
       404 => NotFoundResponse.response()
     }
   )
 
-  def show(%{assigns: %{tenant: tenant}} = conn, %{"id" => id}) do
+  def show(%{assigns: %{tenant: tenant, permissions: {:ok, %{read: true}}}} = conn, %{"id" => id}) do
```
chasers, line 69: Should we start a type for this permissions "policy" and expand it as we go?

test/realtime/tenants/authorization_test.exs
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      {:ok, conn} =
+        Authorization.get_authorizations(
+          Phoenix.ChannelTest.socket(RealtimeWeb.UserSocket),
+          db_conn,
+          params
+        )
+
+      assert {:ok, %{read: false}} = conn.assigns.permissions
```
chasers, line 129: I'm interested in a case where user has `read` permissions but no `write` permissions.

lib/realtime/tenants/repo/migrations/20231204144024_create_rls_helper_functions.ex
filipecabaco: self hosting made easier basically. if we want to keep realtime as self deployable as possible we need to create them on our own if they are not present

lib/realtime/tenants/authorization.ex
filipecabaco: `mode: :savepoint` avoids invalidating the transaction fully and reverts to a previous stage of the transaction which in this case allows us to have a `false` and keeping the transaction ok

```
 In case of an exception, the transaction will rollback to the savepoint and prevent the transaction from failing.
```

from https://hexdocs.pm/ecto/Ecto.Repo.html 

so this should handle part of the exceptions but we should still be careful as we are not doing that when settings configs so some exception handling needs to happen

lib/realtime_web/plugs/rls_authorization.ex
```diff
@@ -0,0 +1,46 @@
...
+  import Plug.Conn
+
+  alias Realtime.Channels
+  alias Realtime.Tenants.Authorization
+  alias Realtime.Tenants.Connect
+
+  def init(opts), do: opts
+
+  def call(%Plug.Conn{assigns: %{tenant: tenant}} = conn, _opts) do
+    params = %{
```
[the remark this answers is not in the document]
filipecabaco: yes we should and in fact could be a struct from `Realtime.Tenants.Authorization` 👀

lib/realtime_web/controllers/channels_controller.ex
filipecabaco: good call, will create a struct for this so we can keep track and change them accordingly instead of being a magic flexible map

test/realtime/tenants/authorization_test.exs
filipecabaco: potentially that will happen at broadcast where users can send information but won't read it (e.g. sending telemetry)

we will bring some of those use cases after we have write permissions ready

chasers (approved):

PR #768, opened 2023-12-23
fix: add realtime/v1 socket endpoint

chasers:
Adds `/realtime/v1` endpoint so the Inspector works locally again.


filipecabaco (commented):
should we have this with a Mix.env() check? or even reuse `code_reloading?`

w3b6x9:
@chasers do we still need this? otherwise, let's close.

PR #770, opened 2023-12-27
fix: Add rpc metrics to promex

filipecabaco:

 Add rpc metrics to promex

### DIFF
--- a/lib/realtime/monitoring/prom_ex/plugins/tenants.ex
+++ b/lib/realtime/monitoring/prom_ex/plugins/tenants.ex
@@ -2,17 +2,40 @@ defmodule Realtime.PromEx.Plugins.Tenants do
   @moduledoc false
 
   use PromEx.Plugin
+
+  alias PromEx.MetricTypes.Event
+
   require Logger
 
   @event_connected [:prom_ex, :plugin, :realtime, :tenants, :connected]
 
   @impl true
+  def event_metrics(opts) do
+    [rpc_metrics(opts)]
+  end
+
+  @impl true
   def polling_metrics(opts) do
     poll_rate = Keyword.get(opts, :poll_rate)
 
-    [
-      metrics(poll_rate)
-    ]
+    [metrics(poll_rate)]
+  end
+
+  defp rpc_metrics(_opts) do
+    Event.build(:realtime, [
+      counter(
+        "realtime.tenants.rpc.call",
+        event_name: [:realtime, :tenants, :rpc],
+        description: "The total count of rpc calls triggered by a tenant action",
+        measurement: :calls
+      ),
+      counter(
+        "realtime.tenants.erpc.call",
+        event_name: [:realtime, :tenants, :rpc],
+        description: "The total count of erpc calls triggered by a tenant action",
+        measurement: :calls
+      )
+    ])
   end
 
   defp metrics(poll_rate) do
--- a/lib/realtime/monitoring/prom_ex/plugins/tenants.ex
+++ b/lib/realtime/monitoring/prom_ex/plugins/tenants.ex
@@ -16,17 +16,11 @@ defmodule Realtime.PromEx.Plugins.Tenants do
 
   defp rpc_metrics(_opts) do
     Event.build(:realtime, [
-      counter(
-        [:prom_ex, :plugin, :realtime, :realtime, :tenants, :rpc],
+      sum(
+        [:realtime, :tenants, :rpc],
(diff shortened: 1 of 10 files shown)


lib/realtime/monitoring/prom_ex/plugins/tenants.ex
```diff
@@ -2,17 +2,40 @@ defmodule Realtime.PromEx.Plugins.Tenants do
...
-      metrics(poll_rate)
-    ]
+    [metrics(poll_rate)]
+  end
+
+  defp rpc_metrics(_opts) do
+    Event.build(:realtime, [
+      counter(
+        "realtime.tenants.rpc.call",
+        event_name: [:realtime, :tenants, :rpc],
```
chasers, line 28: Did you check locally that these show up in the `/metrics` endpoint?

I think these need to match what you `Telemetry.execute`. These are the events telemetry metrics "attach" to.

You are emitting events for `[:realtime, :tenants, :rpc, :calls]` but also `[:realtime, :tenants, :erpc, :calls]` those which I don't think will get picked up by PromEx currently.

chasers: I guess we could have tests for these!

chasers: Oh I think you just updated your PR.

lib/realtime/monitoring/prom_ex/plugins/tenants.ex
```diff
@@ -2,17 +2,40 @@ defmodule Realtime.PromEx.Plugins.Tenants do
...
+        description: "The total count of rpc calls triggered by a tenant action",
+        measurement: :calls
+      ),
+      counter(
+        "realtime.tenants.erpc.call",
+        event_name: [:realtime, :tenants, :rpc],
```
chasers, line 34:
```suggestion
        event_name: [:realtime, :tenants, :erpc],
```

??

lib/realtime/monitoring/prom_ex/plugins/tenants.ex
filipecabaco: I fixed some tests since they used rpc calls and timeouts were still a pos_integer and now they are a keyword list option

indeed we could add some integrations tests to check what the metrics endpoint is returning 🤔

lib/realtime/rpc.ex
```diff
@@ -7,48 +7,42 @@ defmodule Realtime.Rpc do
...
-      target_node: node,
-      origin_node: node()
-    })
-
-    response
-  rescue
-    _ ->
-      Telemetry.execute([:erpc, :call], timeout, %{
+    Telemetry.execute(
+      [:realtime, :tenants, :erpc, :calls],
```
chasers, line 36: We actually probably don't care that there is an `erpc` or an `rpc`. 

And if we do eventually tag them using a single metric will be better because it won't explode the metrics endpoint as much.

lib/realtime/monitoring/prom_ex/plugins/tenants.ex
filipecabaco: yep it was wrong, fixed it afterwards 🤦

lib/realtime/rpc.ex
filipecabaco: good point, will tackle that

lib/realtime/monitoring/prom_ex/plugins/tenants.ex
```diff
@@ -2,17 +2,40 @@ defmodule Realtime.PromEx.Plugins.Tenants do
...
+  defp rpc_metrics(_opts) do
+    Event.build(:realtime, [
+      counter(
+        [:prom_ex, :plugin, :realtime, :realtime, :tenants, :rpc],
+        event_name: [:realtime, :tenants, :rpc],
+        description: "The total count of rpc calls triggered by a tenant action",
+        measurement: :calls
+      ),
+      counter(
+        [:prom_ex, :plugin, :realtime, :realtime, :tenants, :erpc],
```
chasers, line 26: `:realtime, :realtime`?

chasers: Should just start with `:realtime, :tenants` like the other metrics.

chasers: Doh I guess some start with prom_ex etc...

chasers (approved):

PR #771, opened 2023-12-27
fix: Configurable backoff strategy for DBConnection

filipecabaco:

Configurable backoff strategy for DBConnection

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -93,7 +93,7 @@ defmodule Realtime.Helpers do
         queue_target \\ 5_000,
         ssl_enforced \\ true,
         application_name \\ "realtime_supabase",
-        backoff_type \\ :exp
+        backoff_type \\ :rand_exp
       ) do
     Logger.metadata(application_name: application_name)
     metadata = Logger.metadata()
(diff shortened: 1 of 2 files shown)


lib/realtime/helpers.ex
```diff
@@ -87,7 +92,8 @@ defmodule Realtime.Helpers do
         pool \\ 5,
         queue_target \\ 5_000,
         ssl_enforced \\ true,
-        application_name \\ "realtime_supabase"
+        application_name \\ "realtime_supabase",
+        backoff_type \\ :exp
```
chasers, line 96: Default for db connection is rand_exp right? Should we stick with that for the default?

chasers:
@filipecabaco don't you want to use `:stop` when calling `connect_db` from the `Connect` module?

filipecabaco:
that is done with the `check_tenant_connection`. at the moment that is the only time I'm connecting

lib/realtime/helpers.ex
filipecabaco: will change

chasers (approved):

chasers:
> that is done with the `check_tenant_connection`. at the moment that is the only time I'm connecting

Ahh missed that. LGTM!

PR #775, opened 2024-01-09
fix: Add write checks to channels

filipecabaco:
Add write permissions checks for the Authorization modules.

It checks if a given user can write by trying to updating a column in the channels table. After done, reverts the change within the same transaction.

Also took the time to refactor a bit the code to use some helper functions previously built in the Repo module to avoid bare sql code and have a more structured approach.

### DIFF
--- a/lib/realtime/repo.ex
+++ b/lib/realtime/repo.ex
@@ -1,4 +1,6 @@
 defmodule Realtime.Repo do
+  require Logger
+
   use Ecto.Repo,
     otp_app: :realtime,
     adapter: Ecto.Adapters.Postgres
@@ -21,11 +23,11 @@ defmodule Realtime.Repo do
   @doc """
   Lists all records for a given query and converts them into a given struct
   """
-  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module()) ::
+  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module(), [Postgrex.execute_option()]) ::
           {:ok, list(struct())} | {:error, any()}
-  def all(conn, query, result_struct) do
+  def all(conn, query, result_struct, opts \\ []) do
     conn
-    |> run_all_query(query)
+    |> run_all_query(query, opts)
     |> result_to_structs(result_struct)
   end
 
@@ -48,7 +50,7 @@ defmodule Realtime.Repo do
   def insert(conn, changeset, result_struct) do
     with {:ok, {query, args}} <- insert_query_from_changeset(changeset) do
       conn
-      |> Postgrex.query(query, args)
+      |> run_query_with_trap(query, args)
       |> result_to_single_struct(result_struct)
     end
   end
@@ -69,10 +71,10 @@ defmodule Realtime.Repo do
   """
   @spec update(DBConnection.conn(), Ecto.Changeset.t(), module()) ::
           {:ok, struct()} | {:error, any()} | Ecto.Changeset.t()
-  def update(conn, changeset, result_struct) do
+  def update(conn, changeset, result_struct, opts \\ []) do
     with {:ok, {query, args}} <- update_query_from_changeset(changeset) do
       conn
-      |> Postgrex.query(query, args)
+      |> run_query_with_trap(query, args, opts)
       |> result_to_single_struct(result_struct)
     end
   end
@@ -132,13 +134,26 @@ defmodule Realtime.Repo do
     {:ok, to_sql(:update_all, query)}
   end
 
-  defp run_all_query(conn, query) do
+  defp run_all_query(conn, query, opts \\ []) do
     {query, args} = to_sql(:all, query)
-    Postgrex.query(conn, query, args)
+    run_query_with_trap(conn, query, args, opts)
   end
 
   defp run_delete_query(conn, query) do
     {query, args} = to_sql(:delete_all, query)
-    Postgrex.query(conn, query, args)
+    run_query_with_trap(conn, query, args)
+  end
+
+  defp run_query_with_trap(conn, query, args, opts \\ []) do
+    Postgrex.query(conn, query, args, opts)
+  rescue
+    exception ->
+      Logger.error("Postgrex exception: #{inspect(exception)}")
+      {:error, :postgrex_exception}
+  catch
+    :exit, reason ->
+      Logger.error("Postgrex exit: #{inspect(reason)}")
+
+      {:error, :postgrex_exception}
   end
 end
--- a/lib/realtime/tenants/authorization.ex
+++ b/lib/realtime/tenants/authorization.ex
@@ -6,18 +6,25 @@ defmodule Realtime.Tenants.Authorization do
   * read - a boolean indicating whether the connection has read permissions
   """
   require Logger
-  defstruct [:channel_name, :headers, :jwt, :claims, :role]
+
+  import Ecto.Query
+
+  alias Realtime.Repo
+  alias Realtime.Api.Channel
+
+  defstruct [:channel, :headers, :jwt, :claims, :role]
 
   defmodule Permissions do
-    defstruct read: false
+    defstruct read: false, write: false
 
     @type t :: %__MODULE__{
-            :read => boolean()
+            :read => boolean(),
+            :write => boolean()
           }
   end
 
   @type t :: %__MODULE__{
-          :channel_name => binary() | nil,
+          :channel => Channel.t() | nil,
           :claims => map(),
           :headers => keyword({binary(), binary()}),
           :jwt => map(),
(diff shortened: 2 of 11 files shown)


lib/realtime/tenants/authorization.ex
```diff
@@ -73,37 +92,69 @@ defmodule Realtime.Tenants.Authorization do
...
+       set_config('request.headers', $7, true)
+      """,
+      [role, channel_name, role, jwt, sub, claims, headers]
+    )
+  end
+
+  defp check_read_permissions(conn, permissions) do
+    query = from(c in Channel, select: c.name)
+
+    case Repo.all(conn, query, Channel, mode: :savepoint) do
```
chasers, line 116: We really should start wrapping all these in `rescue`s and return exceptions as error tuples.

test/realtime/tenants/authorization_test.exs
```diff
@@ -77,6 +78,54 @@ defmodule Realtime.Tenants.AuthorizationTest do
 
       assert {:ok, %Permissions{read: false}} = conn.assigns.permissions
     end
+
+    @tag role: "authenticated", rls: :write_authenticated_role
+    test "authenticated user has write permissions and reverts check", %{
```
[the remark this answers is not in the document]
filipecabaco: yes 👍 that is the idea. if not the transaction will fail and won't write it

lib/realtime/tenants/authorization.ex
filipecabaco: yes, we can do that at the Repo level so it's simpler 👀 one point of recovery basically

chasers: Everything that ultimately calls Postgrex.query...

For context for others... because in a multi-tenant env we really have no control over how healthy the database is so we don't want to accidentally crash procs if Postgrex.query throws an exception vs an error tuple.

filipecabaco: yes yes what I mean is that we can use the extra methods I added to Realtime.Repo to have those rescues wrapped and for most of it the Authorization module will be able to be rescued 👀

lib/realtime/tenants/authorization.ex
```diff
@@ -73,37 +92,69 @@ defmodule Realtime.Tenants.Authorization do
...
+
+  defp check_write_permissions(_, permissions, %__MODULE__{channel: nil}) do
+    {:ok, %Permissions{permissions | write: false}}
+  end
+
+  defp check_write_permissions(conn, permissions, %__MODULE__{channel: channel}) do
+    check = Ecto.UUID.generate()
+    changeset = Channel.check_changeset(channel, %{check: check})
+
+    case Repo.update(conn, changeset, Channel, mode: :savepoint) do
```
josevalim, line 140: Could you instead do a `Repo.update_all from(Channel, where: [id: ...]), set: [check: true]`? Even if the value is always the same, it would check for permissions, no? Or is there a reason to use a random UUID instead?

josevalim (commented):
Hi @filipecabaco! 👋  I have dropped two comments/questions. :)

lib/realtime/tenants/authorization.ex
```diff
@@ -62,8 +69,20 @@ defmodule Realtime.Tenants.Authorization do
   end
 
   defp get_permissions_for_connection(conn, params) do
+    Postgrex.transaction(conn, fn transaction_conn ->
+      set_config(transaction_conn, params)
+      permissions = %Permissions{}
+
+      with {:ok, permissions} <- check_read_permissions(transaction_conn, permissions),
+           {:ok, permissions} <- check_write_permissions(transaction_conn, permissions, params) do
```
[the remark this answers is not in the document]
filipecabaco: Good call! yes I think I can infer read / write from the write permissions and if that fails then go for a read check

lib/realtime/tenants/authorization.ex
filipecabaco: Yes it would and to be fair there's no good reason to use UUID 🤔 

will change to use that approach to simplify it

filipecabaco: can't use Repo.update_all in this scenario to use the right connection which was dynamically being used but still will avoid string and uuids to make it simpler

lib/realtime/tenants/authorization.ex
filipecabaco: did the change to capture exists to handle some of this exceptions

lib/realtime/repo.ex
```diff
@@ -21,11 +23,11 @@ defmodule Realtime.Repo do
   @doc """
   Lists all records for a given query and converts them into a given struct
   """
-  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module()) ::
+  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module(), keyword()) ::
```
abc3, line 26:
```suggestion
  @spec all(DBConnection.conn(), Ecto.Queryable.t(), module(), [Postgrex.execute_option()]) ::
```

lib/realtime/repo.ex
```diff
@@ -132,13 +134,33 @@ defmodule Realtime.Repo do
...
+  end
+
+  defp run_query_with_trap(conn, query, args, opts \\ []) do
+    task =
+      Task.Supervisor.async_nolink(Realtime.TaskSupervisor, fn ->
+        Postgrex.query(conn, query, args, opts)
+      end)
+
+    result =
+      case(Task.yield(task)) do
```
abc3, line 154:
```suggestion
      case Task.yield(task) do
```

lib/realtime/repo.ex
```diff
@@ -132,13 +134,33 @@ defmodule Realtime.Repo do
...
+        {:ok, result} ->
+          result
+
+        {:exit, _} = error ->
+          Logger.error(inspect(error))
+          {:error, :postrgrex_exception}
```
abc3, line 160: typo
```suggestion
          {:error, :postgrex_exception}
```

lib/realtime/tenants/repo/migrations/20240108234812_add_channels_column_for_write_check.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Realtime.Tenants.Migrations.AddChannelsColumnForWriteCheck do
+  @moduledoc false
+
+  use Ecto.Migration
+
+  def change do
+    alter table(:channels, prefix: "realtime") do
+      add :check, :boolean, null: true
```
abc3, line 8: would it be more informative if `check` was set to `false` by default?

filipecabaco: yeah might be better to keep it with a default value instead of messing around with nil's

lib/realtime/repo.ex
```diff
@@ -132,13 +134,33 @@ defmodule Realtime.Repo do
...
+    run_query_with_trap(conn, query, args, opts)
   end
 
   defp run_delete_query(conn, query) do
     {query, args} = to_sql(:delete_all, query)
-    Postgrex.query(conn, query, args)
+    run_query_with_trap(conn, query, args)
+  end
+
+  defp run_query_with_trap(conn, query, args, opts \\ []) do
```
filipecabaco, line 147: Hey @supabase/dashbit , could you check this section?

We're using Postgrex directly which means that in some occasions we have exceptions that are not properly handled as mentioned in the [query/4](https://hexdocs.pm/postgrex/Postgrex.html#query/4) docs:
```
This function may still raise an exception if there is an issue with types (ArgumentError), connection (DBConnection.ConnectionError), ownership (DBConnection.OwnershipError) or other error (RuntimeError).
```

Added to that I've also noticed that some errors will actually trigger an exit of the process. For example when we're unable to checkout connections from the pool and it [exits the process](https://github.com/elixir-ecto/db_connection/blob/8ef1f2ea54922873590b8939f2dad6b031c5b49c/lib/db_connection/holder.ex#L97) . To cope with all of this situations this was the approach I found since I'm trying to avoid linking processes and "just execute it".

Do you think this is a sane approach? This was based on the docs of [Task](https://hexdocs.pm/elixir/1.14/Task.html )

```
Using Task.Supervisor.async_nolink/2 + Task.yield/2 + Task.shutdown/2 allows you to execute tasks concurrently and retrieve their results or the reason they failed within a given time frame. If the task fails, the caller won't fail. You will receive the error reason either on yield or shutdown.
```

josevalim: Hi @filipecabaco! You don't need the task + trap in this case.

The reason why DBConnection calls exit is to mirror GenServer semantics. In a GenServer, if you try to talk to call a process and it times out, it exists the current process. However, you can catch exits in Elixir:

```
try do
  Postgrex.query(conn, query, args, opts)
catch
  :exit, reason -> ...
rescue
  exception -> ...
end
```

You only need to trap when a linked processes terminates, but that's not the case here. So I think a "defensive" try/catch would do the job here. WDYT?

PR #784, opened 2024-02-01
fix: Move socket opts to connect

filipecabaco:

We only use socket opts to set ipv4 or ipv6 as such we move that logic into the connect_db code to prevent old state from impacting connect_db logic.

### DIFF
--- a/lib/realtime/helpers.ex
+++ b/lib/realtime/helpers.ex
@@ -33,7 +33,6 @@ defmodule Realtime.Helpers do
           :pool => non_neg_integer,
           :port => binary,
           :queue_target => non_neg_integer,
-          :socket_opts => list,
           :ssl_enforced => boolean,
           :user => binary,
           :application_name => binary,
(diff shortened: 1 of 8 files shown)


lib/realtime/helpers.ex
```diff
@@ -382,25 +374,24 @@ defmodule Realtime.Helpers do
...
       {:ok, :inet}
 
       # Using ipv6.google.com
-      iex> Realtime.Helpers.detect_ip_version("8vsiF4ELRsLa1yLdhZGBOw==")
+      iex> Realtime.Helpers.detect_ip_version("ipv6.google.com")
       {:ok, :inet6}
 
       # Using invalid domain
-      iex> Realtime.Helpers.detect_ip_version("ZNVOgBtti0+i/o6eZCPAwA==")
+      iex> Realtime.Helpers.detect_ip_version("potato")
```
chasers, line 389: lol'd chase love potato

chasers (approved):

lib/realtime/helpers.ex
filipecabaco: it's the only valid debugging value 🥔

PR #785, opened 2024-02-15
fix: Move from Logs to Telemetry on Subscription checks

filipecabaco:

Currently logs were leading users to be concerned with their Realtime status when that does not reflect anything of concern since it's part of the normal flow of Realtime to remove phantom subscriptions and cleanup inactive pids from ETS.

As such we're moving this information from Logs to Telemetry so we avoid concerning our users but keep track of happens with the subscription table.

### DIFF
--- a/lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
+++ b/lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
@@ -4,10 +4,13 @@ defmodule Extensions.PostgresCdcRls.SubscriptionsChecker do
   require Logger
 
   alias Extensions.PostgresCdcRls, as: Rls
-  alias Rls.Subscriptions
 
   alias Realtime.Helpers, as: H
   alias Realtime.Rpc
+  alias Realtime.Telemetry
+
+  alias Rls.Subscriptions
+
   @timeout 120_000
   @max_delete_records 1000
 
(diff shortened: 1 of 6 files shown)


lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
```diff
@@ -139,14 +142,13 @@ defmodule Extensions.PostgresCdcRls.SubscriptionsChecker do
     Enum.reduce(pids, [], fn pid, acc ->
       case :ets.lookup(tid, pid) do
         [] ->
-          Logger.error("Can't find pid in subscribers table: #{inspect(pid)}")
+          Telemetry.execute([:realtime, :subscriptions_checker, :pid_not_found], 1, %{})
```
chasers, line 145: Do you need to tag by tenant?

chasers (approved):
If not lgtm.

lib/extensions/postgres_cdc_rls/subscriptions_checker.ex
filipecabaco: we don't have access to them here it would require a bigger refactor so I will keep it as a single number that we can measure for now

abc3: If there's an urgent need for the tenant and you prefer not to pass the ref via a function, it can be obtained through `Logger.metadata()[:project]`

filipecabaco: saw that id was actually the tenant id, used that instead and moved things so we have this as a tenant metric

PR #786, opened 2024-02-16
fix: Refactor Realtime Channel

filipecabaco:

In preparation to add Authorization, we're separating concerns out of the module RealtimeChannel which had most of the logic within the same module.

This will make it easier to extend the Broadcast and Presence features.

We're also moving the Channels endpoints to a v3 route

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -439,7 +439,7 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   defp confirm_token(%{assigns: %{jwt_secret: jwt_secret, access_token: access_token} = assigns}) do
-    secure_key = Application.get_env(:realtime, :db_enc_key)
+    secure_key = Application.fetch_env!(:realtime, :db_enc_key)
 
     with jwt_secret_dec <- Helpers.decrypt!(jwt_secret, secure_key),
          {:ok, %{"exp" => exp} = claims} when is_integer(exp) <-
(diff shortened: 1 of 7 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -474,8 +412,7 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   defp presence_key(params) do
-    with key when is_binary(key) <- params["config"]["presence"]["key"],
-         true <- String.length(key) > 0 do
+    with key when is_binary(key) and key != "" <- params["config"]["presence"]["key"] do
```
philss, line 415: Nice :)

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -502,7 +439,9 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   defp confirm_token(%{assigns: %{jwt_secret: jwt_secret, access_token: access_token} = assigns}) do
-    with jwt_secret_dec <- decrypt_jwt_secret(jwt_secret),
+    secure_key = Application.get_env(:realtime, :db_enc_key)
```
philss, line 442: Nitpick: wouldn't be better to raise if the secure key is not configured?

```suggestion
    secure_key = Application.fetch_env!(:realtime, :db_enc_key)
```

philss (commented):
Nice refactor! LGTM :)

abc3 (approved):

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: oh great call! 🙏

PR #788, opened 2024-02-21
fix: Replication Slot deletion on Tenant delete

filipecabaco:

Replication Slot deletion on Tenant delete

### DIFF
--- a/lib/extensions/postgres_cdc_stream/replication.ex
+++ b/lib/extensions/postgres_cdc_stream/replication.ex
@@ -7,8 +7,8 @@ defmodule Extensions.PostgresCdcStream.Replication do
   use Postgrex.ReplicationConnection
   require Logger
 
-  alias Extensions.PostgresCdcStream, as: Stream
-  alias Realtime.Helpers, as: H
+  alias Extensions.PostgresCdcStream
+  alias Realtime.Helpers
   alias Realtime.Adapters.Postgres.Decoder
 
   alias Decoder.Messages.{
@@ -26,11 +26,9 @@ defmodule Extensions.PostgresCdcStream.Replication do
     opts = connection_opts(args)
 
     slot_name =
-      if args["dynamic_slot"] do
-        args["slot_name"] <> "_" <> (System.system_time(:second) |> Integer.to_string())
-      else
-        args["slot_name"]
-      end
+      if args["dynamic_slot"],
+        do: args["slot_name"] <> "_" <> (System.system_time(:second) |> Integer.to_string()),
+        else: args["slot_name"]
 
     init = %{
       tenant: args["id"],
@@ -42,9 +40,7 @@ defmodule Extensions.PostgresCdcStream.Replication do
   end
 
   @spec stop(pid) :: :ok
-  def stop(pid) do
-    GenServer.stop(pid)
-  end
+  def stop(pid), do: GenServer.stop(pid)
 
   @impl true
   def init(args) do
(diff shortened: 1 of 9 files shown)


lib/realtime/postgres_cdc.ex
```diff
@@ -2,6 +2,11 @@ defmodule Realtime.PostgresCdc do
   @moduledoc false
 
   require Logger
+
+  alias Realtime.Rpc
+  alias Realtime.Tenants
```
philss, line 7:
```suggestion
```

lib/realtime/postgres_cdc.ex
```diff
@@ -37,14 +45,7 @@ defmodule Realtime.PostgresCdc do
...
   def filter_settings(key, extensions) do
-    [cdc] =
-      Enum.filter(extensions, fn e ->
-        if e.type == key do
-          true
-        else
-          false
-        end
-      end)
+    [cdc] = Enum.filter(extensions, fn e -> if e.type == key, do: true, else: false end)
```
philss, line 48: I think this can be further simplified with:

```suggestion
    [cdc] = Enum.filter(extensions, fn e -> e.type == key end)
```

test/realtime_web/controllers/tenant_controller_test.exs
```diff
@@ -127,16 +131,36 @@ defmodule RealtimeWeb.TenantControllerTest do
...
+          start_supervised(
+            {Postgrex,
+             [
+               host: "localhost",
+               database: "realtime_test",
+               username: "postgres",
+               password: "postgres",
+               after_connect: &Postgrex.query!(&1, "set search_path=realtime", [])
+             ]}
+          )
```
philss, line 161: Interesting! Why are you starting this new connection here?

philss (commented):
I added some minor points. But LGTM!

abc3:
what is the actual fix? move delete_tenant_by_external_id into Task?

lib/realtime/postgres_cdc.ex
```diff
@@ -18,15 +23,18 @@ defmodule Realtime.PostgresCdc do
     apply(module, :handle_subscribe, [pg_change_params, tenant, metadata])
   end
 
+  @spec stop(module, Tenant.t(), pos_integer) :: :ok
   def stop(module, tenant, timeout \\ @timeout) do
-    apply(module, :handle_stop, [tenant, timeout])
+    with {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant) do
+      Rpc.enhanced_call(node, module, :handle_stop, [tenant.external_id, timeout])
```
abc3, line 29: why do we need this change?

lib/extensions/postgres_cdc_stream/replication.ex
```diff
@@ -124,7 +121,7 @@ defmodule Extensions.PostgresCdcStream.Replication do
     Logger.debug("Got message: #{inspect(msg)}")
     [{_, columns, schema, table}] = :ets.lookup(state.tid, msg.relation_id)
 
-    %NewRecord{
+    record = %NewRecord{
```
abc3, line 124: This variable will not be used anywhere except in the next function. Did you take this from the coding style guidelines?

test/realtime_web/controllers/tenant_controller_test.exs
filipecabaco: The other way to start a connection depends on the tenant which fails after tenant deletion.

lib/realtime/postgres_cdc.ex
filipecabaco: will remove, killing via pid alone should be enough.

my thought was that the owner node should be the one killing the pid but that doesn't bring much benefit

lib/realtime/postgres_cdc.ex
filipecabaco: 🤦 yeah it does 😆

filipecabaco:
@abc3 yes 😅 at least according to the testing that was the changing factor... probably some weird race condition happening plus the tests were not complete enough so that was my bad with the first implementation 😮‍💨

abc3 (approved):

PR #793, opened 2024-02-27
WIP

filipecabaco:

WIP


filipecabaco:
https://github.com/supabase/realtime/pull/796

Feature was getting too big, breaking things apart to tackle permission rules first and then extend current broadcast implementation to use it

PR #794, opened 2024-02-28
fix: Load test Realtime migration on connect

filipecabaco:

Get Load testing pipeline ready to test the impact of migrations on Connect

### DIFF
--- a/.github/workflows/staging_supabench.yml
+++ b/.github/workflows/staging_supabench.yml
@@ -1,0 +1,31 @@
+name: Supabench
+on:
+  push:
+    branches:
+      - main
+env:
+  FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+jobs:
+  deploy:
+    name: Deploy QA application
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: superfly/flyctl-actions@1.4
+        with:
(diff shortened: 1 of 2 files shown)


abc3 (approved):

.github/workflows/staging_supabench.yml
```diff
@@ -0,0 +1,31 @@
...
+      - uses: superfly/flyctl-actions@1.4
+        with:
+          args: "-c deploy/fly/qa.toml scale count 1"
+      - uses: superfly/flyctl-actions@1.4
+        with:
+          args: "-c deploy/fly/qa.toml deploy --build-arg SLOT_NAME_SUFFIX=${GITHUB_SHA::7}"
+      - uses: superfly/flyctl-actions@1.4
+        if: always()
+        with:
+          args: "-c deploy/fly/qa.toml scale count 0"
```
egor-romanov, line 23: we no longer host on fly, so probably this can be removed

or we have to deploy staging cluster here

egor-romanov: well, the point is that we have to trigger load test `- name: Start Supabench` after staging deploy d=somes through

filipecabaco: we can trigger that manually right?

didn't we do aws testing recently? is it the same flow?

PR #797, opened 2024-03-04
fix: Distributed Tenant Cache invalidation

filipecabaco:

Creates a new method to emit a cache invalidation message to all connected nodes. Also using it in the delete action of Tenant Controller since this could create issues with replication_slots

### DIFF
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -222,7 +222,7 @@ defmodule Realtime.Tenants do
   defp broadcast_operation_event(action, external_id) do
     Phoenix.PubSub.broadcast!(
       Realtime.PubSub,
-      "realtime:operations:suspend_tenant",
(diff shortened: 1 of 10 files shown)


abc3 (approved):

chasers (approved):

josevalim (commented):
LGTM!

One of the things I like to do in such cases is to have a module called Realtime.Messages, where are all of the subscribe functions are defined. Such as `subscribe_to_cache` and `broadcast_cache`. This way you guarantee they are all in sync.

Another alternative, with a similar effect, is a `Realtime.Topics` module, where you store the topic names, to also avoid typos and facilitate reuse.

But those are just suggestions, it is up to you. :)

filipecabaco:
Really like that idea as now it feels tricky to actually find the source of requests and what is being sent where.

Will tackle that refactor for sure.

Thank you for the feedback 🙏

PR #798, opened 2024-03-06
feat: Authorization for Broadcast

filipecabaco:

Implement authorization using the Realtime.Tenants.Authorization logic using RLS rules.

The flow will be:
* User connects
* On Join message we check if it's a private channel. Channel privacy is determined by the Configuration received on WS Join with the option in `%{broadcast: %{public: false}}`
* If it is, we fetch their database connection, fetch the channel from cache if available and set database context
* The policies are "calculated" to that given socket and assigned to it
* If on join the user does not have channel or broadcast read permissions, we return {:error, :unauthorized} to disconnect the socket
* If on broadcast the user does not have broadcast write permissions, no message will be sent

To ensure the change is backward compatible, for now, we will have the assumption that if the configuration does not determine if the channel is private or public, we will assume that it's public


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -586,4 +601,33 @@ defmodule RealtimeWeb.RealtimeChannel do
     Logger.error("Start channel error: " <> error_msg)
     {:error, %{reason: error_msg}}
   end
+
+  defp assign_policies(false, db_conn, sub_topic, access_token, claims, socket) do
+    channel =
+      case ChannelsCache.get_channel_by_name(sub_topic, db_conn) do
```
chasers, line 607: Thinking ... this cache needs to be keyed per Channel / client JWT because the response depends on the JWT passed in the query. 

And it needs to get invalidated when the JWT expires. Which actually you can set on individual keys with Cachex.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -586,4 +601,33 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+    authorization_context =
+      Authorization.build_authorization_params(%{
+        channel: channel,
+        headers: socket.assigns.headers,
+        jwt: access_token,
+        claims: claims,
+        role: claims["role"]
+      })
+
+    {:ok, socket} = Authorization.get_authorizations(socket, db_conn, authorization_context)
```
chasers, line 621: Ah ... if you set the authorization context when you query for the Channel, if RLS says you can't read the Channel than it won't get returned in the query. 

So you're hitting the db unnecessariy above really.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -586,4 +601,33 @@ defmodule RealtimeWeb.RealtimeChannel do
     Logger.error("Start channel error: " <> error_msg)
     {:error, %{reason: error_msg}}
   end
+
+  defp assign_policies(false, db_conn, sub_topic, access_token, claims, socket) do
+    channel =
+      case ChannelsCache.get_channel_by_name(sub_topic, db_conn) do
+        {:error, :not_found} -> raise "Channel #{sub_topic} hasn't been initialized"
```
chasers, line 608: Why do you want to raise here and not return with an error tuple? 

We want to make sure to respond to the client with some message if the Channel is not found or if they do not have read permissions.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -586,4 +601,33 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        channel: channel,
+        headers: socket.assigns.headers,
+        jwt: access_token,
+        claims: claims,
+        role: claims["role"]
+      })
+
+    {:ok, socket} = Authorization.get_authorizations(socket, db_conn, authorization_context)
+
+    case socket.assigns.policies do
```
chasers, line 623: In the case of an authorization error I was thinking we could respond here with the full policy generated.

Would make debugging your policies way easier probably. Unsure of security implications, but I can't think of any.

lib/realtime_web/channels/realtime_channel.ex
chasers: We'll need to cache this too. 

IMPORTANT: ttl on the specific key needs to be set to the same time the JWT expires

filipecabaco: tricky thing here is that we need to fetch the channel before the config is applied

we might be able to do it differently:
* Use only channel_name on the config as is
* After channel read permissions are checked, fetch the channel so we have access to the channel.id which is needed to fetch the broadcast information

we could also modify broadcast to also include channel_name and then we would not need the channel

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: can this be tackled at a different PR? maybe even remove cache for now and tackle cache in a different PR

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: my bad, this should not raise and should be an error as we should have proper reasons as you mentioned

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: that's why I was logging them out so the owner can know but the user isn't fully aware but honestly security by obfuscation isn't the best as it will never fully be useful 😅

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: wdyt?

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: now I noticed I do not log out 🤦

filipecabaco: I returned an error with more information. I can send the policies afterwards if we want to

lib/realtime_web/channels/realtime_channel.ex
chasers: Seem okay tho, and maybe a good way to error on "create this channel first" so you know it's created vs created but filtered because of RLS rules.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -586,4 +604,38 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+          })
+
+        {:ok, socket} = Authorization.get_authorizations(socket, db_conn, authorization_context)
+
+        case socket.assigns.policies do
+          %Policies{broadcast: %BroadcastPolicies{read: false}} ->
+            {:error, "You do not have permissions to access this broadcast"}
+
+          %Policies{channel: %ChannelPolicies{read: false}} ->
+            {:error, "You do not have permissions to access this channel"}
```
chasers, line 630: When would you have access to a Channel and not have access to Broadcast? 

Probably never I guess but you -could- write an RLS policy this way.

chasers: Maybe "You do not have permissions to read Broadcast messages on this Channel"

And "You do not have permissions to read this Channel"

Need to be very specific here in our error responses.

Technically you could have permission to write Broadcast messages to a Channel but not read them.

chasers (approved):

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: > When would you have access to a Channel and not have access to Broadcast?
> Probably never I guess but you -could- write an RLS policy this way.
this two things are "decoupled" as we pass the channel in the context and the channel is fetched as a superuser. We could link both 👀 

> Need to be very specific here in our error responses.

Yeah makes sense to go with more specific error messages as this will be confusing

lib/realtime/channels/cache_supervisor.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Realtime.Channels.CacheSupervisor do
+  @moduledoc """
+  Supervisor for Channels Cache and Operational processes
+  """
+  use Supervisor
+
+  alias Realtime.Channels.Cache
+
+  def start_link(init_arg) do
+    Supervisor.start_link(__MODULE__, init_arg, name: __MODULE__)
```
josevalim, line 10: Just curiosity: any reason for having an intermediate supervisor rather than starting Cachex directly in the supervision tree?

lib/realtime/tenants/authorization.ex
```diff
@@ -36,19 +36,13 @@ defmodule Realtime.Tenants.Authorization do
...
-      channel: channel,
-      headers: headers,
-      jwt: jwt,
-      claims: claims,
-      role: role
+      channel: Map.get(map, :channel),
+      headers: Map.get(map, :headers),
+      jwt: Map.get(map, :jwt),
+      claims: Map.get(map, :claims),
+      role: Map.get(map, :role)
```
josevalim, line 45: Theoretically I prefer the previous code, as it is more assertive against typos and missing keys, although maybe these values can be missing now, hence the change? :)

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -48,9 +52,12 @@ defmodule RealtimeWeb.RealtimeChannel do
          :ok <- limit_joins(socket.assigns),
          :ok <- limit_channels(socket),
          :ok <- limit_max_users(socket.assigns),
-         {:ok, claims, confirm_token_ref} <- confirm_token(socket),
-         {:ok, _} <- Connect.lookup_or_start_connection(tenant),
-         is_new_api <- is_new_api(params) do
+         {:ok, claims, confirm_token_ref, access_token} <- confirm_token(socket),
+         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant),
+         is_new_api = is_new_api(params),
```
josevalim, line 57: I would move `is_new_api(params)` inside the `do`-block. It does not seem to be used by `<-` clauses.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -83,7 +90,8 @@ defmodule RealtimeWeb.RealtimeChannel do
         presence_key: presence_key(params),
         self_broadcast: !!params["config"]["broadcast"]["self"],
         tenant_topic: tenant_topic,
-        channel_name: sub_topic
+        channel_name: sub_topic,
+        public: !!public
```
josevalim, line 94: `is_public` already guarantees it is a boolean, so perhaps: `public: public` is enough?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -501,6 +518,13 @@ defmodule RealtimeWeb.RealtimeChannel do
   defp is_new_api(%{"config" => _}), do: true
   defp is_new_api(_), do: false
 
+  defp is_public(params) do
+    case params["config"]["broadcast"]["public"] do
+      nil -> true
+      public -> !!public
+    end
```
josevalim, line 525: I think this would be equivalent to:

    params["config"]["broadcast"]["public"] != false

PR #800, opened 2024-03-07
fix: Reduce number of Credo warnings

filipecabaco:

First pass on reducing Credo warnings
Before
Screenshot 2024-03-07 at 17 29 33
After
Screenshot 2024-03-07 at 17 29 46


abc3 (approved):
🔥

PR #801, opened 2024-03-07
feat: add jwks support

hf:
Adds support for verifying end-user JWTs using JWKS in addition to the JWT secret. This means that third-party authentication systems that use asymmetric JWTs can be used with Realtime.

WIP


hf:
Closing in favor of #807

PR #806, opened 2024-03-11
fix: Add proper policies around Channel Controller

filipecabaco:

To support API use case and SDK support we need to implement the required policies.

This adds a new Policy to ChannelPolicy that defines:
* If a channel name is in the request body of a post request, we add it to the authorization context
* If the user is able to Create a channel fully and Delete it fully we set it to write true

This will be used by the controller for the create and delete flows. Update flow already had a check since we are able to load a channel and the permissions will be more fine grained to that channel

### DIFF
--- a/lib/realtime_web/plugs/rls_authorization.ex
+++ b/lib/realtime_web/plugs/rls_authorization.ex
@@ -27,7 +27,8 @@ defmodule RealtimeWeb.RlsAuthorization do
          {:ok, conn} <- Authorization.get_authorizations(conn, db_conn, params) do
       conn
     else
-      error -> conn |> FallbackController.call(error) |> halt()
+      error ->
+        conn |> FallbackController.call(error) |> halt()
     end
   end
 
@@ -38,7 +39,7 @@ defmodule RealtimeWeb.RlsAuthorization do
 
     params =
       cond do
-        conn.method == "POST" && Map.get(body_params, "name", nil) ->
+        Map.get(body_params, "name", nil) ->
           name = Map.fetch!(body_params, "name")
           Map.put(params, :channel_name, name)
 
--- a/lib/realtime_web/plugs/rls_authorization.ex
+++ b/lib/realtime_web/plugs/rls_authorization.ex
@@ -10,6 +10,7 @@ defmodule RealtimeWeb.RlsAuthorization do
   alias Realtime.Tenants.Authorization
   alias Realtime.Tenants.Connect
 
+  alias RealtimeWeb.FallbackController
   def init(opts), do: opts
 
   def call(%Plug.Conn{assigns: %{tenant: tenant}} = conn, _opts) do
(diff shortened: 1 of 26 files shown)


filipecabaco:
I will stop using Connect.lookup_or_connect and setup one way to use start_supervised! to reduce the issues.

test/realtime_web/controllers/channels_controller_test.exs
```diff
@@ -86,60 +94,81 @@ defmodule RealtimeWeb.ChannelsControllerTest do
...
+      assert conn.status == 202
+    end
+
+    @tag role: "service_role"
+    test "returns unauthorized if id doesn't exist", %{conn: conn} do
+      conn = delete(conn, ~p"/v3/api/channels/0")
+      assert conn.status == 401
+    end
+
+    @tag role: "authenticated"
```
chasers, line 142: Should we have a test where `authenticated` users can create channels if RLS is setup correctly?

filipecabaco: actually I'm thinking of changing everything to authenticated instead of service role 👀

lib/realtime_web/plugs/rls_authorization.ex
```diff
@@ -26,21 +27,32 @@ defmodule RealtimeWeb.RlsAuthorization do
...
   def call(conn, _opts), do: unauthorized(conn) |> halt()
 
-  defp set_channel_params_for_authorization_check(%{path_params: path_params}, db_conn, params) do
+  defp set_channel_params_for_authorization_check(conn, db_conn, params) do
+    %{path_params: path_params, body_params: body_params} = conn
+
+    params =
+      cond do
+        conn.method == "POST" && Map.get(body_params, "name", nil) ->
+          name = Map.get(body_params, "name")
```
philss, line 42: Nitpick: since the default for `Map.get/2` is already `nil`, you can avoid this here. And since you don't need a default, you can use the access protocol. An additional idea is to use `Map.fetch!` in the second call, since you know it is there. WDYT?

```suggestion
        conn.method == "POST" && body_params["name"] ->
          name = Map.fetch!(body_params, "name")
```

philss (commented):
Looks good to me!

> Tests are failing and I'm aware but due to the changes we should start reviewing and I will keep tracking down flakyness issues. Seems that connections are being checked out at the wrong time.

Please let us know if you think we can help with something here. I believe that using `start_supervised!`, as you mentioned, can be effective.

lib/realtime_web/plugs/rls_authorization.ex
filipecabaco: Thank you! I always forget that 🤦 will change 🙏

filipecabaco:
Thank you @philss ! 

Yeah I think it's the fact that we have something depending on the `Repo.Realtime` sandboxed on the `Realtime.Tenants.Connect` module init so letting Realtime.Tenants.Connect be started with `start_supervised!` probably will help

abc3 (approved):

PR #812, opened 2024-03-20
fix: Infer private channel based on Channel entry

filipecabaco:

To facilitate the usage of Authz in Realtime during this first iteration we will have the assumption that if we have a Channel present in the tenant database we will have the assumption that the user intented to have a private channel which will end calculating the policies based on RLS rules.

Also small fix for counter which was counting failed sent messages

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -57,9 +57,10 @@ defmodule RealtimeWeb.RealtimeChannel do
          channel = ChannelsCache.get_channel_by_name(sub_topic, db_conn),
          {:ok, socket} <- assign_policies(channel, db_conn, access_token, claims, socket) do
       is_new_api = is_new_api(params)
-      Realtime.UsersCounter.add(transport_pid, tenant)
-
+      public? = match?({:ok, _}, channel)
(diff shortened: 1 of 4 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -75,6 +74,12 @@ defmodule RealtimeWeb.RealtimeChannel do
         module: module
       }
 
+      public? =
+        case channel do
+          {:ok, _} -> false
+          _ -> true
+        end
```
abc3, line 81: nitpick
```suggestion
      public? = match?({:ok, _}, channel)
```

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -26,18 +26,24 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
+          nil
+      end
 
-      _ ->
-        nil
-    end
+    socket =
+      if sent?,
+        do: count(socket),
+        else: socket
```
abc3, line 46: nitpick, move the rate counter logic into the count/0 function

abc3 (approved):

PR #813, opened 2024-03-21
feat: Add Presence Permission

filipecabaco:

Implementation for Presence Authorization where users will be able to define track / read permissions for presence using RLS policies.


lib/realtime/api/presence.ex
```diff
@@ -0,0 +1,42 @@
...
+  schema "presences" do
+    field(:check, :boolean, default: false)
+    timestamps()
+
+    belongs_to(:channel, Realtime.Api.Channel)
+  end
+
+  def changeset(broadcast, attrs) do
+    broadcast
+    |> cast(attrs, [:check, :inserted_at, :updated_at, :channel_id])
```
wojtekmach, line 21: This is a bit out of the scope of the PR (*) but perhaps worth taking a step back and talking about casting. Ecto has a nice [External vs internal data](https://hexdocs.pm/ecto/Ecto.Changeset.html#module-external-vs-internal-data) section and long story short is: in our experience it's useful to have a clear distinction between internal and external data and use _casting_ exclusively for the latter. When we work with internal data by definition it already has proper types and values within allowed sets, if it ain't, it's a programming bug. In other words, if we somehow got an invalid value it shouldn't be gracefully handled (which casting does), it's a programming bug that we should fix.

Btw, is there a use case for changing inserted_at/updated_at by callers? They tend to be automatically handled by Ecto. And if you really need to do it on a special occasion, you can always do `change(struct, inserted_at: ...)`

Aside from internal vs external data, there's also the matter of casting FKs like channel_id. In this particular PRs it's fine but we believe it's a good habit to build to almost never do that (and when you do, explain in a comment) because it might open a security hole. A malicious user could pass a `channel_id` that belongs to another tenant and gain access. (This might be totally moot given RLS but I believe it might be instructive anyway.) A quick fix is:

```diff
- Broadcast.changeset(%Broadcast{}, %{channel_id: channel.id})
+ Broadcast.changeset(%Broadcast{channel_id: channel.id}, %{})
```

and conceptually it is: `channel.id` is a value from struct (internal data), `%Broadcast{}` is a struct (internal data). In this case we don't have any external data so we might as well do:

```elixir
change(%Broadcast{channel_id: channel.id})
```

(*) We have similar structure in Api.Broadcast and so it's worth it to keep it uniform. Also apologies if we already had similar discussions in the past!

lib/realtime/api/presence.ex
```diff
@@ -0,0 +1,42 @@
...
+    |> put_timestamp(:updated_at)
+    |> maybe_put_timestamp(:inserted_at)
```
wojtekmach, line 23: I'm sure it has came up before but double-checking, do we need to manually handle the timestamps? Ecto should be doing that already.

lib/realtime/api/presence.ex
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def check_changeset(broadcast, attrs) do
+    broadcast
+    |> change()
+    |> put_change(:check, attrs[:check])
```
wojtekmach, line 29:
```suggestion
    change(broadcast, check: attrs[:check])
```

or given the internal vs external discussion above, I'd treat it as internal data:

```elixir
def check_changeset(broadcast, check) when is_boolean(check) do
  change(broadcast, check: check)
end
```

or remove this function and do everything at the call site.

lib/realtime/tenants/authorization/policies/presence_policies.ex
```diff
@@ -0,0 +1,95 @@
...
+  alias Realtime.Tenants.Authorization
+  alias Realtime.Tenants.Authorization.Policies
+
+  defstruct read: false, write: false
+
+  @behaviour Realtime.Tenants.Authorization.Policies
+
+  @type t :: %__MODULE__{
+          :read => boolean(),
+          :write => boolean()
```
wojtekmach, line 24:
```suggestion
          read: boolean(),
          write: boolean()
```

test/integration/rt_channel_test.exs
```diff
@@ -470,4 +610,23 @@ defmodule Realtime.Integration.RtChannelTest do
...
+    {:ok, db_conn} = Connect.lookup_or_start_connection(tenant.external_id)
+
+    clean_table(db_conn, "realtime", "presences")
+    clean_table(db_conn, "realtime", "broadcasts")
+    clean_table(db_conn, "realtime", "channels")
+    channel = channel_fixture(tenant)
+
+    if Map.has_key?(context, :policies) do
+      create_rls_policies(db_conn, context.policies, channel)
+    end
```
wojtekmach, line 626: this is of course totally fine as is, for completeness a common idiom is:

```suggestion
    if policies = context[:policies] do
      create_rls_policies(db_conn, policies, channel)
    end
```

wojtekmach (commented):
Looks good to me!

lib/realtime/api/presence.ex
filipecabaco: This was really informative! Thank you!

Probably will do it in a different PR but indeed this changes sound important so will apply them to the schemas that need them

lib/realtime/api/presence.ex
filipecabaco: Since we're not using Ecto to directly connect to the database and only Postgrex we end up needed to handle this unfortunately

lib/realtime/api/presence.ex
filipecabaco: will tackle on a different PR so all changes are captured there

test/integration/rt_channel_test.exs
filipecabaco: looks way better will apply

abc3 (approved):

PR #817, opened 2024-03-29
fix: Run Query within transaction of DB Conn owner

filipecabaco:

Postgrex is unable to checkout connections from external nodes (as expected...) so we need to run an rpc call against the db connection owner instead.

This abstracts that work by giving that work to the Connect module which already is the one attributing said connections.


chasers (approved):
Feels like the Connect module is getting a bit large for what it was originally intended. 

Like we're not actually running this through the Connect GenServer...

For now I'd call that a nit, but looks fine otherwise.

chasers:
Should this maybe be a helper somewhere else?

filipecabaco:
potentially yes, can move it to the helpers module since we already need to clean up that module

PR #820, opened 2024-04-01
fix: Posgtres role with Realtime Admin grants

filipecabaco:

Creates a new role called `supabase_realtime_admin`, moves ownership to this role and then we grant all the attributes roles to postgres user so users can add RLS rules to realtime tables

### DIFF
--- a/lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
+++ b/lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
@@ -9,6 +9,7 @@ defmodule Realtime.Tenants.Migrations.CreateRealtimeAdminAndMoveOwnership do
     execute("GRANT ALL PRIVILEGES ON SCHEMA realtime TO supabase_realtime_admin")
     execute("GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA realtime TO supabase_realtime_admin")
     execute("GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA realtime TO supabase_realtime_admin")
+    execute("GRANT ALL PRIVILEGES ON ALL FUNCTIONS IN SCHEMA realtime TO supabase_realtime_admin")
 
     execute("ALTER table realtime.channels OWNER to supabase_realtime_admin")
     execute("ALTER table realtime.broadcasts OWNER to supabase_realtime_admin")
--- a/lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
+++ b/lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
@@ -16,20 +16,20 @@ defmodule Realtime.Tenants.Migrations.CreateRealtimeAdminAndMoveOwnership do
     execute("ALTER function realtime.channel_name() owner to supabase_realtime_admin")
 
     execute("""
-      DO
-      $do$
-      BEGIN
-         IF NOT EXISTS (SELECT FROM pg_catalog.pg_roles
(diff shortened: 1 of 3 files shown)


chasers (approved):

lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
```diff
@@ -0,0 +1,36 @@
...
+          GRANT anon TO authenticator;
+          GRANT authenticated TO authenticator;
+          GRANT service_role TO authenticator;
+          GRANT postgres TO authenticator;
+       END IF;
+    END
+    $do$;
+    """)
+
+    execute("GRANT supabase_realtime_admin TO authenticator")
```
sweatybridge, line 34:
```suggestion
    execute("GRANT supabase_realtime_admin TO postgres")
```

The `authenticator` role is used by PostgREST to assign one of the `anon`, `authenticated`, or `service_role` roles to the incoming user request based on its JWT token. The reason `storage_storage_admin` needs the `authenticator` role is because it needs to [impersonate](https://github.com/supabase/postgres/pull/590) as one of those 3 roles without going through PostgREST.

We use `grant authenticator to ...` as a shorthand for `grant anon, authenticated, service_role to ...` because it is a [member](https://github.com/supabase/postgres/blob/develop/migrations/db/init-scripts/00000000000000-initial-schema.sql#L29) of all 3 roles. It also future proofs our roles in case we want to add new ones to `authenticator`.

We certainly don't want to grant `supabase_realtime_admin` to `authenticator` because that would allow any users to impersonate as `supabase_realtime_admin` via PostgREST. So line 34 should be removed.

My question is does realtime need to impersonate as one of `anon`, `authenticated`, or `service_role` roles like storage? Looking back at the [slack thread](https://supabase.slack.com/archives/C01M1EGBQC9/p1711796153411699), I believe the original intention is to allow users to create RLS policy through studio dashboard, which always authenticates as `postgres` role. And because only schema owners can create [RLS policy](https://www.postgresql.org/docs/current/ddl-rowsecurity.html), the only way to achieve this is to `grant supabase_realtime_admin to postgres`.

Yes, it does open a can of worms when user accidentally modifies realtime schema but it is not worse than what's already happening to auth and storage. We really need to prioritise the [long term fix](https://www.notion.so/supabase/Revoke-supabase_-_admin-membership-from-postgres-7161e667fb0243f3b14eea4382fff3d9?d=33841a15a9994710803b30901f71594a) which requires an update to supautils to allow non-owners of a schema to create RLS policy.

lib/realtime/tenants/repo/migrations/20240401105812_create_realtime_admin_and_move_ownership.ex
```diff
@@ -0,0 +1,36 @@
...
+  @moduledoc false
+
+  use Ecto.Migration
+
+  def change do
+    execute("CREATE ROLE supabase_realtime_admin WITH NOINHERIT NOLOGIN NOREPLICATION")
+
+    execute("GRANT ALL PRIVILEGES ON SCHEMA realtime TO supabase_realtime_admin")
+    execute("GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA realtime TO supabase_realtime_admin")
+    execute("GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA realtime TO supabase_realtime_admin")
```
sweatybridge, line 11:
```suggestion
    execute("GRANT ALL PRIVILEGES ON ALL SEQUENCES IN SCHEMA realtime TO supabase_realtime_admin")
    execute("GRANT ALL PRIVILEGES ON ALL FUNCTIONS IN SCHEMA realtime TO supabase_realtime_admin")

```

Are we missing a grant on functions here?

PR #829, opened 2024-04-03
fix: Prioritize channel policies; Run migrations on healthcheck

filipecabaco:

Prioritize channel policies; Run migrations on healthcheck

### DIFF
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -77,6 +77,10 @@ defmodule Realtime.Tenants do
         {:ok, %{healthy: true, db_connected: true, connected_cluster: connected_cluster}}
 
       connected_cluster when is_integer(connected_cluster) ->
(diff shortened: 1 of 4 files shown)


lib/realtime/tenants.ex
```diff
@@ -78,8 +69,10 @@ defmodule Realtime.Tenants do
       nil ->
         {:error, :tenant_not_found}
 
-      {:ok, _health_conn} ->
+      {:ok, health_conn} ->
         connected_cluster = UsersCounter.tenant_users(external_id)
+        %{extensions: [%{settings: settings} | _]} = Cache.get_tenant_by_external_id(external_id)
+        Helpers.transaction(health_conn, fn _ -> Migrations.run_migrations(settings) end)
```
chasers, line 75: Needs to run even if db_connected is false

filipecabaco: made the change, not sure if the best approach

chasers (approved):

PR #831, opened 2024-04-04
fix: Broadcast HTTP endpoint respects policies per channel

filipecabaco:

Fix to handle policies on the Broadcast endpoint. We need to check policies one by one in this scenario since the user can broadcast a batch of messages to different channels


chasers (approved):
Well this is unfortunate. Later on perhaps we pull the topic out of the body.

filipecabaco:
yes at a later stage we might be able to improve this ... it feels wasteful at the moment 😞

PR #838, opened 2024-04-10
fix: Add prefix on secure channels

filipecabaco:

To avoid cross communication between public and private channels we will add a prefix on private channels which is determined by checking if a channel exists in realtime.channel with a given name.


chasers (approved):
I’d add a doctest to make it even more obvious but looks good otherwise.

PR #840, opened 2024-04-10
fix: typos

w3b6x9:


filipecabaco (approved):
will need different version but lgmt

PR #841, opened 2024-04-10
fix: Move to UUIDs on realtime schema support tables

filipecabaco:

Move to UUIDs on realtime schema support tables


filipecabaco:
Not needed, closing

PR #842, opened 2024-04-11
fix: Add endpoint to disconnect all users for a given tenant

filipecabaco:

Add endpoint to disconnect all users for a given tenant


lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -244,6 +244,31 @@ defmodule RealtimeWeb.TenantController do
...
+    ],
+    responses: %{
+      200 => TenantHealthResponse.response(),
+      403 => EmptyResponse.response()
+    }
+  )
+
+  def disconnect_clients(%{assigns: %{tenant: tenant}} = conn, %{"tenant_id" => tenant_id}) do
+    Tenants.disconnect_clients(tenant_id)
+    PostgresCdc.stop_all(tenant, @stop_timeout)
```
w3b6x9, line 268: @filipecabaco this returns either `:ok` or `:error`. Shouldn't we include the correct resp code depending on the result of `PostgresCdc.stop_all/2`? That way whatever consumes the endpoint will know if something went wrong.

w3b6x9: Feels inconsistent when taking into account the other places this is called:

- https://github.com/supabase/realtime/blob/4770d8b6cfd1e523d3725eb6f64d0f04e71dc848/lib/realtime_web/controllers/tenant_controller.ex#L193

  - If this returns `:error` then `500` is returned.
  
- https://github.com/supabase/realtime/blob/4770d8b6cfd1e523d3725eb6f64d0f04e71dc848/lib/realtime_web/controllers/tenant_controller.ex#L242
    - Returns 204 no matter what.

I would think that for a proper disconnect upstream would want proper feedback that if failed would get `500` in response.

w3b6x9, line 268: @filipecabaco are there any potential race conditions here?

For example, what would happen if a postgres changes subscribed client gets disconnected and immediately reconnects? This client has a record in `realtime.subscriptions` again upon reconnection. Then when the call to stop PostgresCdc finally goes through would it wipe away the record? Does calling `PostgresCdc.stop_all/2` kick out clients too?

Would it be better to first call `PostgresCdc.stop_all/2` and if it returns `:ok` then call `Tenants.disconnect_clients/1`?

filipecabaco: This is a tricky one as we have no way of doing this without being fully async since we will need to kill potentially thounsands of processes in their own lifecycle.

We can actually suspend the user before hand, wait a couple of seconds (which will already kill socket connections) and then kill postgres cdc

chasers: Yeah probably need to:
- suspend
- bust tenant cache
- stop_all (yeah before disconnect)
- disconnect
- unsuspend

Note: need to bust the tenant cache tho too because suspend flag is there

chasers: If you disconnect and they can't reconnect, then yeah you could stop_all after.

chasers: Looks like suspend busts the tenant cache already 🥳

lib/realtime_web/router.ex
```diff
@@ -104,6 +104,7 @@ defmodule RealtimeWeb.Router do
     pipe_through([:open_cors, :tenant_api, :secure_tenant_api])
 
     post("/broadcast", BroadcastController, :broadcast)
+    delete("/disconnect_clients", TenantController, :disconnect_clients)
```
chasers, line 107: Should we have only a `/suspend` and `/unsuspend` endpoint where it disconnects clients and kills all the db connections including extensions? 

I was having trouble this week figuring out how to restart the `realtime_subscription_manager_pub` connections. 

Having one endpoint where we could confidently disable Realtime for all the db and client conns for a tenant is needed.

filipecabaco: that's a really good , will implement that instead

filipecabaco:
this will be rethinked

PR #861, opened 2024-04-23
fix: Introduce error operational code

filipecabaco:

Introduce error operational code so users can have better information on what is happening within Realtime


README.md
```diff
@@ -172,10 +172,48 @@ Verify JWT claims by setting JWT_CLAIM_VALIDATORS:
...
+| ErrorStartingPostgresCDC           | Error when starting the Postgres CDC extension which is used for Postgres Changes                                                   | Contact support for further instructions                                                           |
+| ReplicationSlotBeingUsed           | The replication slot is being used by another transaction                                                                           | Contact support for further instructions                                                           |
+| PoolingReplicationPreparationError | Error when preparing the replication slot                                                                                           | Contact support for further instructions                                                           |
+| PoolingReplicationError            | Error when pooling the replication slot                                                                                             | Contact support for further instructions                                                           |
```
chasers, line 205: You mean Polling? Not Pooling?

chasers: Also why not snake_case?

chasers: I guess it's like a `defexception` where the exception is the module name?

lib/extensions/postgres_cdc_rls/cdc_rls.ex
```diff
@@ -85,7 +86,8 @@ defmodule Extensions.PostgresCdcRls do
         error
 
       error ->
-        Logger.error("Error starting Postgres Extension: #{inspect(error, pretty: true)}")
+        Logger.error(%{error_code: "ErrorStartingPostgresCDC", error_message: to_log(error)})
```
chasers, line 89: I'm not even sure what these log lines will look like? Just the map? 

Need to think about these in a logging UI like Logflare and Studio.  The entire event_message should be something kind of human readable. 

Most devs don't know what Elixir map notation is or means.

I would do something maybe like:

```elixir
Logger.error( "ErrorStartingPostgresCDC: #{to_log(error)}", error_code: "ErrorStartingPostgresCDC")
```

So the stream would be prefixed with these. Easy to see which you're getting a lot of.

And then if the error code is in the metadata we can group by it if needed.

README.md
filipecabaco: We're complying with the other error codes in Supabase ,that way we keep a single format in use

lib/extensions/postgres_cdc_rls/cdc_rls.ex
filipecabaco: To be honest in console it appears fine but that is a different backend from the prod one so that is a good call.

I will change to your suggestion.

chasers (approved):

PR #863, opened 2024-05-02
fix: Use INSERT instead of UPDATE

filipecabaco:

Uses an INSERT RLS policy instead of a UPDATE to check if user can connect.

This is done by checking if the user was able to insert and it failed due to unique violation so they had permissions to try and insert.


filipecabaco:
we move out this system for authz

PR #864, opened 2024-05-03
fix: Case insensitive Bearer prefix

filipecabaco:

Bearer prefix at the moment was case sensitive which could be a problem with some clients.

### DIFF
--- a/lib/realtime_web/plugs/auth_tenant.ex
+++ b/lib/realtime_web/plugs/auth_tenant.ex
@@ -39,16 +39,27 @@ defmodule RealtimeWeb.AuthTenant do
     authorization = get_req_header(conn, "authorization")
     apikey = get_req_header(conn, "apikey")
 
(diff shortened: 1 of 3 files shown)


lib/realtime_web/plugs/auth_tenant.ex
```diff
@@ -39,16 +39,27 @@ defmodule RealtimeWeb.AuthTenant do
     authorization = get_req_header(conn, "authorization")
     apikey = get_req_header(conn, "apikey")
 
-    cond do
-      authorization != [] && match?(["Bearer " <> _], authorization) ->
-        ["Bearer " <> token] = authorization
-        token
+    authorization =
+      case authorization do
```
abc3, line 43: Is it okay to not cover the case when there is only one element in the list?

abc3 (approved):

PR #870, opened 2024-05-13
fix: Remove unnecessary checks on DB for Broadcast endpoint

filipecabaco:

Remove unnecessary checks on DB for Broadcast endpoint


chasers (approved):
Sure. But will only work until most have enabled it.

filipecabaco:
yes, we will need to change the conn pool setup at a later stage

PR #1029, opened 2024-05-23
fix: Reimplement Authz with newer simpler policies

filipecabaco:

New Authorization checking made with simpler RLS rules. We'll now check a single table called messages and for a specific set of elements for each row.

At the moment we are prefilling the table with some information on lib/realtime/tenants/authorization.ex and that is then used to check read policies and after that we try to insert new values onto the tables to check write policies.

All the changes made during this checks is reverted upon ending the transaction ensuring we don't leave extra clutter.


lib/realtime/tenants.ex
```diff
@@ -231,9 +231,12 @@ defmodule Realtime.Tenants do
...
-  def tenant_topic(external_id, sub_topic, public?) do
-    private_prefix = if public?, do: "", else: ":private"
-    "#{external_id}#{private_prefix}:#{sub_topic}"
+  def tenant_topic(external_id, sub_topic, false) do
+    "#{external_id}:private:#{sub_topic}"
+  end
+
+  def tenant_topic(external_id, sub_topic, true) do
+    "#{external_id}:#{sub_topic}"
   end
```
abc3, line 240:
```suggestion
  def tenant_topic(external_id, sub_topic, true), do: "#{external_id}:#{sub_topic}"
```

lib/realtime/tenants/authorization.ex
```diff
@@ -4,24 +4,22 @@ defmodule Realtime.Tenants.Authorization do
   creates a Realtime.Tenants.Policies struct with the accumulated results of the policies
   for a given user and a given channel context
 
-  Each feature will have their own set of ways to check Policies against the Authorization context.
+  Each feature will have their own set of ways to check Policies against the Authorization context but we will create some setup data to be used by the policies.
```
abc3, line 7: their -> its?
```suggestion
  Each feature will have its own set of ways to check Policies against the Authorization context but we will create some setup data to be used by the policies.
```

lib/realtime/tenants/authorization.ex
```diff
@@ -130,40 +118,72 @@ defmodule Realtime.Tenants.Authorization do
...
-       ) do
-    with {:ok, policies} <-
-           policy_mod.check_write_policies(transaction_conn, policies, authorization_context),
-         {:ok, policies} <-
-           policy_mod.check_read_policies(transaction_conn, policies, authorization_context) do
-      policies
-    else
-      {:error, error} -> Postgrex.rollback(transaction_conn, error)
-    end
+  defp get_write_policy_for_connection_and_feature(conn, policies, authorization_context) do
```
abc3, line 178: just to check if we ever need to debug this function at runtime because it's private

lib/realtime/tenants/authorization/policies/broadcast_policies.ex
```diff
@@ -24,74 +23,64 @@ defmodule Realtime.Tenants.Authorization.Policies.BroadcastPolicies do
...
-        channel: %Channel{id: channel_id}
-      }) do
-    Postgrex.transaction(conn, fn transaction_conn ->
-      query = from(b in Broadcast, where: b.channel_id == ^channel_id)
+  def check_read_policies(conn, %Policies{} = policies, %Authorization{channel_name: channel_name}) do
+    query =
+      from(m in Message,
+        where: m.channel_name == ^channel_name,
+        where: m.feature == :broadcast,
+        limit: 1
```
abc3, line 35: is the channel_name not unique?

lib/realtime/tenants/authorization/policies/broadcast_policies.ex
```diff
@@ -24,74 +23,64 @@ defmodule Realtime.Tenants.Authorization.Policies.BroadcastPolicies do
...
-        {:error, :not_found} ->
-          Policies.update_policies(policies, :broadcast, :write, false)
-      end
-    end)
+  def check_write_policies(conn, policies, %Authorization{channel_name: channel_name}) do
+    changeset =
+      Message.changeset(%Message{}, %{
+        channel_name: channel_name,
+        feature: :broadcast,
+        event: "dummy"
```
abc3, line 68: maybe add a default event name?

lib/realtime/tenants/authorization/policies/presence_policies.ex
```diff
@@ -25,76 +24,66 @@ defmodule Realtime.Tenants.Authorization.Policies.PresencePolicies do
...
-        channel: %Channel{id: channel_id}
-      }) do
-    Postgrex.transaction(conn, fn transaction_conn ->
-      query = from(b in Presence, where: b.channel_id == ^channel_id)
+  def check_read_policies(conn, %Policies{} = policies, %Authorization{channel_name: channel_name}) do
+    query =
+      from(m in Message,
+        where: m.channel_name == ^channel_name,
+        where: m.feature == :presence,
+        limit: 1
```
abc3, line 36: do we really need a limit?

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -53,75 +48,40 @@ defmodule RealtimeWeb.BroadcastController do
...
          events_per_second_key = Tenants.events_per_second_key(tenant),
          :ok <- check_rate_limit(events_per_second_key, tenant, length(messages)) do
-      events =
-        Enum.map(messages, fn %{changes: %{topic: sub_topic} = event} -> {sub_topic, event} end)
-
-      channel_names = events |> Enum.map(fn {sub_topic, _} -> sub_topic end) |> MapSet.new()
+      tenant_db_conn =
+        if tenant.enable_authorization,
+          do: Connect.lookup_or_start_connection(tenant.external_id),
+          else: nil
```
abc3, line 54: `else` is nil by default
```suggestion
```

abc3 (approved):
awesome work 🔥

abc3:
how can I test that with a js client?

lib/realtime/tenants/authorization/policies/broadcast_policies.ex
filipecabaco: no, we will have multiple messages from the same channel stored in this table

lib/realtime/tenants/authorization/policies/broadcast_policies.ex
filipecabaco: yes, unsure of what to be honest 😓

lib/realtime/tenants/authorization/policies/presence_policies.ex
filipecabaco: yes, this table might have a lot of entries but we just need to fetch that the user can read the first one and if so RLS has worked and the user has the correct permissions

lib/realtime_web/controllers/broadcast_controller.ex
filipecabaco: true!

lib/realtime/tenants/authorization.ex
filipecabaco: we can get the same output from `Realtime.Tenants.Authorization.get_authorizations/2` in fact I will need to think on a good tool to test this in admin studio 👀

lib/realtime/tenants/authorization/policies/broadcast_policies.ex
filipecabaco: `check_write_policy` for now, later we might want this to also be something we check

filipecabaco:
@abc3 not yet, will do the changes to the realtime-js next version to be able to test this. in theory we just add the "private" option in the channel startup

chasers (approved):

PR #1071, opened 2024-06-05
fix: Change private topic format

filipecabaco:

Change private topic format

### DIFF
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -225,13 +225,14 @@ defmodule Realtime.Tenants do
   @spec tenant_topic(Tenant.t() | binary(), String.t(), boolean()) :: String.t()
   def tenant_topic(external_id, sub_topic, public? \\ true)
 
-  def tenant_topic(%Tenant{external_id: external_id}, sub_topic, public?) do
-    tenant_topic(external_id, sub_topic, public?)
-  end
+  def tenant_topic(%Tenant{external_id: external_id}, sub_topic, public?),
+    do: tenant_topic(external_id, sub_topic, public?)
 
-  def tenant_topic(external_id, sub_topic, false), do: "#{external_id}-private:#{sub_topic}"
+  def tenant_topic(external_id, sub_topic, false),
+    do: "#{external_id}-private:#{sub_topic}"
 
-  def tenant_topic(external_id, sub_topic, true), do: "#{external_id}:#{sub_topic}"
+  def tenant_topic(external_id, sub_topic, true),
+    do: "#{external_id}:#{sub_topic}"
 
   @doc """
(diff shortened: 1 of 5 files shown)


lib/realtime/tenants.ex
```diff
@@ -231,7 +231,7 @@ defmodule Realtime.Tenants do
     tenant_topic(external_id, sub_topic, public?)
   end
 
-  def tenant_topic(external_id, sub_topic, false), do: "#{external_id}:private:#{sub_topic}"
+  def tenant_topic(external_id, sub_topic, false), do: "#{external_id}:private-#{sub_topic}"
```
chasers, line 234:
```suggestion
  def tenant_topic(external_id, sub_topic, false), do: "#{external_id}-private:#{sub_topic}"
```

Otherwise client could subscribe to topic of `private-topic_name` and receive private channel messages.

With `#{external_id}-private:#{sub_topic}` they cannot.

chasers: Can you add a specific test case for this?

filipecabaco: my bad 🤦 will change and yes will add one for security purposes

w3b6x9 (approved):

test/integration/rt_channel_test.exs
```diff
@@ -263,6 +264,55 @@ defmodule Realtime.Integration.RtChannelTest do
...
+         %{topic: topic} do
+      {anon_socket, _} = get_connection("anon")
+      {socket, _} = get_connection("authenticated")
+      topic = "realtime:#{topic}"
+
+      WebsocketClient.join(socket, topic, %{
+        config: %{broadcast: %{self: true}, private: true}
+      })
+
+      WebsocketClient.join(anon_socket, topic, %{
```
w3b6x9, line 282: @filipecabaco the topic for `private: false` case should end up being `realtime:private:topic` b/c that's the loophole that this PR patches.

w3b6x9 (changes requested):
see: https://github.com/supabase/realtime/pull/1071#discussion_r1635592116

chasers (approved):

test/integration/rt_channel_test.exs
chasers: Should be just `private:topic`

chasers (approved):
Otherwise approved

w3b6x9 (approved):

test/integration/rt_channel_test.exs
filipecabaco: made the change in the test 👍

filipecabaco:
replicated the issues in tests and made the required changes in the test, will merge after CI

PR #866, opened 2024-05-03
fix: Allow Overriding of `max_channels_per_client` via Environment Variable

barrownicholas:
## What kind of change does this PR introduce?

Bug fix: allows overriding the default of `100` for `max_channels_per_client` on a realtime tenant. 

## What is the current behavior?

The default is locked at 100 without any ready way to override it.

## What is the new behavior?

Allows a user to specify `MAX_CHANNELS_PER_CLIENT` in the runtime environment to override this value.

## Additional context

Fixes https://github.com/supabase/realtime/issues/843

Issue #843, opened 2024-04-11
too_many_channels error: `"{:error, :too_many_channels}"`


barrownicholas:
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I'm occasionally getting a `"{:error, :too_many_channels}"` on self-hosted/local. Is there a way to increase the number of channels (env var, etc.)? I saw `MAX_CONNECTIONS` but that doesn't seem right (channels v. connections). 

### Edit 1
It looks like there is `max_channels_per_client`, but I'm not sure/don't see the ability to set this dynamically. I'm not familiar enough with the realtime component to know if this loads dynamically or is set in the db. (It looks like a db row?)

### Edit 2
Okay so the true dev/local using the CLI does **not** use the same as self-hosted/docker `supabase/realtime`... I checked a deployment, and I see below. Is there a way to override the default of `100`?

<img width="1624" alt="Screenshot 2024-04-10 at 10 53 54 PM" src="https://github.com/supabase/realtime/assets/75350582/dc284b3c-de3b-4623-8a8e-a9b21c55678d">

barrownicholas:
Hi All: would it be possible to get a simple `MAX_CLIENTS` variable to override the hard-coded default of 100?

barrownicholas:
I attempted a solution ^

filipecabaco:
this is an issue with CLI environment correct?

barrownicholas:
I get this in our production (self-hosted) environment as well, although theoretically this would exist with the Supabase-hosted too (just given that the default is at 100 with no way of overriding it, aside from the linked PR I submitted).

filipecabaco:
when you use it in self-hosted, do you use the seeded user? would increasing there be enough?

barrownicholas:
@filipecabaco I was trying to brainstorm a solution (like https://github.com/supabase/realtime/pull/866) that would work for both local and self-hosted if that makes sense; I think a CLI solution does definitely fix my localhost CLI issues, it just would leave the self-hosted.

filipecabaco:
sorry for the long silence, tackling other subjects 😓 

for self hosted I think this would also work as at the moment we seed the system with the same file if I'm not mistaken 🤔 cc @w3b6x9

[...]

PR #1078, opened 2024-06-19
Override Default Tenant Limits via Env Vars

barrownicholas:

Feature: in continuation of the conversation [here](https://github.com/supabase/realtime/pull/866#issuecomment-2176530588) with @filipecabaco , this PR allows overriding a host of the default limits (used when creating tenants for the first time) using environment variables. This will greatly simplify self-hosting initial-setup. 

Tenant default values are hard-coded and need to be manually changed **after** the tenant is created, making for a more difficult onboarding process for self-hosted users. 

## What is the new behavior?

Specifying certain `TENANT_...` environment variables overrides the default values whenever a tenant is created. 

## Additional context

This is just an extension of the solution used in #843 and #866


mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.29.10",
+      version: "2.30.0",
```
barrownicholas, line 7: I went with a `minor` bump instead of a `patch` since this was a bit more substantative, but happy to use `2.29.11` if that's better

filipecabaco: lets go with `2.29.11` for now as all of this changes are non-breaking additive elements

barrownicholas:
@filipecabaco I bumped the version back to `2.29.11` so this should be good to go when you have a sec to look at it

filipecabaco (approved):

PR #1082, opened 2024-06-25
fix: Remove flag; Add seed only flag

filipecabaco:

Removed flag for authorization
Added a new way to start the container only for seeding


priv/repo/migrations/20240625211759_remove_enable_authorization_flag.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Realtime.Repo.Migrations.RemoveEnableAuthorizationFlag do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tenants) do
+      remove :enable_authorization
+    end
+  end
+end
```
abc3, line 9: Need to call it in Realtime.Tenants.Migrations, no?

filipecabaco: no, this are on realtime itself

abc3 (approved):

PR #1085, opened 2024-07-01
fix: Fix improper policy checking

filipecabaco:

* Presence had a bad check when tracking changes
* JWT checking had a bad check and a broken test
* Fixed some migrations to be less susceptible to errors on migrations
* Added one connection to db_pool
* Removed one connection from subcriber_pool_size

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -239,6 +239,8 @@ defmodule RealtimeWeb.RealtimeChannel do
 
   @impl true
   def handle_info(:confirm_token, %{assigns: %{pg_change_params: pg_change_params}} = socket) do
+    IO.inspect(socket.policies, label: "confirm_token")
+
     case confirm_token(socket) do
       {:ok, claims, confirm_token_ref, _, _} ->
         pg_change_params = Enum.map(pg_change_params, &Map.put(&1, :claims, claims))
@@ -249,9 +251,11 @@ defmodule RealtimeWeb.RealtimeChannel do
            pg_change_params: pg_change_params
          })}
 
-      {:error, error} ->
-        message = "Access token has expired: " <> Helpers.to_log(error)
+      {:error, error} when is_binary(error) ->
+        shutdown_response(socket, error)
 
+      {:error, error} ->
+        message = Helpers.to_log(error)
         shutdown_response(socket, message)
     end
   end
@@ -328,6 +332,9 @@ defmodule RealtimeWeb.RealtimeChannel do
         }
 
         {:noreply, assign(socket, assigns)}
+
+      {:error, error} when is_binary(error) ->
+        shutdown_response(socket, error)
 
       {:error, error} ->
         message = "Received an invalid access token from client: " <> inspect(error)
@@ -486,17 +493,15 @@ defmodule RealtimeWeb.RealtimeChannel do
          exp_diff when exp_diff > 0 <- exp - Joken.current_time(),
          {:ok, socket} <- validate_policy(socket, claims, check_policy) do
       if ref = assigns[:confirm_token_ref], do: Helpers.cancel_timer(ref)
+
       interval = min(@confirm_token_ms_interval, exp_diff * 1_000)
       ref = Process.send_after(self(), :confirm_token, interval)
+
       {:ok, claims, ref, access_token, socket}
     else
       {:error, e} -> {:error, e}
       e -> {:error, e}
     end
-  end
-
-  defp validate_policy(%{assigns: %{check_authorization?: true}} = socket, _claims, _check_policy) do
-    {:ok, socket}
   end
 
   defp validate_policy(socket, _claims, false) do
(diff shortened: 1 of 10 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -239,6 +239,8 @@ defmodule RealtimeWeb.RealtimeChannel do
 
   @impl true
   def handle_info(:confirm_token, %{assigns: %{pg_change_params: pg_change_params}} = socket) do
+    IO.inspect(socket.policies, label: "confirm_token")
+
```
chasers, line 243:
```suggestion
```

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -486,19 +493,17 @@ defmodule RealtimeWeb.RealtimeChannel do
          exp_diff when exp_diff > 0 <- exp - Joken.current_time(),
          {:ok, socket} <- validate_policy(socket, claims, check_policy) do
```
chasers, line 494: In the join we have a pipe of (pseudo) `|> confirm_token |> assign_policy` and it's obvious that policies get assigned to the socket this way.

I would call `assign_policy` from the `access_token` `handle_in` callback so it's obvious policies are assigned there.

Right now this `validate_policy` is maybe called, and then it maybe calls `assign_policy`.

chasers (changes requested):

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: did a refactor to check the policy jwt change, could you recheck?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -312,25 +312,25 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-            _ -> nil
-          end
-
-        assigns = %{
-          pg_sub_ref: pg_sub_ref,
-          confirm_token_ref: confirm_token_ref,
-          pg_change_params: pg_change_params
-        }
+    with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
+         {:ok, socket} <- validate_policy(socket, claims) do
```
chasers, line 316: Call `validate_policy` `maybe_assign_policy`? It's not really validating anything.

chasers (approved):

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -310,24 +312,27 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        pg_sub_ref =
-          case pg_change_params do
-            [_ | _] -> postgres_subscribe()
-            _ -> nil
-          end
+      pg_sub_ref =
+        case pg_change_params do
+          [_ | _] -> postgres_subscribe()
+          _ -> nil
+        end
```
abc3, line 324:
```suggestion
      pg_sub_ref = if pg_change_params != [], do: postgres_subscribe()
```

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -249,9 +249,11 @@ defmodule RealtimeWeb.RealtimeChannel do
            pg_change_params: pg_change_params
          })}
 
-      {:error, error} ->
-        message = "Access token has expired: " <> Helpers.to_log(error)
+      {:error, error} when is_binary(error) ->
+        shutdown_response(socket, error)
 
+      {:error, error} ->
+        message = Helpers.to_log(error)
```
abc3, line 256: just checking if really don't need "Access token has expired: "

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -310,24 +312,27 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        pg_sub_ref: pg_sub_ref,
+        confirm_token_ref: confirm_token_ref,
+        pg_change_params: pg_change_params
+      }
 
-        {:noreply, assign(socket, assigns)}
+      {:noreply, assign(socket, assigns)}
+    else
+      {:error, error} when is_binary(error) ->
+        shutdown_response(socket, error)
```
abc3, line 335: missed "Received an invalid access token from client:"?
```suggestion
        shutdown_response(socket, "Received an invalid access token from client: " <> error)
```

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: In this case I'm not sure honestly 🤔

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: I think we might need to rethink this error 🤔 more on the side of "Your token as been invalidated" or something like that

chasers (approved):

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: added back

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: will probably tackle this in another PR

fullstackwebdev:
this is broken code

PR #1086, opened 2024-07-02
feat: Support LISTEN broadcast in Realtime

filipecabaco:

Support LISTEN/NOTIFY on `realtime:broadcast` which allows users to trigger broadcast messages by sending a payload to this topic.

```sql
select pg_notify(
          'realtime:broadcast',
          json_build_object(
              'private', $1::boolean,
              'topic', $2::text,
              'event', $3::text,
              'payload', $4::jsonb
          )::text
      );
```

We want to keep it as close to the Broadcast endpoint as possible and it already includes the feature flag `notify_private_alpha` to enabled it (defaults to false)

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -16,6 +16,7 @@ defmodule Realtime.Tenants.Connect do
   alias Realtime.Database
   alias Realtime.Rpc
   alias Realtime.Tenants
+  alias Realtime.Tenants.Listen
   alias Realtime.Tenants.Migrations
   alias Realtime.UsersCounter
 
@@ -26,6 +27,7 @@ defmodule Realtime.Tenants.Connect do
   defstruct tenant_id: nil,
             db_conn_reference: nil,
             db_conn_pid: nil,
+            listen_pid: nil,
             check_connected_user_interval: nil,
             connected_users_bucket: [1]
 
@@ -83,7 +85,9 @@ defmodule Realtime.Tenants.Connect do
       check_connected_user_interval: check_connected_user_interval
     }
 
-    GenServer.start_link(__MODULE__, state, name: {:via, :syn, name})
+    opts = Keyword.put(opts, :name, {:via, :syn, name})
+
+    GenServer.start_link(__MODULE__, state, opts)
   end
 
   ## GenServer callbacks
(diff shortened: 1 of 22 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -121,8 +124,26 @@ defmodule Realtime.Tenants.Connect do
...
+  end
 
-    {:noreply, state}
+  def handle_continue(
+        :setup_listen,
+        %{tenant_id: tenant_id} = state
+      ) do
+    tenant = Tenants.get_tenant_by_external_id(tenant_id)
+
+    if tenant.notify_private_alpha do
```
chasers, line 136: Nice

lib/realtime/tenants/connect.ex
```diff
@@ -164,6 +185,26 @@ defmodule Realtime.Tenants.Connect do
     {:noreply, state}
   end
 
+  def handle_info(
+        {:notification, _, _, "realtime:broadcast", payload},
```
chasers, line 189: Feels like this is odd in the connect module. Too much responsibility for it, but nit for now.

lib/realtime_web/controllers/broadcast_controller.ex
```diff
@@ -43,89 +32,8 @@ defmodule RealtimeWeb.BroadcastController do
...
-            Enum.each(events, fn %{topic: sub_topic, payload: payload, event: event} ->
-              send_message_and_count(tenant, sub_topic, event, payload, false)
-            end)
-
-          _ ->
-            nil
-        end
-      end)
-
+    with :ok <- Realtime.Tenants.BatchBroadcast.broadcast(conn, tenant, attrs) do
```
chasers, line 35: Good helper

abc3:
@filipecabaco when does the `Listen` connection stop?

filipecabaco:
> @filipecabaco when does the `Listen` connection stop?

The processes are linked so I would expect it to stop the connection but worth to check! I will try to do a test to check the behaviour for sanity sake 👍

lib/realtime/tenants/connect.ex
filipecabaco: agreed, will try to check what I can do to move this out of connect (start the db conn here and move the listen somewhere else 🤔 )

chasers (changes requested):
I was thinking this should be a separate extension where `event` would be `listen_notify` or similar.

And users would have to include this `extension` in their RLS policies. 

I feel like this warrants a separate section in the docs and this would make it a real "extension".

Clients should not be able to send each other `listen_notify` messages so we could filter these out this way also.

chasers (approved):

filipecabaco:
@abc3 good call on the test, without restart: :transient we had the process auto restart wrongly, added a test to check what happens when we suspend a tenant (which kills conns)

psteinroe:
just saw this on my timeline, and I think it should be noted in the docs for this that there is a relatively low limit on the notify payload size.

filipecabaco:
@psteinroe yes, 8kb. we want to give a new channel of communication that is more native to the database vs pg_net

we'll need to be careful to properly explain in the docs and this won't be launched for a while as this will be "private alpha" and feature flagged

PR #1089, opened 2024-07-09
fix: Longpoll configuration setting properly session

filipecabaco:

Longpoll configuration setting properly session

### DIFF
--- a/lib/realtime_web/endpoint.ex
+++ b/lib/realtime_web/endpoint.ex
@@ -20,7 +20,13 @@ defmodule RealtimeWeb.Endpoint do
         {Phoenix.Socket.V2.JSONSerializer, "~> 2.0.0"}
       ]
     ],
-    longpoll: true
+    longpoll: [
+      connect_info: [:peer_data, :uri, :x_headers],
+      serializer: [
+        {Phoenix.Socket.V1.JSONSerializer, "~> 1.0.0"},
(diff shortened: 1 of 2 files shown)


chasers (approved):

chasers:
I was thinking this should be a separate extension where `event` would be `listen_notify` or similar.

And users would have to include this `extension` in their RLS policies. 

I feel like this warrants a separate section in the docs and this would make it a real "extension".

Clients should not be able to send each other `listen_notify` messages so we could filter these out this way also.

chasers (changes requested):
I was thinking this should be a separate extension where `event` would be `listen_notify` or similar.

And users would have to include this `extension` in their RLS policies. 

I feel like this warrants a separate section in the docs and this would make it a real "extension".

Clients should not be able to send each other `listen_notify` messages so we could filter these out this way also.

filipecabaco:
> I was thinking this should be a separate extension where `event` would be `listen_notify` or similar.
> 
> And users would have to include this `extension` in their RLS policies.
> 
> I feel like this warrants a separate section in the docs and this would make it a real "extension".
> 
> Clients should not be able to send each other `listen_notify` messages so we could filter these out this way also.

is this for this PR or the other one? 😅

chasers (approved):

chasers:
Ya yep, approved.

PR #1104, opened 2024-07-13
fix: Avoid migration conflicts if table exists

filipecabaco:

Avoid migration conflicts if table exists


filipecabaco:
will tackle this differently by avoiding concurrent migration processes ( https://github.com/supabase/realtime/pull/1109 ) and later we will squash migrations

PR #1105, opened 2024-07-15
fix: log unknown Broadcast errors

chasers:
- adds a catchall in the FallbackController and logs unknown errors

### DIFF
--- a/lib/realtime_web/controllers/fallback_controller.ex
+++ b/lib/realtime_web/controllers/fallback_controller.ex
@@ -56,7 +56,7 @@ defmodule RealtimeWeb.FallbackController do
   end
 
   def call(conn, response) do
-    Helpers.log_error("UnknownError",response)
+    Helpers.log_error("UnknownError", response)
 
     conn
     |> put_status(:not_found)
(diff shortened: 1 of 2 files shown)


lib/realtime_web/controllers/fallback_controller.ex
```diff
@@ -54,4 +54,13 @@ defmodule RealtimeWeb.FallbackController do
     |> put_view(RealtimeWeb.ChangesetView)
     |> render("error.json", changeset: changeset)
   end
+
+  def call(conn, response) do
+    Helpers.log_error("UnknownError",response)
+
+    conn
+    |> put_status(:not_found)
```
filipecabaco, line 62: this might be better as a 422 and not a 404

filipecabaco (approved):

lib/realtime_web/controllers/fallback_controller.ex
chasers: That is for the TenantController for when tenants are not found though

chasers: Oh...

PR #1112, opened 2024-08-02
fix: Improvements to Authz flow

filipecabaco:

* Insert All vs single inserts when adding the entries to be checked
* Removed TopicPolicies as Broadcast is doing the same work
* Removed rollbacks from policies, leaving only one rollback at the end of the policy checking


abc3 (approved):

chasers (approved):

chasers:
Tests are flaky, merging...

PR #1115, opened 2024-08-03
fix: avoid pool size override

filipecabaco:

Avoid pool size override

### DIFF
--- a/lib/realtime/database.ex
+++ b/lib/realtime/database.ex
@@ -102,7 +102,7 @@ defmodule Realtime.Database do
   If `use_settings` is set to `false`, the connection pool will default to 1.
   If `use_settings` is set to `true` and settings has the key `db_pool`, the connection pool will be set to the value of `db_pool`.
   """
-  @spec check_tenant_connection(Tenant.t(), binary(), integer() | nil) ::
+  @spec check_tenant_connection(Tenant.t(), binary(), boolean()) ::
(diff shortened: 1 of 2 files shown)


chasers (approved):

chasers:
Approved for now, but can we have a test where the actual pool size(s) are what we are setting?

w3b6x9 (changes requested):
before this goes in let's make sure `check_tenant_connection(tenant, "replication_slot_teardown")` uses just one db conn.

lib/realtime/database.ex
```diff
@@ -98,17 +98,24 @@ defmodule Realtime.Database do
   @cdc "postgres_cdc_rls"
   @doc """
   Checks if the Tenant CDC extension information is properly configured and that we're able to query against the tenant database.
+
+  If `use_settings` is set to `false`, the connection pool will default to 1.
+  If `use_settings` is set to `true` and settings has the key `db_pool`, the connection pool will be set to the value of `db_pool`.
   """
-  @spec check_tenant_connection(Tenant.t(), binary(), integer()) ::
+  @spec check_tenant_connection(Tenant.t(), binary(), integer() | nil) ::
```
w3b6x9, line 105:
```suggestion
  @spec check_tenant_connection(Tenant.t(), binary(), boolean()) ::
```

w3b6x9 (approved):

PR #1116, opened 2024-08-05
fix: Sanitize logs and refactor database connection

filipecabaco:

Currently logs from transactions where not being properly managed so this fixes the issue

Also took the opportunity to clean up the code from the Database module to make it less error prone

### DIFF
--- a/lib/realtime/database.ex
+++ b/lib/realtime/database.ex
@@ -2,6 +2,7 @@ defmodule Realtime.Database do
   @moduledoc """
   Handles tenant database operations
   """
+  import Realtime.Helpers, only: [log_error: 2]
 
   alias Realtime.Api.Tenant
   alias Realtime.Crypto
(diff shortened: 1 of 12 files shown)


lib/realtime/database.ex
```diff
@@ -133,6 +126,14 @@ defmodule Realtime.Database do
     end)
   end
 
+  def connect(tenant, application_name, pool, backoff \\ :stop) do
+    tenant
+    |> then(&Realtime.PostgresCdc.filter_settings(@cdc, &1.extensions))
+    |> then(&Realtime.Database.from_settings(&1, application_name, backoff))
+    |> then(&%{&1 | pool: pool})
```
chasers, line 133: This is defo still confusing because some pool sizes we set in the tenant record and some we override here?

chasers (approved):

lib/realtime/database.ex
filipecabaco: yeah we need to potentially refactor more to have a better way to handle this but this at least already provides a better escape hatch and more concise code

PR #1119, opened 2024-08-06
fix: Avoid matching on authorization error

filipecabaco:

We were matching directly which made the GenServer crash. This will avoid it and it will bubble up the transactional error. Plus we were wrapping too many tuples on authorization calculations.


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -651,18 +651,18 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        {:error, "You do not have permissions to read Broadcast messages from this channel"}
+        using_broadcast? &&
+            match?(%Policies{broadcast: %BroadcastPolicies{read: false}}, socket.assigns.policies) ->
+          {:error, "You do not have permissions to read Broadcast messages from this channel"}
 
-      true ->
-        {:ok, socket}
+        true ->
+          {:ok, socket}
+      end
```
w3b6x9, line 665: we should have an `else` block to catch `Authorization.get_authorizations/3` errors.

lib/realtime/tenants/authorization.ex
```diff
@@ -57,7 +57,7 @@ defmodule Realtime.Tenants.Authorization do
   def get_authorizations(%Phoenix.Socket{} = socket, db_conn, authorization_context) do
     case get_authorizations(db_conn, authorization_context) do
       %Policies{} = policies -> {:ok, Phoenix.Socket.assign(socket, :policies, policies)}
-      error -> {:error, error}
+      error -> error
```
w3b6x9, line 60: this move may break the return contract: https://github.com/supabase/realtime/pull/1119/files#diff-9542ac1036d63c2de0cf927b9cebf2c4d045ce6388432f56a0ba26ca83920c8bR56

w3b6x9 (changes requested):
isn't the problem that we're not returning {:error, error} here? https://github.com/supabase/realtime/blob/6dbb2b637ef473002620e070c2f6dfdb45c15cc1/lib/realtime/database.ex#L230

that error return doesn't match the caught return error here: https://github.com/supabase/realtime/blob/6dbb2b637ef473002620e070c2f6dfdb45c15cc1/lib/realtime/database.ex#L235

and that's going to lead to issues. better to standardize around returning `{:error, error}`.

filipecabaco:
I think it's two fold:

* we do need to return the error tuple but the other issue was the match on only :ok which would raise a match error in realtime channel killing the process

filipecabaco:
actually we're just returning what Postgrex.transaction returns and that is an error the already 🤔 I will still check

w3b6x9:
@filipecabaco example of what we could do: https://github.com/supabase/realtime/commit/44d48b120c58507525648d9c8f306271485240af

lib/realtime/messages.ex
```diff
@@ -14,7 +14,7 @@ defmodule Realtime.Messages do
   def create_message(attrs, conn, opts \\ [mode: :savepoint]) do
     channel = Message.changeset(%Message{}, attrs)
 
-    result =
+    {:ok, result} =
```
w3b6x9, line 17: @filipecabaco do you intend for this to raise do to left not matching right?

otherwise, should stick with `result = ...` for this and then update the `case result do...` below:

```elixir
case result do
  %Ecto.Changeset{valid?: false} = error -> {:error, error}
  {:error, error} -> {:error, error}
  {:ok, result} -> {:ok, result}
end
```

lib/realtime_web/channels/realtime_channel.ex
w3b6x9: covered in https://github.com/supabase/realtime/pull/1120

w3b6x9 (approved):
lgtm!

PR #1120, opened 2024-08-06
fix: handle erpc call errors and raises

w3b6x9:


lib/realtime/rpc.ex
```diff
@@ -31,22 +31,34 @@ defmodule Realtime.Rpc do
...
-      [:realtime, :rpc],
-      %{latency: latency},
-      %{
-        mod: mod,
-        func: func,
-        target_node: node,
-        origin_node: node()
-      }
-    )
+    try do
```
w3b6x9, line 38: @filipecabaco got this from [Supavisor](https://github.com/supabase/supavisor/blob/9f56b879e7421a6245846d3b9a61dc81756d6afd/lib/supavisor/helpers.ex#L330).

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -656,13 +656,15 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        using_broadcast? &&
+        using_broadcast? and
             match?(%Policies{broadcast: %BroadcastPolicies{read: false}}, socket.assigns.policies) ->
           {:error, "You do not have permissions to read Broadcast messages from this channel"}
 
         true ->
           {:ok, socket}
       end
+    else
+      error -> error
```
filipecabaco, line 667: this won't do anything as with already "pops up" the error

filipecabaco (approved):

PR #1121, opened 2024-08-07
fix: Allow setting selfhost jwt jwks value

filipecabaco:

Allow setting selfhost jwt jwks value


abc3 (approved):

hf (approved):
Looks good!

PR #1131, opened 2024-08-26
fix: Add dump process data function

filipecabaco:

To help debugging potential issues with processes being created without restrictions were adding a function to help debug those issues.

### DIFF
--- a/lib/realtime/operations.ex
+++ b/lib/realtime/operations.ex
@@ -98,4 +98,9 @@ defmodule Realtime.Operations do
       # credo:disable-for-next-line
       IO.inspect({"Failed to stop", tenant, kind, reason})
   end
+
+  def dump_processes do
+    term = Process.list() |> Enum.map(&Process.info/1) |> :erlang.term_to_binary()
+    File.write!("/tmp/erlang_process_dump.csv", term)
+  end
 end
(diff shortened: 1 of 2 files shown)


abc3 (approved):

lib/realtime/operations.ex
```diff
@@ -98,4 +98,58 @@ defmodule Realtime.Operations do
       # credo:disable-for-next-line
       IO.inspect({"Failed to stop", tenant, kind, reason})
   end
+
+  def dump_processes do
+    {:ok, device} = File.open("/tmp/erlang_process_dump.csv", [:write, :utf8])
+
+    device
+    |> IO.write(
+      "intial_call\tstatus\tmemory\tstack_size\tcurrent_stacktrace\tmessage_queue_len\tmessages\n"
```
chasers, line 107: You can do `term_to_binary` and just write the whole term from `Process.list()`

chasers (changes requested):
I'd also like to see an `analyze` fn which lets you give it a process list and groups processes by something in the dictionary and gives you counts of each group.

lib/realtime/operations.ex
filipecabaco: 👍 will change

chasers (approved):

PR #1137, opened 2024-08-29
fix: Increase RPC timeout

filipecabaco:

Currently we're getting a lot of RPC timeouts due to the fact that the Postgrex and RPC timeout time matches. We should avoid have overlapping timeouts as it can potencially create issues and will "drown" errors on RPC calls.


lib/realtime/rpc.ex
```diff
@@ -10,7 +10,7 @@ defmodule Realtime.Rpc do
   """
   @spec call(atom(), atom(), atom(), any(), keyword()) :: any()
   def call(node, mod, func, args, opts \\ []) do
-    timeout = Keyword.get(opts, :timeout, 30_000)
+    timeout = Keyword.get(opts, :timeout, Application.get_env(:realtime, :rpc_timeout))
```
abc3, line 13: maybe at compile time? with Application.compile_env

abc3:
```elixir
@rpc_timeout Application.compile_env(:realtime, :rpc_timeout)
```

PR #1142, opened 2024-09-02
feat: add sharding for the :users scope

abc3:


lib/realtime/syn_shards.ex
```diff
@@ -0,0 +1,59 @@
...
+          else: :syn.group_names(shard)
+
+      group_names ++ acc
+    end)
+    |> Enum.uniq()
+  end
+
+  @spec join(atom(), term(), pid()) :: :ok | {:error, term()}
+  def join(scope, name, pid) do
+    scope
```
chasers, line 31: Shard by pid instead? Otherwise a tenant will be pinned to one syn process.

abc3: Damn, such a stupid mistake. Of course, sharing by name doesn’t make sense in our case 🤦‍♂️🤦‍♂️🤦‍♂️

filipecabaco (approved):
this almost feels like it should be part of `:syn` itself 😅

PR #1146, opened 2024-09-04
fix: Setup busy wait variable

filipecabaco:

Sets busy wait to none to avoid holding on to scheduler compute. We're also reverting the syn partitioning for the time being.


rel/vm.args.eex
```diff
@@ -15,3 +15,6 @@
 
 ## Set distribution buffer busy limit (default is 1024)
 +zdbbl 100000
+
+## Disable Busy Wait
++sbwt none
```
chasers, line 20:
```suggestion
+sbwt none
+sbwtdio none
+sbwtdcpu none
```

chasers (changes requested):

chasers (approved):

PR #1151, opened 2024-09-11
fix: Shutdown Connect when migrations fail

filipecabaco:
Currently we had an issue where Connect module was not shutting down properly leading to more and more connections to be created against the tenant database. Now we use reason :shutdown to enforce the process is killed.

Also the existing Listen/Notify approach will be deleted for a different approach to be built in the future.


lib/realtime/tenants/connect.ex
```diff
@@ -97,37 +95,21 @@ defmodule Realtime.Tenants.Connect do
     Logger.metadata(external_id: tenant_id, project: tenant_id)
 
     with %Tenant{} = tenant <- Tenants.get_tenant_by_external_id(tenant_id),
-         res <- Database.check_tenant_connection(tenant, @application_name),
+         {:ok, conn} <- Database.check_tenant_connection(tenant, @application_name),
+         ref = Process.monitor(conn),
```
abc3, line 99: Is it worth using link() instead of monitor()?

abc3 (approved):
LGTM

lib/realtime/tenants/connect.ex
filipecabaco: We're using monitor here to ensure that we track the connection shutdown and handle it to also kill the Connect process

PR #1156, opened 2024-09-17
feat: Add broadcast functionality from triggers

filipecabaco:

Adds a new functionality to broadcast db changes. It adds:
* Adds `uuid` column to `realtime.messages` so we can later change all ids from serial id to uuid format
* Adds `realtime.send` postgres function to broadcast messages
* Adds `realtime.broadcast_changes` postgres function to broadcast broadcast changes specifically and be used in trigger functions 
* Adds BroadcastChanges feature by tracking the WAL of `realtime.messages`

Example:
```sql
CREATE TABLE test_table (
    id serial PRIMARY KEY,
    name text NOT NULL,
    value integer NOT NULL,
    created_at timestamp DEFAULT CURRENT_TIMESTAMP
);

CREATE OR REPLACE FUNCTION broadcast_changes_for_test_table_trigger ()
    RETURNS TRIGGER
    AS $$
DECLARE
    topic text;
BEGIN
    topic = 'event:' || COALESCE(NEW.id::text, OLD.id::text);
    PERFORM
        realtime.broadcast_changes (topic, TG_OP, TG_OP, TG_TABLE_NAME, TG_TABLE_SCHEMA, NEW, OLD, TG_LEVEL);
    RETURN NULL;
END;
$$
LANGUAGE plpgsql;

CREATE TRIGGER broadcast_changes_for_test_table
    AFTER INSERT OR UPDATE OR DELETE ON test_table
    FOR EACH ROW
    EXECUTE FUNCTION broadcast_changes_for_test_table_trigger ();

-- Insert operation
INSERT INTO test_table (name, value) VALUES ('example_name', 42);

-- Update operation
UPDATE test_table SET value = 100 WHERE name = 'example_name';

-- Delete operation
DELETE FROM test_table WHERE name = 'example_name';

-- Send a message to the topic
SELECT realtime.send (to_jsonb ('{}'::text), 'test', 'test', FALSE);

```

```
   topic  | extension |     inserted_at     |     updated_at      |                                                                                                                                       payload                                                                                                                                       | event  | private |                  id                  
---------+-----------+---------------------+---------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------+---------+--------------------------------------
 event:1 | broadcast | 2024-09-19 15:55:51 | 2024-09-19 15:55:51 | {"table": "test_table", "record": {"id": 1, "name": "example_name", "value": 42, "created_at": "2024-09-19T15:55:50.539754"}, "schema": "public", "operation": "INSERT", "old_record": null}                                                                                        | INSERT | t       | 39d6a1d8-04a4-4099-a2a3-e3c5a9b4605b
 event:1 | broadcast | 2024-09-19 15:55:53 | 2024-09-19 15:55:53 | {"table": "test_table", "record": {"id": 1, "name": "example_name", "value": 100, "created_at": "2024-09-19T15:55:50.539754"}, "schema": "public", "operation": "UPDATE", "old_record": {"id": 1, "name": "example_name", "value": 42, "created_at": "2024-09-19T15:55:50.539754"}} | UPDATE | t       | a34a5f73-13d2-4d11-b4ca-aaf041f8ad7a
 event:1 | broadcast | 2024-09-19 15:55:56 | 2024-09-19 15:55:56 | {"table": "test_table", "record": null, "schema": "public", "operation": "DELETE", "old_record": {"id": 1, "name": "example_name", "value": 100, "created_at": "2024-09-19T15:55:50.539754"}}                                                                                       | DELETE | t       | 6f5a291c-37b4-4051-96f4-98accdc5b201
 test    | broadcast | 2024-09-19 15:55:58 | 2024-09-19 15:55:58 | "{}"                                                                                                                                                                                                                                                                                | test   | f       | 6cc07f9e-4a59-42dd-9aa1-2d1b15e7827d
```


lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
```diff
@@ -0,0 +1,73 @@
...
+    BEGIN
+        INSERT INTO realtime.messages (payload, event, topic, extension)
+        VALUES (payload, event, topic, extension);
+    END;
+    $$
+    LANGUAGE plpgsql;
+    """
+
+    execute """
+    CREATE OR REPLACE FUNCTION realtime.broadcast ()
```
chasers, line 27: Should this fn maybe be `realtime.broadcast_change`?

filipecabaco: will change 👍 otherwise it might be too abstract

chasers:
Should our payload match [what we have here](https://github.com/supabase/cli/blob/9c1b27a4ccc133270288f7dd19d33828e5e796e7/internal/utils/templates/initial_schemas/14.sql#L904) in the Webhooks trigger?

olirice (commented):
It wouldn't hurt to have a check like
```sql
  IF TG_LEVEL = 'STATEMENT' THEN
    RAISE EXCEPTION 'realtime.broadcast_changes should be triggered for each row, not for each statement';
  END IF;
```
to make sure no one tries to "optimize" their trigger for bulk operations

lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
w3b6x9: I like realtime.broadcast bc then I can use it via Postgrest or something to broadcast anything. realtime.broadcast_change seems very specific

lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
```diff
@@ -0,0 +1,59 @@
...
+    execute """
+    CREATE OR REPLACE FUNCTION realtime.broadcast_change ()
+        RETURNS TRIGGER
+        AS $$
+    DECLARE
+        -- Declare a variable to hold the JSONB representation of the row
+        row_data jsonb := '{}'::jsonb;
+        -- Declare entry that will be written to the realtime.messages table
+        topic_name text := TG_ARGV[0]::text;
+        event_name text := COALESCE(TG_ARGV[1]::text, TG_TABLE_SCHEMA || '.' || TG_TABLE_NAME);
```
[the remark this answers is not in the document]
chasers: Our event names are arbitrary, this is just prescriptive for broadcasting changes from tables.

@filipecabaco we could just not make this an argument for now. One less thing for devs to think about.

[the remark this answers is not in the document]
filipecabaco: the idea is to have this be flexible so if a user wants to have multiple tables to be funnelled into a single topic with a single event they are able to. just giving extra flexibility (and an extra gun to potentially lose a foot to be fair)

I'll bring this into the discussion

lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
chasers: This is a trigger function which specifically broadcasts changes from a table tho.

filipecabaco: that would be the `send` function we have above. we could rename it to be `broadcast` and that one indeed can send anything detached from pg operations

chasers: Use `realtime.send` anywhere:

```
SELECT realtime.send(
    '{"message": "Hello, world!"}'::jsonb, -- payload
    'new_message',                         -- event
    'some_room_id',                        -- topic
    'broadcast'                            -- extension (optional, defaults to 'broadcast')
);
```

w3b6x9: Or both are called realtime.broadcast with different signatures and that way devs don't have to remember two different function names

chasers: @filipecabaco btw nit but I would do the args like: 

```
SELECT realtime.send(
    'some_room_id',                                         -- topic
    'new_message',                                          -- event
    '{"message": "Hello, world!"}'::jsonb,      -- payload (optional, defaults to '{}')
    'broadcast',                                                 -- extension (optional, defaults to 'broadcast')
    true                                                              -- private (optional, defaults to true)
);
```

chasers: > Or both are called realtime.broadcast with different signatures and that way devs don't have to remember two different function names

Even more confusing imo.

lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
```diff
@@ -0,0 +1,59 @@
...
+    BEGIN
+        -- Ensure topic_name is provided
+        IF topic_name IS NULL THEN
+            RAISE EXCEPTION 'Topic name must be provided';
+        END IF;
+        -- Check the operation type and handle accordingly
+        IF TG_OP = 'INSERT' OR TG_OP = 'UPDATE' OR TG_OP = 'DELETE' THEN
```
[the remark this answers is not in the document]
filipecabaco: added the exception on statement calls 👍 that will prevent this type of situation. thank you for the heads up on this one 🙏

lib/realtime/tenants/repo/migrations/20240917170412_add_payload_to_messages.ex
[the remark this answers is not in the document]
filipecabaco: will keep the event overridable as it might make it easier for users to migrate to this approach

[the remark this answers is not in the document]
olirice: > Our event names are arbitrary, this is just prescriptive for broadcasting changes from tables.

oh nice. In that case this looks like a great solution as long as its documented well

lib/realtime/api/message.ex
```diff
@@ -5,16 +5,22 @@ defmodule Realtime.Api.Message do
...
+    field :event, :string
+    field :private, :boolean
+
     timestamps()
   end
 
   def changeset(message, attrs) do
     message
-    |> cast(attrs, [:topic, :extension, :inserted_at, :updated_at])
+    |> cast(attrs, [:topic, :extension, :payload, :event, :private, :inserted_at, :updated_at])
```
wojtekmach, line 23: I don't think you need to allow casting timestamps, I think they get overwritten by `put_timestamp`, `maybe_put_timestamp` anyway?

```suggestion
    |> cast(attrs, [:topic, :extension, :payload, :event, :private])
```

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,204 @@
...
+      case reply do
+        :now -> standby_status(wal_end, wal_end, wal_end, reply)
+        :later -> hold()
+      end
+
+    {:reply, message}
+  end
+
+  def call(msg, state) do
+    Logger.info("Unknown message received: #{inspect(%{msg: parse(msg), state: state})}")
```
wojtekmach, line 63: nitpick: maybe consider warning so it shows up in prod logs? Seems like a bug that this would get unrelated message?

```suggestion
    Logger.warning("Unknown message received: #{inspect(%{msg: parse(msg), state: state})}")
```

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,204 @@
...
+
+            to_broadcast =
+              %{
+                topic: topic,
+                event: event,
+                private: private,
+                payload: payload
+              }
+
+            buffer = buffer ++ [to_broadcast]
```
wojtekmach, line 117: for large lists appending is expensive, in that case I'd prepend and Enum.reverse before emptying the buffer. If these are rather small lists, nevermind!

wojtekmach (commented):
Looks good to me, I left some minor notes below.

lib/realtime/broadcast_changes/postgres_replication.ex
```diff
@@ -0,0 +1,263 @@
...
+      {:error, {:case_clause, {:disconnect, reason, _}}} -> {:error, reason}
+      {:error, error} -> {:disconnect, error}
+    end
+  end
+
+  @impl true
+  def init(attrs) do
+    Logger.info(
+      "Initializing connection with the status: #{inspect(attrs |> Map.from_struct() |> Map.drop([:connection_opts]))}"
+    )
```
wojtekmach, line 92: btw, since the Inspect implementation already drops :connection_opts, I think you can use it here too:

```suggestion
  def init(%__MODULE__{} = attrs) do
    Logger.info(
      "Initializing connection with the status: #{inspect(attrs, pretty: true}"
    )
```

filipecabaco: nice call! will change 👍

josevalim:
Hi @filipecabaco! This looks exciting!

I have two questions:

Today you start a single replication process, but as far as I understand, it has a single handler, and the handler only looks into a single key in the registry. In this case, what value is the registry buying? Is it in case someone reconnects, you get will get a new entry in the registry and dispatch to that? And what happens if multiple clients are trying to subscribe to the same replication? Also, should the replication connection eventually terminate if no one is connected?

lib/realtime/tenants/connect.ex
```diff
@@ -117,18 +130,22 @@ defmodule Realtime.Tenants.Connect do
   def init(%{tenant_id: tenant_id} = state) do
     Logger.metadata(external_id: tenant_id, project: tenant_id)
 
-    with %Tenant{} = tenant <- Tenants.get_tenant_by_external_id(tenant_id),
+    with %Tenant{} = tenant when not is_nil(tenant) <-
```
josevalim, line 133:
```suggestion
    with %Tenant{} = tenant <-
```

`tenant` cannot be nil here, since it has to match a struct.

filipecabaco:
Hi @josevalim and thank you!

We actually have multiple handlers, one per tenant and the same happens for PosgtresReplication modules. Meaning that we need to register one process of Handler and one process of PostgresReplication per tenant on connect.

Assumptions:
* We have multiple tenants and each one of them has their own connection details
* Tenants are identified with a tenant_id

So the flow is as follows:
* First client joins which means that we test the connection and keep the connection against the Tenant Database for future usage in the Realtime.Tenants.Connect
* On success, we start their BroadcastChanges.Handler which is registered with the name `{:via, Registry, {__MODULE__.Registry, tenant_id}}`
* BroadcastChanges.Handler starts PostgresReplication against `realtime.messages` which is registered with the name `{:via, Registry, {__MODULE__.Registry, tenant_id}}`
* First client connection is upgraded to be a WS (uses Phoenix.Channels)
* Second client connects, we already have a DB connection via Connect; a message Handler and a PostgresReplication process associated with this tenant_id so nothing is created
* Second client connection gets upgraded a WS (uses Phoenix.Channels)

On the PostgresReplication / Handler:
* On trigger, `realtime.messages` write to the WAL
* PostgresReplication for tenant_id receives the event and sends it to the Handler of tenant_id
* Handler receives messages and does the various steps to get it ready to be sent (get relation data (relation) => join relation data with payload content and sets a format (insert) => broadcast (commit)
* Broadcast uses PubSub to broadcast to users that are connected

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,205 @@
...
+
+            to_broadcast =
+              %{
+                topic: Map.fetch!(to_broadcast, "topic"),
+                event: Map.fetch!(to_broadcast, "event"),
+                private: Map.fetch!(to_broadcast, "private"),
+                payload: Map.put(payload, "id", id)
+              }
+
+            buffer = Enum.reverse([to_broadcast | buffer])
```
josevalim, line 172: You should accumulate them all and only reverse at the end, when you emit it during the commit.

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,205 @@
...
+        message |> decode_message() |> then(&send(pid, &1))
+        :noreply
+
+      _ ->
+        Logger.error("Unable to find BroadcastChanges for tenant: #{tenant_id}")
+        :shutdown
+    end
+  end
+
+  def call(message, _metadata) when is_keep_alive(message) do
```
josevalim, line 111: I would consider moving this to PostgrexReplication, as I can't see a handler wanting to do anything but returning the keep alive, and this can be more focused on the specific application aspects.

josevalim:
> We actually have multiple handlers, one per tenant and the same happens for PosgtresReplication modules. Meaning that we need to register one process of Handler and one process of PostgresReplication per tenant on connect.

Thank you, that makes complete sense! The existing connection is already responsible for setting up the concurrency and termination. With this in mind, I think you could simplify the current implementation?

For example, you might not need the registry for handlers: instead, a handler starts the replication and pass `self()` as a handler to the replication layer, and the replication layer sends messages to the handler process. This allows you to get rid of the behaviour and of the registry. Testing should still be straight-forward, because you can pass the test process to the replication. The only downside is that you cannot give a result back, but if you move the keep alive handling to the replication, then that should be fine anyway.

You probably don't need a registry for the postgrex replication either? Each connection knows the replication pid and the handler will be directly messaged.

WDYT?

There is even a question if you should have both a handler and a listener and if they could be the same process, but we can discuss this later. :)

lib/realtime/broadcast_changes/handler.ex
filipecabaco: good call will move

filipecabaco:
Makes sense! Will work on removing the registries from both of them and will actually move the broadcast into PostgresReplication.

I did this heavily based on a quick lib I did to learn more about replication ( github.com/filipecabaco/postgres_replication ) and kept that format but for this specific scenario it's just adding more complexity without gains.

Will rework this today / tomorrow. 

Thank you for the feedback 🙏

filipecabaco:
@josevalim did the change 👍 feels way simpler and less error prone. Still needed a Registry due to the fact that the tests were creating multiple entries and ended up clashing.

There might be some really heavy improvements to be done on the Realtime.Tenants.Connect module to simplify this.

josevalim:
> @josevalim did the change 👍 feels way simpler and less error prone. Still needed a Registry due to the fact that the tests were creating multiple entries and ended up clashing.

Fantastic! Final question: how would they clash? Where they named before? If so, why did they have to be named?

filipecabaco:
They were named before yes.

When we do `Realtime.Tenants.Connect.lookup_or_start_connection` which either fetches the pid or starts the connection. For some reason in testing with the same pid it tries to create multiple instances of postgres replication ending up in an error of "replication slot already in use".

I do suspect that it's due to the way `test/integration/rt_channel_test.exs` are written where I'm constantly tearing down connections to have a "fresh start" 🤔

josevalim:
The tenant should store in it its state if it has a replication connection or not, so you should not have dupes, right? Unless you want to access this information externally, from another process, and you don't want to ask the tenant for the PID every time.

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,349 @@
...
+    "all_table_slot"
+  end
+
+  def replication_slot_name(%__MODULE__{replication_slot_name: nil, table: table, schema: schema}) do
+    "#{schema}_#{table}_replication_slot"
+  end
+
+  def replication_slot_name(%__MODULE__{replication_slot_name: replication_slot_name}) do
+    replication_slot_name
+  end
```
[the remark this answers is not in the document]
chasers: ^^ can wait until we support permanent slots

chasers (changes requested):

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,349 @@
...
+  def handle_info(%Decoder.Messages.Commit{}, %{buffer: []} = state) do
+    {:noreply, state}
+  end
+
+  def handle_info(%Decoder.Messages.Commit{}, state) do
+    %{buffer: buffer, tenant_id: tenant_id} = state
+    buffer = Enum.reverse(buffer)
+    tenant = Cache.get_tenant_by_external_id(tenant_id)
+
+    case BatchBroadcast.broadcast(nil, tenant, %{messages: buffer}, true) do
```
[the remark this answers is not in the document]
filipecabaco: no strong opinion honestly. it would be less logic so I would be ok with that.

since this is in our side to control, all inserts into realtime.messages should be quite "sane" and probably won't hit any issues

lib/realtime/broadcast_changes/handler.ex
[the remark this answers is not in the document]
filipecabaco: will use the same slot suffix variable from the existing env vars

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,349 @@
...
+    Logger.info("Create replication slot #{replication_slot_name} using plugin #{output_plugin}")
+
+    query =
+      "CREATE_REPLICATION_SLOT #{replication_slot_name} TEMPORARY LOGICAL #{output_plugin} NOEXPORT_SNAPSHOT"
+
+    {:query, query, %{state | step: :check_publication}}
+  end
+
+  def handle_result(
+        [%Postgrex.Result{}],
```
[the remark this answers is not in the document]
filipecabaco: no, this step i always want to run independent of the result of `CREATE_REPLICATION_SLOT`. 

As long it's not a Postgrex.Error, we run the check

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,349 @@
...
+  def handle_result(
+        [%Postgrex.Result{num_rows: 1}],
+        %__MODULE__{step: :create_publication} = state
+      ) do
+    {:query, "SELECT 1", %{state | step: :start_replication_slot}}
+  end
+
+  @impl true
+  def handle_result(
+        [%Postgrex.Result{}],
```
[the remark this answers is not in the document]
filipecabaco: Same for this one, we're just ensuring it's not a Postgrex.Error and if not, we start the stream

lib/realtime/broadcast_changes/handler.ex
```diff
@@ -0,0 +1,311 @@
...
+  @impl true
+  def handle_info(%Decoder.Messages.Relation{} = msg, state) do
+    %Decoder.Messages.Relation{id: id, namespace: namespace, name: name, columns: columns} = msg
+    %{relations: relations} = state
+    relation = %{name: name, columns: columns, namespace: namespace}
+    relations = Map.put(relations, id, relation)
+    {:noreply, %{state | relations: relations}}
+  end
+
+  def handle_info(%Decoder.Messages.Insert{} = msg, state) do
```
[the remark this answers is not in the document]
filipecabaco: as discussed in the sync we will check at a later stage

PR #1164, opened 2024-10-16
fix: Delete realtime.messages entries older than 72h on connect

filipecabaco:

Starts a clean up task on each node to delete the messages in `realtime.message`. It checks:
* Which tenants need cleanup on a given region
* If the tenant should be cleaned by a given node
* Chunks the groups of tenants to avoid overloading Realtime.Repo and number of connections from a given node
* Starts connections for each tenant
* Runs query to delete messages older than 72h

We need two setups needed for tests:
* Self hosted isn't connected as a cluster nor has a region so we need a default behaviour for "regionless" setups
* For Node mode, we need to emulate the region and emulate the node connected

This impacts Cachex tests which is why we had to change to have max_cases set to 1 to avoid parallel tests from other modules to interfer

Finally, we also need to define if the application starts or not the scheduled tasks. This helps with tests but also helps if we need to stop this processes in production.


lib/realtime/tenants/connect.ex
```diff
@@ -135,7 +135,8 @@ defmodule Realtime.Tenants.Connect do
          ref = Process.monitor(conn),
          [%{settings: settings} | _] <- tenant.extensions,
          migrations = %Migrations{tenant_external_id: tenant.external_id, settings: settings},
-         :ok <- Migrations.run_migrations(migrations) do
+         :ok <- Migrations.run_migrations(migrations),
+         {:ok, _} <- Realtime.Messages.delete_old_messages(conn) do
```
chasers, line 139: What if I have one client connected to Realtime for 365 days?

chasers (changes requested):

lib/realtime/messages.ex
```diff
@@ -0,0 +1,18 @@
...
+  alias Realtime.Api.Message
+
+  @doc """
+  Deletes messages older than 72 hours for a given tenant connection
+  """
+  @spec delete_old_messages(pid()) :: {:ok, any()} | {:error, any()}
+  def delete_old_messages(conn) do
+    limit = NaiveDateTime.utc_now() |> NaiveDateTime.add(-72, :hour)
+    query = from m in Message, where: m.inserted_at <= ^limit
+    Repo.del(conn, query)
```
abc3, line 16: wondering if we want to do a batch delete (a loop with a limit), for example, when needing to delete a large number of records, which can affect the performance of the database

lib/realtime/nodes.ex
```diff
@@ -161,4 +173,21 @@ defmodule Realtime.Nodes do
         host
     end
   end
+
+  @doc """
+  Fetches the tenant regions for a given realtime reagion
+  """
+  @spec region_to_tenant_regions(term()) :: term()
```
abc3, line 180:
```suggestion
  @spec region_to_tenant_regions(String.t()) :: list()
```

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+    timer = Application.get_env(:realtime, :schedule_clean, :timer.hours(5))
+    region = Application.get_env(:realtime, :region)
+    chunks = Application.get_env(:realtime, :chunks, 10)
+    state = %__MODULE__{timer: timer, region: region, chunks: chunks}
+    GenServer.start_link(__MODULE__, state, name: __MODULE__)
+  end
+
+  @impl true
+  def init(%__MODULE__{timer: timer} = state) do
+    Process.send_after(self(), :delete_old_messages, timer)
```
abc3, line 30: Maybe need to try removing it when the process starts and then after a timeout. For example, tables will not be cleared if the node's lifetime is less than the timeout (in case of rebooting or an incident, etc)

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+  import Ecto.Query
+  import Realtime.Helpers, only: [log_error: 2]
+
+  alias Realtime.Api.Tenant
+  alias Realtime.Database
+  alias Realtime.Messages
+  alias Realtime.Nodes
+  alias Realtime.Repo
+
+  defstruct timer: nil, region: nil, chunks: nil
```
abc3, line 18:
```suggestion
  @type t :: %__MODULE__{
    timer: non_neg_integer() | nil,
    region: String.t(),
    chunks: pos_integer() | nil
  }
  
  defstruct timer: nil, region: nil, chunks: nil
```

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+        if regions != nil,
+          do:
+            base
+            |> join(:inner, [t], e in assoc(t, :extensions))
+            |> where([t, e], fragment("? -> 'region' in (?)", e.settings, splice(^regions))),
+          else: base
+
+      query
+      |> Repo.stream()
+      |> Stream.filter(&node_responsible_for_cleanup?(&1, region_nodes))
```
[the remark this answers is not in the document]
filipecabaco: No as we need to fetch the tenant_id to check which node will run it

lib/realtime/messages.ex
filipecabaco: good call, might be worth it to explore that 🤔 IMO we can keep as is in this first iteration and check the impact

lib/realtime/tenants/scheduled_message_cleanup.ex
filipecabaco: this is in a way intended, when we have the node starting this might increase the load on boot which can impact it (it's during boot that we see most load / created processes as expected)

we can find other ways, e.g. start after 10 minutes (aka have a new config for "scheduled_start_after")

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  @impl true
+  def handle_info(:delete_old_messages, state) do
+    %{region: region, chunks: chunks, timer: timer} = state
+    regions = Nodes.region_to_tenant_regions(region)
+    region_nodes = Nodes.region_nodes(region)
+
+    Realtime.Repo.transaction(fn ->
+      base = from(t in Tenant, select: t)
```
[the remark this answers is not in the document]
filipecabaco: yes it will check all in a loop but we do need to keep it like this. for users that do not have a lot of activity they would potentially never be cleaned up right?

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+        index = :erlang.phash2(external_id, length(region_nodes))
+        Enum.at(region_nodes, index) == Node.self()
+    end
+  end
+
+  defp run_cleanup_on_tenants(tenants) do
+    Task.start(fn ->
+      Repo.transaction(fn ->
+        Enum.each(tenants, &run_cleanup_on_tenant/1)
+      end)
```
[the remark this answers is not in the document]
filipecabaco: still debating if we need it. I kind of wanted to have it in a transaction to avoid the Repo.preload to overwhelm the system so this way we do it all under the same connection

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,95 @@
...
+  alias Realtime.Api.Tenant
+  alias Realtime.Database
+  alias Realtime.Messages
+  alias Realtime.Nodes
+  alias Realtime.Repo
+
+  defstruct timer: nil, region: nil, chunks: nil
+
+  def start_link(_args) do
+    timer = Application.get_env(:realtime, :schedule_clean, :timer.hours(5))
```
abc3, line 27: A small hint to avoid running all selects across the cluster at the same time
```suggestion
    timer = Application.get_env(:realtime, :schedule_clean, Enum.random(250..350) * 1000)
```

lib/realtime/tenants/scheduled_message_cleanup.ex
abc3: agreed, "scheduled_start_after" is nice

lib/realtime/tenants/scheduled_message_cleanup.ex
abc3: every new send_after can be invoked with a random timeout

lib/realtime/tenants/scheduled_message_cleanup.ex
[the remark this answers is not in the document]
abc3: yeah, but my main concern is that every node will go through the entire table every time

lib/realtime/tenants/scheduled_message_cleanup.ex
filipecabaco: good one 👍  will adapt slightly

lib/realtime/tenants/scheduled_message_cleanup.ex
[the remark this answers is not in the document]
filipecabaco: to be tackled in future PR by adding a `last_active_at` which will be set on Realtime.Tenants.Connect process shutdown

chasers (approved):

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,119 @@
...
+      _ ->
+        index = :erlang.phash2(external_id, length(region_nodes))
+        Enum.at(region_nodes, index) == Node.self()
+    end
+  end
+
+  defp run_cleanup_on_tenants(tenants) do
+    Task.start(fn ->
+      Repo.transaction(fn ->
+        Enum.each(tenants, &run_cleanup_on_tenant/1)
```
chasers, line 100: Mmmm, not sure.

chasers: I think this could be very bad if we end up keep a lot of transactions open on our database because you're connecting to a bunch of tenant databases while holding open a transaction to our database.

filipecabaco: that's why I had the transaction and I chunk the stream. that way we have 1 Conn to get tenants against RDS, one Conn per chunk to RDS to preload within the transaction and one connection to the tenant to delete it

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,119 @@
...
+  defp node_responsible_for_cleanup?(%Tenant{external_id: external_id}, region_nodes) do
+    case Node.self() do
+      :nonode@nohost ->
+        true
+
+      _ ->
+        index = :erlang.phash2(external_id, length(region_nodes))
+        Enum.at(region_nodes, index) == Node.self()
+    end
+  end
```
chasers, line 95: We have this logic elsewhere for starting a tenant on a region node.

Should be consolidated into a util function somewhere and used where we start tenants and here.

lib/realtime/tenants/scheduled_message_cleanup.ex
chasers: Still could have major consequences, and I don't see the benefit of have a transaction open to our database while you do this work. Unless I'm not seeing something?

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,119 @@
...
+
+  defp run_cleanup_on_tenants(tenants) do
+    Task.start(fn ->
```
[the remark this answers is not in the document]
chasers: Or if you want to, limit the task concurrency that this process is allowed to spin off.

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -0,0 +1,119 @@
...
+    end)
+  end
+
+  defp run_cleanup_on_tenant(tenant) do
+    Logger.metadata(project: tenant.external_id, external_id: tenant.external_id)
+    tenant = Repo.preload(tenant, :extensions)
+    Logger.info("ScheduledMessageCleanup cleaned realtime.messages")
+
+    with {:ok, conn} <-
+           Database.connect(tenant, "realtime_clean_messages", 1),
```
chasers:
```suggestion
           Database.connect(tenant, "realtime_janitor", 1),
```

chasers (changes requested):

lib/realtime/tenants/scheduled_message_cleanup.ex
filipecabaco: agreed, we already have the `Realtime.Nodes` as a spot where we can move this into.

will tackle it later

lib/realtime/tenants/scheduled_message_cleanup.ex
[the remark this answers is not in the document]
filipecabaco: added task supervisor and a env var to configure max children

lib/realtime/tenants/scheduled_message_cleanup.ex
filipecabaco: removed the transaction 👍

chasers (approved):

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -54,17 +62,27 @@ defmodule Realtime.Tenants.ScheduledMessageCleanup do
...
       |> Stream.filter(&node_responsible_for_cleanup?(&1, region_nodes))
       |> Stream.chunk_every(chunks)
-      |> Enum.map(&run_cleanup_on_tenants/1)
+      |> Enum.map(fn chunks ->
+        Task.Supervisor.async(
+          __MODULE__.TaskSupervisor,
+          fn -> run_cleanup_on_tenants(chunks) end
+        )
+      end)
+      |> Task.await_many(task_timeout)
```
chasers, line 88: > It is not recommended to await a long-running task inside an OTP behaviour such as [GenServer](https://hexdocs.pm/elixir/1.13/GenServer.html). Instead, you should match on the message coming from a task inside your [GenServer.handle_info/2](https://hexdocs.pm/elixir/1.13/GenServer.html#c:handle_info/2) callback.

https://hexdocs.pm/elixir/1.13/Task.html#await/2

lib/realtime/tenants/scheduled_message_cleanup.ex
```diff
@@ -54,17 +62,27 @@ defmodule Realtime.Tenants.ScheduledMessageCleanup do
...
+        where: t.notify_private_alpha == true
+      )
       |> where_region(regions)
-      |> Repo.stream()
+      |> Repo.all()
       |> Stream.filter(&node_responsible_for_cleanup?(&1, region_nodes))
       |> Stream.chunk_every(chunks)
-      |> Enum.map(&run_cleanup_on_tenants/1)
+      |> Enum.map(fn chunks ->
+        Task.Supervisor.async(
```
chasers, line 83: > Raises an error if supervisor has reached the maximum number of children.

This GenServer would crash then if you reach max_children. 

Which might be fine, but I would expect you to reach max_children potentially frequently once we're doing this for a lot of dbs. 

Consider using `Task.Supervisor.start_child` to handle the max_children error tuple.

chasers (changes requested):

README.md
```diff
@@ -148,7 +148,10 @@ If you're using the default tenant, the URL is `ws://realtime-dev.localhost:4000
...
+| RUN_SCHEDULED                  | boolean | Do you want to scheduled tasks to run                                                                                                                                                                                                                                                                                           |
+| MAX_CHILDREN_SCHEDULED_CLEANUP | number  | Maximum number of concurrent tasks working on scheduled cleanup                                                                                                                                                                                                                                                                 |
+| SCHEDULED_CLEANUP_TASK_TIMEOUT | number  | Timeout for each async task for scheduled cleanup                                                                                                                                                                                                                                                                               |
```
chasers, line 154: A bit of a nit, but if we have the application_name of `realtime_janitor` we should change these to `RUN_JANITOR` etc cuz people might be searching our repo for "janitor" after seeing this application_name.

lib/realtime/tenants/scheduled_message_cleanup.ex
filipecabaco: went with async_nolink and set a limit at the supervisor level

chasers (approved):

PR #1170, opened 2024-10-18
Fix: detect IPv6 and use appropriate socket for Realtime.Repo

tvogel:

This change allows to use a `DB_HOST` connected via IPv6.

When the `DB_HOST` resolves to an IPv6, connecting to the database fails with `:nxdomain`.

## What is the new behavior?

The connection is established successfully.

## Additional context

Uses logic from `Realtime.Database.detect_ip_version/1` and applies it to the `Realtime.Repo` `socket_options`, too. Unfortunately, I was not able to directly call `Realtime.Database.detect_ip_version/1` from `runtime.exs`. I am an Elixir newbie, so there may be a more elegant way.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1,5 +1,15 @@
 import Config
 
+detect_ip_version = fn host when is_binary(host) ->
+  host = String.to_charlist(host)
+
+  cond do
+    match?({:ok, _}, :inet6_tcp.getaddr(host)) -> {:ok, :inet6}
+    match?({:ok, _}, :inet.gethostbyname(host)) -> {:ok, :inet}
+    true -> {:error, :nxdomain}
+  end
+end
+
 config :logflare_logger_backend,
   url: System.get_env("LOGFLARE_LOGGER_BACKEND_URL", "https://api.logflare.app")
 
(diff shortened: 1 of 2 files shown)


filipecabaco:
this is already automatically by `Realtime.Database.detect_ip_version/1` and we already run it in production with IPv6

is your use case not covered by this?

tvogel:
> this is already automatically by `Realtime.Database.detect_ip_version/1` and we already run it in production with IPv6
> 
> is your use case not covered by this?

Right, that only buys part of IPv6 support as described above: That detection is also needed when setting the socket-options for the Realtime.Repo. As written above, I got an `:nxdomain` in a pure IPv6 environment without these changes.

filipecabaco:
oh got it! what do you think of making an explicit check for the repo? 

e.g.

environment variable determining that the root db will be IPV6 true / false and use that to set things? Less code and easier to understand what is happening

tenants required to be more dynamic but the base database should be more "set in stone"

filipecabaco:
we'll also need to update the mix.exs file to bump up the version 👀

tvogel:
> we'll also need to update the mix.exs file to bump up the version 👀

How do you handle the versions? I have updated it before but obviously, it's a race vs. other PRs. How to proceed?

filipecabaco:
we had a couple of tricky weeks, hopefully after some of the pending PRs are tackled I will grab this and merge it myself. I hope to do it within a week

sorry for the inconvenience 😞 just really intense weeks overall

filipecabaco:
Sorry for the delay on this PR, I've created this one so we can proceed faster

https://github.com/supabase/realtime/pull/1250

filipecabaco:
the PR has been merged, thank you again for bringing this issue up @tvogel 🙏 and big sorry for the time it took to tackle it

PR #1172, opened 2024-10-19
fix: UUID auto generate; Error handling & e2e tests

filipecabaco:

* UUID auto generation on migration
* Better error handling
* Fix e2e tests


test/e2e/tests.ts
```diff
@@ -275,11 +291,16 @@ describe("authorization check", () => {
...
   const table = "broadcast_changes";
-  const id = 1;
+  const id = crypto.randomUUID();
 
   it("authenticated user receives insert broadcast change from a specific topic based on id", async () => {
     let supabase = await createClient(url, token, { realtime });
-    let accessToken = await signInUser(supabase, "test1@test.com", "test_test");
+    let accessToken = await signInUser(
+      supabase,
+      "filipe@supabase.io",
```
J0, line 300: /nit Not sure what the realtime stance is but in Auth we generally exclude references to @supabsae.io

Can consider changing the test email address

J0 (approved):
Looks like we move the auto generation logic from Ecto to Postgres, add additional error handling and a few e2e tests

Nice work.

test/e2e/tests.ts
filipecabaco: we need it for Supabase staging projects 😓

PR #1174, opened 2024-10-21
fix: Adds plug like logic to Connect

filipecabaco:

Add Plug like logic to Connect so we can pipe operations of the Connect operation to ensure we guarantee full operation of all features

### DIFF
--- a/lib/realtime/tenants/connect/piper.ex
+++ b/lib/realtime/tenants/connect/piper.ex
@@ -14,10 +14,8 @@ defmodule Realtime.Tenants.Connect.Piper do
           {:halt, {:error, error}}
 
         _e ->
-          {:halt, {:error, :invalid_pipe_result}}
+          raise ArgumentError, "must return {:ok, _} or {:error, _}"
       end
     end)
-  rescue
-    e -> {:error, e}
   end
 end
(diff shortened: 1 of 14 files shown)


chasers (approved):
Cool!!

lib/realtime/tenants/connect/piper.ex
```diff
@@ -0,0 +1,23 @@
...
+
+        {:error, error} ->
+          {:halt, {:error, error}}
+
+        _e ->
+          {:halt, {:error, :invalid_pipe_result}}
+      end
+    end)
+  rescue
+    e -> {:error, e}
```
wojtekmach, line 21: were you rescuing errors prior? We prefer errors to bubble up and catch them in the super generic handlers at the very top of the app. By catching it here you'd also lose stack trace.

test/realtime/tenants/connect/piper_test.exs
```diff
@@ -0,0 +1,66 @@
+defmodule Realtime.Tenants.Connect.PiperTest do
+  use ExUnit.Case
```
wojtekmach, line 2: If we can make this asynchronous we definitely should, if not, consider adding an explicit `async: false` with a comment explaining why not. We like to do that sort of thing because looking at such code again down the road gives us an opportunity to check if the status quo have changed since and perhaps we can make it async again.

```suggestion
  use ExUnit.Case, async: true
```

wojtekmach (commented):
I left some comments below, let me know what you think!

lib/realtime/tenants/connect/piper.ex
filipecabaco: 👍 will remove and let it bubble up

PR #1186, opened 2024-10-31
fix: Add partitioning to realtime.messages

filipecabaco:

Adds logic to partition the realtime.messages table. The current way works as follows:
* new table is created called realtime.messages_new that is partitioned by `inserted_at`
* migrate existing policies from realtime.messages to realtime.messages_new
* drop existing realtime.messages table
* rename realtime.messages_new to realtime.messages
* Assign the existing ID sequence to the new table
* Provide all required Grants
* On tenant connect, create dynamically 3 partitions: one for yesterday, tomorrow and today

Everything else works as expected and no changes were required.

### DIFF
--- a/lib/realtime/messages.ex
+++ b/lib/realtime/messages.ex
@@ -2,17 +2,39 @@ defmodule Realtime.Messages do
   @moduledoc """
   Handles `realtime.messages` table operations
   """
-  import Ecto.Query
-  alias Realtime.Repo
-  alias Realtime.Api.Message
 
   @doc """
   Deletes messages older than 72 hours for a given tenant connection
   """
-  @spec delete_old_messages(pid()) :: {:ok, any()} | {:error, any()}
+  @spec delete_old_messages(pid()) :: :ok
   def delete_old_messages(conn) do
-    limit = NaiveDateTime.utc_now() |> NaiveDateTime.add(-72, :hour)
-    query = from m in Message, where: m.inserted_at <= ^limit
-    Repo.del(conn, query)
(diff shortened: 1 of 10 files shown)


lib/realtime/messages.ex
```diff
@@ -2,17 +2,39 @@ defmodule Realtime.Messages do
...
+        AND nmsp_child.nspname = 'realtime'
+        """,
+        []
+      )
+
+    rows
+    |> Enum.filter(fn ["messages_" <> date] ->
+      date |> String.replace("_", "-") |> Date.from_iso8601!() |> Date.compare(limit) == :lt
+    end)
+    |> Enum.each(&Postgrex.query!(conn, "DROP TABLE IF EXISTS realtime.#{&1}", []))
```
chasers, line 36: I think this is probably fine but you should confirm that you don't need to call `truncate` first.

lib/realtime/tenants/connect/create_partitions.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Realtime.Tenants.Connect.CreatePartitions do
```
chasers, line 1: Maybe okay for the short term, but I guarantee we'll get folks trying to create messages here when partitions don't exist because the creation of these depend on a client connection.

chasers (approved):

lib/realtime/messages.ex
filipecabaco: added a test that does add data to all partitions. this should test if we need truncate

PR #1191, opened 2024-11-06
fix: Add claims to invalid jwt error

filipecabaco:

Add claims to invalid jwt error


chasers (changes requested):
I don't get it. We already have the invalid claim here and log it.

And they could use the log_level param to log the full claims when they connect.

PR #1193, opened 2024-11-06
fix: Implement private only check on socket connect

filipecabaco:
Implement private only check on socket connect

### DIFF
--- a/lib/realtime/api/tenant.ex
+++ b/lib/realtime/api/tenant.ex
@@ -74,7 +74,8 @@ defmodule Realtime.Api.Tenant do
       :max_channels_per_client,
       :max_joins_per_second,
       :suspend,
-      :notify_private_alpha
+      :notify_private_alpha,
+      :private_only
     ])
     |> validate_required([
       :external_id,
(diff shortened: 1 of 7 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -676,4 +684,13 @@ defmodule RealtimeWeb.RealtimeChannel do
   defp maybe_assign_policies(_, _, _, _, socket) do
     {:ok, assign(socket, policies: nil)}
   end
+
+  defp only_private?(tenant_id, %{assigns: %{check_authorization?: check_authorization?}}) do
```
abc3, line 688:
```suggestion
  @spec only_private?(String.t(), map()) :: :ok | {:error, :private_only}
  defp only_private?(tenant_id, %{assigns: %{check_authorization?: check_authorization?}}) do
```

lib/realtime/api/tenant.ex
```diff
@@ -74,7 +74,8 @@ defmodule Realtime.Api.Tenant do
       :max_channels_per_client,
       :max_joins_per_second,
       :suspend,
-      :notify_private_alpha
+      :notify_private_alpha,
+      :private_only
```
abc3, line 78: double checking if a migration is not needed for this

filipecabaco: migrations was in a different PR and has defaults to false

abc3 (approved):

PR #1198, opened 2024-11-08
fix: Add prefix to version on custom docker build

filipecabaco:

Add prefix to version on custom docker build


filipecabaco:
to be tackled when we formalize hotfix process

PR #1200, opened 2024-11-08
fix: Add last_active_at to Tenant

filipecabaco:

Adds last_active_at to Tenant so we can use it in the Janitor module which will handle tenant cleanup of partitions.

For now we will only clean up partitions of the users that have been online during the last 5 days


lib/realtime/tenants.ex
```diff
@@ -287,11 +292,62 @@ defmodule Realtime.Tenants do
...
+  An active tenant is a tenant that has `last_active_at` higher than the given days.
+
+  Defaults to 5 days.
+  """
+  def stream_active_tenants(days \\ 5) do
+    region = Application.get_env(:realtime, :region)
+    regions = Nodes.region_to_tenant_regions(region)
+    region_nodes = Nodes.region_nodes(region)
+
+    from(t in Tenant,
```
abc3, line 318: how should we handle new customers who have worked for more than five days but have not yet been terminated?

filipecabaco: updated to be in handle_continue

lib/realtime/tenants/janitor.ex
```diff
@@ -96,7 +84,8 @@ defmodule Realtime.Tenants.Janitor do
 
   def handle_info({:DOWN, ref, _, _, :normal}, state) do
```
chasers, line 85: You're handling the failed case but doesn't look like the successful case. 

Seems like your state would grow unbounded with tasks over time. 

The async_nolink docs:

https://hexdocs.pm/elixir/Task.Supervisor.html#async_nolink/3

```elixir
  # The task completed successfully
  def handle_info({ref, answer}, %{ref: ref} = state) do
    # We don't care about the DOWN message now, so let's demonitor and flush it
    Process.demonitor(ref, [:flush])
    # Do something with the result and then return
    {:noreply, %{state | ref: nil}}
  end
  ```

chasers (changes requested):

lib/realtime/tenants/janitor.ex
chasers: Oh wait... I see it now.

chasers: I guess because you don't care about the response from the task you're handling the normal down message.

chasers (approved):

filipecabaco:
https://github.com/supabase/realtime/pull/1201

PR #1205, opened 2024-11-11
fix: Removes sleep from Connection checking

filipecabaco:

On connection initialisation / process finding for DB process, we were putting Connect module to sleep which could have a high impact and create an infinite loop.

This removes that code and if Syn as yet to find the process the cache has most likely not been propagated yet.


abc3 (approved):

filipecabaco:
This one will need to be properly tested because I think it might be impacting negatively

filipecabaco:
https://github.com/supabase/realtime/pull/1209

PR #1222, opened 2024-11-14
allow default self-hosted tenant name to be configurable via ENV

sbeardsley:

self-hosted seed update to allow tenant_name to be configurable via environment variable. 

The current behavior requires the container name to be "realtime-dev". If it isn't then it will not match the default tenant. In cases where the container name will be different (i.e. kubernetes deployments) it is impossible to use default tenant. 

## What is the new behavior?

The new behavior allows to configure the default tenant_name via an environment variable named TENANT_NAME with fallback behavior to the original "realtime-dev".

## Additional context

This was an issue for the supabase-kubernetes project. I have submitted a PR there as well to overwrite the seeds.exs file in https://github.com/supabase-community/supabase-kubernetes/pull/91

Ultimately it would be better to have this be configuration change upstream in this repo.

### DIFF
--- a/priv/repo/seeds.exs
+++ b/priv/repo/seeds.exs
@@ -2,7 +2,7 @@ require Logger
 alias Realtime.{Api.Tenant, Repo}
 import Ecto.Adapters.SQL, only: [query: 3]
 
-tenant_name = "realtime-dev"
+tenant_name = System.get_env("TENANT_NAME", "realtime-dev")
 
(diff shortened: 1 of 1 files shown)


filipecabaco:
@sbeardsley I opened a PR with some minor changes https://github.com/supabase/realtime/pull/1249 we can go from there. can you also give some insight on other useful elements we should change?

PR #1226, opened 2024-11-15
fix: Improve errors logged out and erpc error handling

filipecabaco:

Improve errors logged out and erpc error handling

### DIFF
--- a/lib/realtime/rpc.ex
+++ b/lib/realtime/rpc.ex
@@ -36,38 +36,29 @@ defmodule Realtime.Rpc do
   def enhanced_call(node, mod, func, args \\ [], opts \\ []) do
     timeout = Keyword.get(opts, :timeout, Application.get_env(:realtime, :rpc_timeout))
 
-    try do
-      :timer.tc(fn -> :erpc.call(node, mod, func, args, timeout) end)
-    catch
-      kind, reason ->
-        log_error("ErrorOnRpcCall", %{target: node, mod: mod, func: func, error: {kind, reason}})
-        {:error, "RPC call error"}
-    else
-      {_, {:EXIT, reason}} ->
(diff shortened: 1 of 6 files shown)


lib/realtime/rpc.ex
```diff
@@ -36,38 +36,29 @@ defmodule Realtime.Rpc do
...
-        {:error, "RPC call error"}
-    else
-      {_, {:EXIT, reason}} ->
-        log_error("ErrorOnRpcCall", %{target: node, mod: mod, func: func, error: {:EXIT, reason}},
-          mod: mod,
-          func: func,
-          target: node
-        )
+    with {latency, {status, _} = response} <-
+           :timer.tc(fn -> :erpc.call(node, mod, func, args, timeout) end) do
```
abc3, line 40: is it safe that `erpc` can raise an error?

lib/realtime/rpc.ex
```diff
@@ -36,38 +36,29 @@ defmodule Realtime.Rpc do
   def enhanced_call(node, mod, func, args \\ [], opts \\ []) do
```
abc3, line 36: wondering if it is clearer 
```suggestion
  def enhanced_call(node, mod, func, args \\ [], logger_ metadata \\ []) do
```

abc3: actually, I don't see where these opts are applied to the logger, or am I missing something

filipecabaco: Screenshot 2024-11-18 at 13 08 35
used in on the rescue

lib/realtime/rpc.ex
filipecabaco: I would expect the catch to handle those scenarios but still kind of leaning towards just using `:rpc`... wanted to avoid bigger code changes

abc3 (approved):

PR #1227, opened 2024-11-15
fix: Move migrations to non blocking callback

filipecabaco:

Moves migrations to a non blocking callback meaning that multiple migrations will be able to be started. We're also handling errors better when the migrations are already running preventing multiple migrations from starting

### DIFF
--- a/lib/realtime/tenants/migrations.ex
+++ b/lib/realtime/tenants/migrations.ex
@@ -133,7 +133,8 @@ defmodule Realtime.Tenants.Migrations do
     spec = {__MODULE__, attrs}
 
     case DynamicSupervisor.start_child(supervisor, spec) do
-      :ignore -> :ok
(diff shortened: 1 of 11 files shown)


lib/realtime/tenants/migrations.ex
```diff
@@ -195,6 +199,18 @@ defmodule Realtime.Tenants.Migrations do
...
   end
 
+  def terminate(:normal, _) do
+    Logger.info("Migrations ran successfully")
+
+    :ok
+  end
+
+  def terminate({:shutdown, error}, _) do
+    IO.inspect("here")
```
CarelFdeWaal, line 209: Forgot to remove?

filipecabaco: 🤦 yes... thank you

PR #1228, opened 2024-11-18
fix: Change when we track and create partitions

filipecabaco:

* we will use the existing syn scope of the Connect module to check which users still have database connections active
* remove ets as this is no longer required
* Janitor will also tackle creating partitions

### DIFF
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -298,32 +298,6 @@ defmodule Realtime.Tenants do
     |> tap(fn _ -> Cache.invalidate_tenant_cache(tenant_id) end)
   end
 
-  @doc """
-  Tracks the active tenant by external_id and stores the time when it was tracked in the ETS table named `:active_tenants`.
-  """
-  @spec track_active_tenant(String.t()) :: :ok
-  def track_active_tenant(external_id) do
-    :ets.insert(:active_tenants, {external_id, NaiveDateTime.utc_now()})
-    :ok
-  end
-
-  @doc """
-  Lists all active tenants from the ETS table named `:active_tenants`.
-  """
-  @spec track_active_tenant(String.t()) :: list({String.t(), NaiveDateTime.t()})
-  def list_active_tenants() do
-    :ets.tab2list(:active_tenants)
-  end
-
-  @doc """
-  Untracks the active tenant by external_id from the ETS table named `:active_tenants`.
-  """
-  @spec untrack_active_tenant(String.t()) :: :ok
-  def untrack_active_tenant(external_id) do
-    :ets.delete(:active_tenants, external_id)
-    :ok
-  end
-
   defp broadcast_operation_event(action, external_id) do
     Phoenix.PubSub.broadcast!(
       Realtime.PubSub,
--- a/lib/realtime/tenants/janitor.ex
+++ b/lib/realtime/tenants/janitor.ex
@@ -11,8 +11,8 @@ defmodule Realtime.Tenants.Janitor do
   alias Realtime.Api.Tenant
   alias Realtime.Database
   alias Realtime.Messages
-  alias Realtime.Repo
   alias Realtime.Tenants
+  alias Realtime.Tenants.Migrations
 
   @type t :: %__MODULE__{
           timer: pos_integer() | nil,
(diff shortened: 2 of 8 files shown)


test/realtime/tenants/janitor_test.exs
```diff
@@ -53,7 +54,7 @@ defmodule Realtime.Tenants.JanitorTest do
     %{tenants: tenants}
   end
 
-  test "cleans messages older than 72 hours from tenants that were active and untracks the user",
+  test "cleans messages older than 72 hours and creats partitions from tenants that were active and untracks the user",
```
CarelFdeWaal, line 57: typo: creates

lib/realtime/tenants.ex
```diff
@@ -310,7 +310,7 @@ defmodule Realtime.Tenants do
   @doc """
   Lists all active tenants from the ETS table named `:active_tenants`.
   """
-  @spec track_active_tenant(String.t()) :: list({String.t(), NaiveDateTime.t()})
+  @spec list_active_tenants() :: list({String.t(), NaiveDateTime.t()})
   def list_active_tenants() do
     :ets.tab2list(:active_tenants)
   end
```
abc3, line 316: could it be an issue with having a large number of tenants?

lib/realtime/tenants/janitor.ex
```diff
@@ -118,7 +124,8 @@ defmodule Realtime.Tenants.Janitor do
 
     with %Tenant{} = tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_external_id),
          {:ok, conn} <- Database.connect(tenant, "realtime_janitor", 1),
-         :ok <- Messages.delete_old_messages(conn) do
+         :ok <- Messages.delete_old_messages(conn),
+         :ok <- Migrations.create_partitions(conn) do
       Logger.info("Janitor finished")
       GenServer.stop(conn)
```
abc3, line 130: how can you prevent `conn` from becoming a zombie?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -94,6 +94,8 @@ defmodule RealtimeWeb.RealtimeChannel do
         db_conn: db_conn
       }
 
+      Tenants.track_active_tenant(tenant_id)
```
abc3, line 97: where is the implementation of `track_active_tenant` now? You've changed it to `list_active_tenants` in the code above

lib/realtime/tenants/janitor.ex
```diff
@@ -52,6 +53,11 @@ defmodule Realtime.Tenants.Janitor do
   def init(%__MODULE__{start_after: start_after} = state) do
     timer = timer(state) + start_after
     Process.send_after(self(), :delete_old_messages, timer)
+
+    if :ets.whereis(:active_tenants) == :undefined do
+      :ets.new(:active_tenants, [:named_table, :set, :public, write_concurrency: true])
+    end
```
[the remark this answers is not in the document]
filipecabaco: shouldn't we avoid that? I thought that the best practices was to create it on a parent process.

will revert it to be started in application.ex only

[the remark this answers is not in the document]
abc3: I mean that it is already created in application.ex, and now this table tries to be created inside the janitor

lib/realtime/tenants.ex
filipecabaco: we're seeing 3k max across the cluster so per node, worse case scenario, we would have 3k entries at this point in time which doesn't seem that big for a tuple of `{String.t(), DateTime.t()}`

also we only run this command once every ~4 hours so it should not be a big impact but we can look into have some sorte of offset to batch it 🤔 

but would rather do it if required. Worst node we saw had 300 entries

lib/realtime/tenants/janitor.ex
```diff
@@ -118,7 +124,8 @@ defmodule Realtime.Tenants.Janitor do
 
     with %Tenant{} = tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_external_id),
          {:ok, conn} <- Database.connect(tenant, "realtime_janitor", 1),
-         :ok <- Messages.delete_old_messages(conn) do
+         :ok <- Messages.delete_old_messages(conn),
+         :ok <- Migrations.create_partitions(conn) do
```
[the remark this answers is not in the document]
filipecabaco: good call, we would need to rename it to something else

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: changed the spec only as it was not properly set 🤦 

it's still in `Realtime.Tenants`

lib/realtime/tenants/janitor.ex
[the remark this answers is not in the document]
filipecabaco: it was created in application.ex for the testing scenario as we would need it for that. I moved all back to `application.ex` for now

lib/realtime/tenants/janitor.ex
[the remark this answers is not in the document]
filipecabaco: `perform_mantaince_task/1`

chasers (approved):

PR #1233, opened 2024-11-21
fix: Metrics cleanup based on syn module

filipecabaco:

Adds a new worker that will cleanup metrics based on our syn information to reduce the amount of data pulled by Prometheus.


config/runtime.exs
```diff
@@ -22,6 +22,8 @@ config :realtime,
     System.get_env("TENANT_MAX_EVENTS_PER_SECOND", "100") |> String.to_integer(),
   tenant_max_joins_per_second:
     System.get_env("TENANT_MAX_JOINS_PER_SECOND", "100") |> String.to_integer(),
+  metrics_cleaner_scheduke_timer_in_ms:
```
abc3, line 25:
```suggestion
  metrics_cleaner_schedule_timer_in_ms:
```

lib/realtime/metrics_cleaner.ex
```diff
@@ -0,0 +1,60 @@
...
+    |> Enum.reject(fn tenant_id -> tenant_id in tenant_ids end)
+    |> Enum.each(fn tenant_id -> delete_metric(tenant_id) end)
+  end
+
+  @doc """
+  Deletes all metrics that contain the given tenant.
+  """
+  @spec delete_metric(String.t()) :: :ok
+  def delete_metric(tenant) do
+    :ets.select_delete(@metrics_table, [{{{:_, %{tenant: tenant}}, :_}, [], [true]}])
```
abc3, line 57: is it always only the tenant value in the metadata?

filipecabaco: I pulled a sample from production and yes it's always tenant

lib/realtime/metrics_cleaner.ex
```diff
@@ -0,0 +1,60 @@
...
+    |> Enum.uniq()
+    |> Enum.reject(fn tenant_id -> tenant_id in tenant_ids end)
+    |> Enum.each(fn tenant_id -> delete_metric(tenant_id) end)
+  end
+
+  @doc """
+  Deletes all metrics that contain the given tenant.
+  """
+  @spec delete_metric(String.t()) :: :ok
+  def delete_metric(tenant) do
```
abc3, line 56: :ets.select_delete returns a count of deleted items, which can be useful

filipecabaco: yes but I was ignoring it for now as we don't have a good use case for it now

abc3 (approved):

lib/realtime/metrics_cleaner.ex
```diff
@@ -0,0 +1,60 @@
...
+  @metrics_table Realtime.PromEx.Metrics
+  @filter_spec [{{{:_, %{tenant: :"$1"}}, :_}, [], [:"$1"]}]
+  @tenant_id_spec [{{:"$1", :_, :_, :_, :_, :_}, [], [:"$1"]}]
+  defp loop_and_cleanup_metrics_table do
+    tenant_ids = :ets.select(@table_name, @tenant_id_spec)
+
+    :ets.select(@metrics_table, @filter_spec)
+    |> Enum.uniq()
+    |> Enum.reject(fn tenant_id -> tenant_id in tenant_ids end)
+    |> Enum.each(fn tenant_id -> delete_metric(tenant_id) end)
```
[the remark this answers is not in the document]
filipecabaco: I think I prefer the simpler readability since it will be handling a low amount of data... at least for now

PR #1241, opened 2024-12-03
fix: set proper size for realtime_rls pool

filipecabaco:

Connection pool for realtime_rls was not set properly and was using a higher value than it should leading to Realtime attempting to re-use replication slots creating issues upon connecting to listen to changes
    
Now we will set `realtime_rls` and `realtime_broadcast_changes` to hardcoded `1` and `realtime_migrations` to hardcoded `2`

### DIFF
--- a/lib/realtime/database.ex
+++ b/lib/realtime/database.ex
@@ -76,20 +76,51 @@ defmodule Realtime.Database do
   @doc """
   Returns the pool size for a given application name. Override pool size if provided.
 
+  `realtime_rls` and `realtime_broadcast_changes` will be handled as a special scenario as it will need to be hardcoded as 1 otherwise replication slots will be tried to be reused leading to errors
+  `realtime_migrations` will be handled as a special scenario as it requires 2 connections.
   ## Examples
 
-      iex> Realtime.Database.pool_size_by_application_name("realtime_rls", %{}, 1)
+      iex> Realtime.Database.pool_size_by_application_name("realtime_connect", %{}, 1)
       1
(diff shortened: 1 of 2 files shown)


pcnc (approved):

lib/realtime/database.ex
```diff
@@ -99,16 +130,18 @@ defmodule Realtime.Database do
...
         "realtime_connect" -> settings["db_pool"]
         "realtime_health_check" -> 1
-        "realtime_broadcast_changes" -> 1
-        "realtime_migrations" -> 2
         "realtime_janitor" -> 1
         _ -> 1
       end
 
-    if override_pool, do: override_pool, else: pool
+    case application_name do
```
chasers, line 139: Two case statements for application_name?

lib/realtime/database.ex
```diff
@@ -99,16 +130,18 @@ defmodule Realtime.Database do
...
+      "realtime_rls" -> 1
+      "realtime_broadcast_changes" -> 1
+      "realtime_migrations" -> 2
+      _ -> if override_pool, do: override_pool, else: pool || 1
```
chasers, line 143: And I'd pull this override out into it's own line.

lib/realtime/database.ex
filipecabaco: will change 👍

lib/realtime/database.ex
filipecabaco: will change 👍

PR #1246, opened 2024-12-10
fix: broadcast disconnect to socket processes

filipecabaco:

broadcast disconnect to socket processes

### DIFF
--- a/lib/realtime/api.ex
+++ b/lib/realtime/api.ex
@@ -14,6 +14,8 @@ defmodule Realtime.Api do
   alias Realtime.GenCounter
   alias Realtime.Tenants
 
+  alias RealtimeWeb.UserSocket
+
   @doc """
   Returns the list of tenants.
 
@@ -124,7 +126,9 @@ defmodule Realtime.Api do
          data: %{external_id: external_id}
        })
        when is_map_key(changes, :jwt_jwks) or is_map_key(changes, :jwt_secret) do
-    Phoenix.PubSub.broadcast!(Realtime.PubSub, "user_socket:" <> external_id, "disconnect")
+    external_id
+    |> UserSocket.subscribers_id()
+    |> RealtimeWeb.Endpoint.broadcast("disconnect", %{})
   end
 
   defp maybe_trigger_disconnect(_), do: nil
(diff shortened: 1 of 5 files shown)


lib/realtime/api.ex
```diff
@@ -124,7 +124,7 @@ defmodule Realtime.Api do
          data: %{external_id: external_id}
        })
        when is_map_key(changes, :jwt_jwks) or is_map_key(changes, :jwt_secret) do
-    Phoenix.PubSub.broadcast!(Realtime.PubSub, "realtime:operations:" <> external_id, :disconnect)
+    Phoenix.PubSub.broadcast!(Realtime.PubSub, "user_socket:" <> external_id, "disconnect")
```
chasers, line 127: This works?

chasers (changes requested):

lib/realtime/api.ex
filipecabaco: this has changed to something different but I can't test the socket closing but I do get a call on the terminate callback of RealtimeChannel... 

still the websocket process is still alive in the integrations tests so I'm not fully sure on how to properly tackle it

also we only receive the terminate meaning we can't inform the user that the channeil is going to close due to server request

Ppushed latest to this PR

test/integration/rt_channel_test.exs
```diff
@@ -885,19 +885,11 @@ defmodule Realtime.Integration.RtChannelTest do
...
+
       tenant = Tenants.get_tenant_by_external_id(@external_id)
       Realtime.Api.update_tenant(tenant, %{jwt_jwks: %{keys: ["potato"]}})
-
-      assert_receive %Phoenix.Socket.Message{
-                       topic: ^realtime_topic,
-                       event: "system",
-                       payload: %{
-                         "extension" => "system",
-                         "message" => "Server requested disconnect",
```
chasers, line 896: This seemed like a good message, because we can have them console log something to see why they were disconnected.

If we just disconnect they will think it's an error on our end.

chasers (changes requested):

test/integration/rt_channel_test.exs
filipecabaco: Yes but with this approach of killing the socket we can't send it as the socket will close before we can send the message from the RT channel 😞 

I'm thinking of keeping implementation as it was and then from RealtimeChannel emit the kill socket message... more steps but we can get some of this "niceties"

w3b6x9 (changes requested):
strongly advise against sending disconnect message to socket process from channel processes that are monitoring said socket process.

filipecabaco:
to be tackled in another PR

PR #1253, opened 2024-12-20
fix: listen on notifications for error handling

filipecabaco:

* Adds Realtime.Tenants.Listen to receive error messages from functions so we can better inform our users
* Changes send function to not handle any partition creation
* Changes send function to pg_notify to `realtime:system` so we can capture and handle in Realtime

### DIFF
--- a/lib/realtime/tenants/migrations.ex
+++ b/lib/realtime/tenants/migrations.ex
@@ -68,7 +68,8 @@ defmodule Realtime.Tenants.Migrations do
     MessagesUsingUuid,
     FixSendFunction,
     RecreateEntityIndexUsingBtree,
-    FixSendFunctionPartitionCreation
+    FixSendFunctionPartitionCreation,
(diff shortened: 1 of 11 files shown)


lib/realtime/tenants/migrations.ex
```diff
@@ -252,9 +254,9 @@ defmodule Realtime.Tenants.Migrations do
     Logger.info("Creating partitions for realtime.messages")
     today = Date.utc_today()
     yesterday = Date.add(today, -1)
-    tomorrow = Date.add(today, 1)
+    future = Date.add(today, 10)
```
chasers, line 257:
```suggestion
    future = Date.add(today, 3)
```

I do think we should keep this at 3 for now and setup a project on our end to continuously ping it so we can surface edge cases around this logic faster.

chasers (changes requested):

chasers (approved):

PR #1255, opened 2024-12-24
fix: realtime.send sets the proper value for realtime.topic

filipecabaco:

`realtime.send` sets the proper value for realtime.topic so RLS policies are applied properly.

### DIFF
--- a/lib/realtime/tenants/repo/migrations/20241224161212_realtime_send_sets_config.ex
+++ b/lib/realtime/tenants/repo/migrations/20241224161212_realtime_send_sets_config.ex
@@ -10,7 +10,7 @@ defmodule Realtime.Tenants.Migrations.RealtimeSendSetsConfig do
     BEGIN
       BEGIN
         -- Set the topic configuration
-        SET SESSION realtime.topic TO topic;
+        SET LOCAL realtime.topic TO topic;
 
         -- Attempt to insert the message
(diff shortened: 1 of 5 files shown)


chasers (approved):

lib/realtime/tenants/repo/migrations/20241224161212_realtime_send_sets_config.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  # We missed the schema prefix of `realtime.` in the create table partition statement
+  def change do
+    execute("""
+    CREATE OR REPLACE FUNCTION realtime.send(payload jsonb, event text, topic text, private boolean DEFAULT true ) RETURNS void
+    AS $$
+    BEGIN
+      BEGIN
+        -- Set the topic configuration
+        SET SESSION realtime.topic TO topic;
```
chasers, line 13: Actually we do not want to set this for the whole session, only for the transaction

```
set_config('realtime.topic', 'topic', true)
```

chasers (changes requested):

lib/realtime/tenants/repo/migrations/20241224161212_realtime_send_sets_config.ex
chasers: Or `SET LOCAL`

chasers (approved):

PR #1259, opened 2025-01-07
fix: Async postgres changes process startup

filipecabaco:

Async postgres changes process startup


chasers (approved):
I think this will work!

PR #1266, opened 2025-01-14
fix: Improve test around Connect resilience

filipecabaco:

Improve test around Connect resilience


lib/realtime/tenants/replication_connection.ex
```diff
@@ -118,9 +118,9 @@ defmodule Realtime.Tenants.ReplicationConnection do
         socket_options: [ip_version],
         backoff_type: :stop,
         sync_connect: true,
-        parameters: [
-          application_name: connection_opts.application_name
```
chasers, line 122: This doesn't work?

filipecabaco: from what I understood no, it's better to set after connection

filipecabaco:
Hello @supabase/dashbit 👋 

We're hitting some issues with Postgrex + DBConnection where it seems connections end up not being properly linked to a "parent" pid ending up in what seems to be an accumulation of processes retrying connections.

To improve that I'm trying to use `backoff: :stop` and `max_restarts: 0` to try and ensure that on failure DBConnection just dies, taking with it Postgrex and letting the parent pid (`Realtime.Tenants.Connect`) are the ones handling potential reconnects.

We're also seeing some weird behaviour with ReplicationConnection where after getting killed it seems to retain the replication slot for longer, bringing flakiness to tests

lib/realtime/tenants/replication_connection.ex
chasers: Unless it errors and never sets the application_name.  

We can see the application_name in the connection logs in Postgres, and that is a lot of times helpful to know if some application is reconnecting a lot.

filipecabaco: I will need to recheck to see if this works but I don't think it does 🤔

chasers: You can check the Postgres logs and see it's working:

`connection authorized: user=supabase_admin database=postgres application_name=realtime_rls SSL enabled (protocol=TLSv1.3, cipher=TLS_AES_256_GCM_SHA384, bits=256)`

chasers: At least with our other connections...

josevalim:
@filipecabaco my suggestion would be isolate and reproduce under which scenario they do not end-up "linking". For example:

```elixir
spawn(fn ->
  Postgrex.start_link(name: FooBar)
  exit(:shutdown)
end)

Process.monitor(FooBar)

receive do
  {:DOWN, ^ref, _, _, reason} -> IO.inspect {:yes, reason}
after
  2000 -> IO.inspect :no
end
```

However beware that OTP processes do not necessarily crash on links, unless the link is broken by the parent process. The above assumes that's the case, that the linking is done via parent-child, but if it is done via `Process.link`, you would need to change above and potentially consider other approaches.

> We're also seeing some weird behaviour with ReplicationConnection where after getting killed it seems to retain the replication slot for longer, bringing flakiness to tests

Is this a consequence of the process getting killed (and therefore abruptly terminating the socket)? Should we add a `terminate` callback to ReplicationConnection that would allow you to immediately close any slot before closing down the socket?

filipecabaco:
Hi @josevalim thank you for checking and will follow your advice!

And that is a good call... The issue with the replication slot previously is that when we kill the DB pid for that connection DBConnection reconnects but since we would be doing that under the terminate callback it might differ so will check it 👍 

Big thank you 🙏

lib/realtime/tenants/replication_connection.ex
filipecabaco: will add back and remove this change

chasers (approved):

PR #1285, opened 2025-01-28
fix: handle channel shutdown; fix other exist reasons

filipecabaco:

handle channel shutdown; fix other exist reasons


chasers (approved):

lib/realtime/tenants/replication_connection.ex
```diff
@@ -339,7 +339,7 @@ defmodule Realtime.Tenants.ReplicationConnection do
 
   @impl true
   def handle_disconnect(state) do
-    Logger.warning("Disconnecting broadcast changes handler: #{inspect(state, pretty: true)}")
+    Logger.warning("Disconnecting broadcast changes handler in the step : #{state.step}")
```
abc3, line 342: maybe inspect is missed

filipecabaco: it's an atom but yeah... better do it just in case 😅

PR #1289, opened 2025-01-30
fix: reject connections if tenant db has low connections available

filipecabaco:

Reject connections if tenant db has low connections available

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -49,7 +49,7 @@ defmodule Realtime.Tenants.Connect do
   @spec lookup_or_start_connection(binary(), keyword()) ::
           {:ok, pid()} | {:error, term()}
   def lookup_or_start_connection(tenant_id, opts \\ []) do
-    case get_status(tenant_id) |> IO.inspect() do
+    case get_status(tenant_id) do
       {:ok, conn} ->
         {:ok, conn}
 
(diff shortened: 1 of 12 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -49,7 +49,7 @@ defmodule Realtime.Tenants.Connect do
   @spec lookup_or_start_connection(binary(), keyword()) ::
           {:ok, pid()} | {:error, term()}
   def lookup_or_start_connection(tenant_id, opts \\ []) do
-    case get_status(tenant_id) do
+    case get_status(tenant_id) |> IO.inspect() do
```
abc3, line 52:
```suggestion
    case get_status(tenant_id) |> do
```

lib/realtime/tenants/connect.ex
```diff
@@ -165,12 +171,14 @@ defmodule Realtime.Tenants.Connect do
   def init(%{tenant_id: tenant_id} = state) do
     Logger.metadata(external_id: tenant_id, project: tenant_id)
 
-    with {:ok, acc} <- Piper.run(@pipes, state) do
+    with {:ok, acc} <- Piper.run(@pipes, state) |> IO.inspect() do
```
abc3, line 174:
```suggestion
    with {:ok, acc} <- Piper.run(@pipes, state) do
```

filipecabaco: 🤦

lib/realtime/tenants/connect.ex
filipecabaco: 🤦

chasers (approved):

PR #1295, opened 2025-02-11
fix: refactor presence handler and fix tracking reconnect

filipecabaco:

refactor presence handler and fix tracking reconnect


chasers (approved):

josevalim:
I am on this!

@filipecabaco btw, can you please add @steffende and @jonatanklosko to the Dashbit team? Thank you :)

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -17,23 +18,28 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
+  @spec call(map() | :sync_presence, Phoenix.Socket.t()) ::
           {:noreply, Phoenix.Socket.t()} | {:reply, :error | :ok, Phoenix.Socket.t()}
-  def call(
-        %{"event" => event} = payload,
-        %{assigns: %{is_new_api: true, presence_key: _, tenant_topic: _}} = socket
-      ) do
-    socket = count(socket)
-    {result, socket} = handle_presence_event(event, payload, socket)
+  def call(%{"event" => event} = payload, socket) do
+    event = String.downcase(event)
```
josevalim, line 24: For performance, you may do the following if applicable:

```suggestion
    event = String.downcase(event, :ascii)
```

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -332,48 +311,39 @@ defmodule RealtimeWeb.RealtimeChannel do
     {:stop, :shutdown, socket}
   end
 
+  def handle_info(:sync_presence = msg, socket) do
+    PresenceHandler.call(msg, socket)
```
josevalim, line 315: Nitpick: rename the function from `call` to something else, such as "handle" or "run". The reason is because `call` in Elixir is typically used for GenServers and synchronous communication, so I was originally thinking that this would cross a process (which would be a no-go as that would cause the whole socket to be copied :D).

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -17,23 +18,28 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
   alias RealtimeWeb.Presence
   alias RealtimeWeb.RealtimeChannel.Logging
 
-  @spec call(map(), Phoenix.Socket.t()) ::
+  @spec call(map() | :sync_presence, Phoenix.Socket.t()) ::
```
josevalim, line 21: Perhaps have those as two different functions, `call(map())` and then `sync(socket)`?

josevalim (commented):
LGTM! I have added some comments, but they are mainly nitpicks related to the refactoring :)

PR #1303, opened 2025-02-20
fix: Reduce janitor scope to only local tenants

filipecabaco:

 Reduce janitor scope to only local tenants

### DIFF
--- a/lib/realtime/tenants/janitor.ex
+++ b/lib/realtime/tenants/janitor.ex
@@ -65,7 +65,7 @@ defmodule Realtime.Tenants.Janitor do
     Logger.info("Janitor started")
     %{chunks: chunks, tasks: tasks} = state
     all_tenants = :ets.select(@table_name, [{{:"$1"}, [], [:"$1"]}])
-    connected_tenants = :ets.select(@syn_table, [{{:"$1", :_, :_, :_, :_, :_}, [], [:"$1"]}])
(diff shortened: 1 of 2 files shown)


lib/realtime/tenants/janitor.ex
```diff
@@ -65,7 +65,7 @@ defmodule Realtime.Tenants.Janitor do
     Logger.info("Janitor started")
     %{chunks: chunks, tasks: tasks} = state
     all_tenants = :ets.select(@table_name, [{{:"$1"}, [], [:"$1"]}])
-    connected_tenants = :ets.select(@syn_table, [{{:"$1", :_, :_, :_, :_, :_}, [], [:"$1"]}])
+    connected_tenants = :ets.select(@syn_table, [{{:"$1", :_, :_, :_, :_, Node.self()}, [], [:"$1"]}])
```
hauleth, line 68: Sweet match specs. Technically it probably should be `{:const, Node.self()}` but I think that it will not cause any issues there.

hauleth (approved):

lib/realtime/tenants/janitor.ex
filipecabaco: will do the change so it's more correct

filipecabaco: Screenshot 2025-02-20 at 11 02 38
yeah this is better and more correct :D

PR #1304, opened 2025-02-20
fix: suspend before delete to trigger killing all related processes

filipecabaco:

suspend before delete to trigger killing all related processes

### DIFF
--- a/lib/realtime_web/controllers/tenant_controller.ex
+++ b/lib/realtime_web/controllers/tenant_controller.ex
@@ -1,6 +1,7 @@
 defmodule RealtimeWeb.TenantController do
   use RealtimeWeb, :controller
   use OpenApiSpex.ControllerSpecs
+
   require Logger
   import Realtime.Logs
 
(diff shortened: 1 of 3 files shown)


lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -10,22 +11,16 @@ defmodule RealtimeWeb.TenantController do
...
-
-  alias RealtimeWeb.Endpoint
+  alias RealtimeWeb.OpenApiSchemas.EmptyResponse
+  alias RealtimeWeb.OpenApiSchemas.ErrorResponse
+  alias RealtimeWeb.OpenApiSchemas.NotFoundResponse
+  alias RealtimeWeb.OpenApiSchemas.TenantHealthResponse
+  alias RealtimeWeb.OpenApiSchemas.TenantParams
+  alias RealtimeWeb.OpenApiSchemas.TenantResponse
+  alias RealtimeWeb.OpenApiSchemas.TenantResponseList
+  alias RealtimeWeb.OpenApiSchemas.UnauthorizedResponse
```
hauleth, line 21: Love it. I prefer explicit `alias`es over grouping them as it makes it easier to reorder or remove them.

hauleth (approved):

lib/realtime_web/controllers/tenant_controller.ex
filipecabaco: same! also quicker to ctrl+f for usage

hauleth: True, also, I suggest checking out Lexical as a LSP for Elixir. It is awesome.

PR #1311, opened 2025-02-24
fix: handle suspend properly; handle timeout rpc error

filipecabaco:

handle suspend properly; handle timeout rpc error


chasers (approved):
Approved, but why is every socket getting suspend messages for every tenant?

Shouldn't that be targeted when we send it?

filipecabaco:
It was (the message itself has it) but on the Connect module we were not using that information to actually kill the Connect module.

PR #1333, opened 2025-03-27
fix: temporarily disable log_level setting from params

filipecabaco:

* temporarily disable log_level setting from params
* track long_message_queue


lib/realtime/monitoring/erl_sys_mon.ex
```diff
@@ -11,7 +11,8 @@ defmodule Realtime.ErlSysMon do
     :busy_dist_port,
     :busy_port,
     {:long_gc, 250},
-    {:long_schedule, 100}
+    {:long_schedule, 100},
+    {:long_message_queue, true}
```
chasers, line 15: What is "long" here? This could log way too many times.

In Logfare we do: `{:long_message_queue, {0, 1_000}}`

chasers (changes requested):

chasers:
Also should change these to `false`

https://github.com/supabase/realtime/blob/f19902242338ef065c60991dc2bdc7b4b168189c/lib/realtime_web.ex#L87

chasers (approved):

PR #1354, opened 2025-04-29
fix: connect backoff mechanism

filipecabaco:

Simple mechanism to prevent to many Connect processes from trying to startup in the same node.


lib/realtime/tenants/connect/backoff.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  defp start_connects_per_second_counter(tenant_id, limit) do
+    id = Tenants.connection_attempts_per_second_key(tenant_id)
+    GenCounter.new(id)
+
+    res =
+      RateCounter.new(id,
+        idle_shutdown: :infinity,
+        telemetry: %{
+          event_name: [:channel, :joins],
```
chasers, line 32: This telemetry stuff should change. We probably don't need this to emit metrics at all.

chasers (changes requested):

lib/realtime/tenants/connect/backoff.ex
```diff
@@ -0,0 +1,45 @@
...
+  """
+  alias Realtime.RateCounter
+  alias Realtime.GenCounter
+  alias Realtime.Tenants
+  @behaviour Realtime.Tenants.Connect.Piper
+
+  @impl Realtime.Tenants.Connect.Piper
+  def run(acc) do
+    %{tenant_id: tenant_id} = acc
+    connect_backoff_limit = Application.get_env(:realtime, :connect_backoff_limit)
```
edgurgel, line 13: Should we use `fetch_env!` so there is no way this is returning `nil` because the key is absent?

lib/realtime/tenants/connect/backoff.ex
```diff
@@ -0,0 +1,45 @@
...
+      GenCounter.add(counter)
+      {:ok, acc}
+    else
+      _ -> {:error, :tenant_create_backoff}
+    end
+  end
+
+  defp start_connects_per_second_counter(tenant_id, limit) do
+    id = Tenants.connection_attempts_per_second_key(tenant_id)
+    GenCounter.new(id)
```
edgurgel, line 26: Do we need to do this given that RateCounter creates them on `init` ?

https://github.com/supabase/realtime/blob/5ddfbc534c8950ee883656071bb35e126b9e469a/lib/realtime/rate_counter/rate_counter.ex#L218-L221

lib/realtime/tenants.ex
```diff
@@ -210,6 +212,23 @@ defmodule Realtime.Tenants do
...
+  The GenCounter key to use when counting connection attempts against Realtime.Tenants.Connect
+  ## Examples
+    iex> Realtime.Tenants.connection_attempts_per_second_key("tenant_id")
+    {:channel, :connection_attempts, "tenant_id"}
+    iex> Realtime.Tenants.connection_attempts_per_second_key(%Realtime.Api.Tenant{external_id: "tenant_id"})
+    {:channel, :connection_attempts, "tenant_id"}
+  """
+  @spec connection_attempts_per_second_key(Tenant.t() | String.t()) :: {:channel, :connection_attempts, String.t()}
+  def connection_attempts_per_second_key(tenant) when is_binary(tenant) do
+    {:channel, :connection_attempts, tenant}
```
edgurgel, line 225: Why `:channel` is used here? Is it related to a specific channel in this context?

lib/realtime/tenants/connect/backoff.ex
```diff
@@ -0,0 +1,45 @@
+defmodule Realtime.Tenants.Connect.Backoff do
```
jonatanklosko, line 1: Backoff typically means a gradually increasing delay between retry attempts. Perhaps a more accurate name would be Throttler or RateLimiter.

For the config option, instead of `connect_backoff_limit`, it could be `connect_throttle_limit`, `connect_throttle_threshold` (it is a threshold on the average), or similar.

jonatanklosko (commented):
Hey @filipecabaco!

I looked at the counter modules, I have two notes:

1. I noticed you use `:erlang.phash2/1`:
   https://github.com/supabase/realtime/blob/5ddfbc534c8950ee883656071bb35e126b9e469a/lib/realtime/gen_counter/gen_counter.ex#L36
   This function isn't the best at generating unique values. It has 2^32 possible return values, which is relatively low for hashing. A typical usage of phash would be to deterministically assign value to a partition (in fact, `phash2` accepts a second argument N, to cap the range). It is fair to expect collisions are very much possible.

   What you could do is lookup by `phash2`, but handle a case where there are multiple results and filter those based on the exact keys. That said, as long as you use simple terms (I see you use a tuple with two atoms and id), it's most likely fine to not hash altogether.

2. Is there any reason `GenCounter` starts a process? It seems to me you only really use the registry to store and lookup keys, but not the process itself. If that's the case, what you actually want is to store the counters in an ETS table (which is exactly what Registry uses under the hood).

lib/realtime/tenants.ex
filipecabaco: moving it to `{:connection_attempts, tenant}`

lib/realtime/tenants/connect/backoff.ex
filipecabaco: will keep it for now as I did see in the past some issues with this but unsure if this is the fix... avoiding it now 😅

lib/realtime/tenants/connect/backoff.ex
filipecabaco: will do

lib/realtime/tenants/connect/backoff.ex
filipecabaco: removed

chasers (approved):

edgurgel (approved):

PR #1352, opened 2025-04-25
fix: seed file temporarily starts partition supervisor for migrations

filipecabaco:
## What kind of change does this PR introduce?

seed file temporarily starts partition supervisor for migrations

Issue #1351, opened 2025-04-25
update from v2.34.43 to v2.34.52 failed


shugotek:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

On a kubernetes deployment, I update image of realtime from 2.34.43 to 2.34.52 and now have an error at startup during migration part : 
```
init-db postgresql.default.svc.cluster.local:5432 - accepting connections
supabase-realtime 07:11:47.188 [info] Migrations already up
supabase-realtime 07:11:51.077 [info] Replica region not found, defaulting to Realtime.Repo
supabase-realtime ** (exit) exited in: PartitionSupervisor.partitions(Realtime.Tenants.Migrations.DynamicSupervisor)
supabase-realtime     ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
supabase-realtime     (elixir 1.17.3) lib/partition_supervisor.ex:345: PartitionSupervisor.name_partitions/1
supabase-realtime     (elixir 1.17.3) lib/partition_supervisor.ex:431: PartitionSupervisor.whereis_name/1
supabase-realtime     (elixir 1.17.3) lib/gen_server.ex:1300: GenServer.whereis/1
supabase-realtime     (elixir 1.17.3) lib/gen_server.ex:1119: GenServer.call/3
supabase-realtime     (realtime 2.34.52) lib/realtime/tenants/migrations.ex:162: Realtime.Tenants.Migrations.run_migrations/1
supabase-realtime     /app/lib/realtime-2.34.52/priv/repo/seeds.exs:45: (file)
supabase-realtime     /app/lib/realtime-2.34.52/priv/repo/seeds.exs:13: (file)
```

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Deploy a Supabase  with realtime in version 2.34.43
2. Wait finish of deployment
3. Update image of realtime to 2.34.52

## Expected behavior

Update Realtime to 2.34.52

## System information

- OS: Node on Amazon Linux 2023 / Official docker image of Supabase Realtime (Kubernetes 1.32)

## Additional context

Can it be related to change on migration on version between 2.34.43 and 2.34.52  (#1344)?

shugotek:
[...]

PR #1356, opened 2025-05-02
Fix seed generation

edgurgel:

* Fix dev seeds by avoiding usage of the Cache module.
* Run ecto.setup (which includes dev seeds) on CI so we know if it breaks as it's not something someone would run all the time

`make seed` returns this:

```
** (KeyError) key :migrations_ran not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (realtime 2.34.53) lib/realtime/tenants.ex:304: Realtime.Tenants.run_migrations?/1
    (realtime 2.34.53) lib/realtime/tenants/migrations.ex:163: Realtime.Tenants.Migrations.run_migrations/1
    (ecto_sql 3.11.3) lib/ecto/adapters/sql.ex:1358: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
    (db_connection 2.7.0) lib/db_connection.ex:1756: DBConnection.run_transaction/4
    priv/repo/seeds.exs:13: (file)
```

## What is the new behavior?

seeds work

## Additional context

I was trying to reset my local db with seeds and I ran into this issue where `Tenants.run_migrations?("realtime-dev")` was failing because it couldn't find the tenant. 

Tracing it out I the problem is that `run_migrations?` was using `Cache.get_tenant_by_external_id` and this was inside a transaction where the tenant was just created inside seeds.exs: https://github.com/supabase/realtime/blob/e2275cd3f3cc938b2d14be5dcaf4100f2282c242/priv/repo/seeds.exs#L13-L46

`tenant = Tenants.get_tenant_by_external_id(tenant_name)` would work fine but `Tenants.Migrations.run_migrations(tenant)` would fail and this error would show up:

```
** (KeyError) key :migrations_ran not found in: nil

If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
    (realtime 2.34.53) lib/realtime/tenants.ex:304: Realtime.Tenants.run_migrations?/1
    (realtime 2.34.53) lib/realtime/tenants/migrations.ex:163: Realtime.Tenants.Migrations.run_migrations/1
    (ecto_sql 3.11.3) lib/ecto/adapters/sql.ex:1358: anonymous fn/3 in Ecto.Adapters.SQL.checkout_or_transaction/4
    (db_connection 2.7.0) lib/db_connection.ex:1756: DBConnection.run_transaction/4
    priv/repo/seeds.exs:13: (file)
```

The issue was that running `Cache.get_tenant_by_external_id` will be called by Cachex https://github.com/supabase/realtime/blob/e2275cd3f3cc938b2d14be5dcaf4100f2282c242/lib/realtime/context_cache.ex#L12

And Cachex uses a separate process outside of the process where the `Repo.transaction` is happening so the Tenant is not found because it ends up using a different Postgres connection.

In the end I noticed that we only have one place using `Tenant.run_migrations?/1` and already has the Tenant in memory so I just updated it to use it instead of fetching from Cache. 

I found this discussion on elixir forums https://elixirforum.com/t/ecto-timeout-errors-when-wrapping-into-cachex-calls-in-tests/19078 which is sort of similar

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -19,7 +19,7 @@ on:
 permissions:
   contents: read
 jobs:
-  format:
+  tests:
     name: Tests
     runs-on: ubuntu-latest
 
@@ -70,6 +70,8 @@ jobs:
           mix dialyzer.build
       - name: Run dialyzer
         run: mix dialyzer
+      - name: Run dev seeds
+        run: DB_ENC_KEY="1234567890123456" mix ecto.setup
       - name: Start epmd
         run: epmd -daemon
       - name: Run tests
--- a/lib/realtime/tenants.ex
+++ b/lib/realtime/tenants.ex
@@ -317,9 +317,8 @@ defmodule Realtime.Tenants do
   @doc """
   Checks if migrations for a given tenant need to run.
   """
-  @spec run_migrations?(binary()) :: boolean()
-  def run_migrations?(external_id) do
-    tenant = Cache.get_tenant_by_external_id(external_id)
+  @spec run_migrations?(Tenant.t()) :: boolean()
+  def run_migrations?(%Tenant{} = tenant) do
     tenant.migrations_ran < Enum.count(Migrations.migrations())
   end
 
--- a/lib/realtime/tenants/migrations.ex
+++ b/lib/realtime/tenants/migrations.ex
@@ -160,7 +160,7 @@ defmodule Realtime.Tenants.Migrations do
 
     spec = {__MODULE__, attrs}
 
-    if Tenants.run_migrations?(tenant.external_id) do
+    if Tenants.run_migrations?(tenant) do
       case DynamicSupervisor.start_child(supervisor, spec) do
         :ignore -> :ok
         error -> error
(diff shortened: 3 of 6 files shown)


filipecabaco (changes requested):
update mix.exs

edgurgel:
Version bumped!

filipecabaco:
shall we also add this changes?
https://github.com/supabase/realtime/pull/1352

edgurgel:
@filipecabaco I don't think that change is needed tbh. The application is started when you call `mix run priv/repo/seeds.exs` so that Supervisor is there for sure. If we were running `mix run --no-start priv/repo/seeds.exs` then yeah the supervisor would not be available.

But while testing this out I noticed that we only "reset" the tenant but we don't do the same for the tenant database. So I changed the seeds to also do that. WDYT?

filipecabaco:
We can't reset the tenant database as this file is also used by our self hosting and xli logic. if we reset the tenant we end up deleting their data.

We should have two separate seed files: one for testing and another for self hosting / cli

filipecabaco:
The bug that requires partition supervisor to start manifests in self hosting, I was able to replicate it using the docker compose provided in the Supabase repository 

https://github.com/supabase/realtime/issues/1351

edgurgel:
Closing in favor of https://github.com/supabase/realtime/pull/1367

PR #1361, opened 2025-05-06
fix: adjust gh actions

filipecabaco:

Adjusts GH actions to:
* set proper permissions
* only trigger tests and validations on relevant changes
* remove manual trigger actions


.github/workflows/version_updated.yml
```diff
@@ -3,6 +3,12 @@ on:
     types: [opened, reopened, synchronize]
     branches:
       - "main"
+    paths:
```
edgurgel, line 6: Nice! Would we want to include Dockerfile, priv, assets, rel, run.sh here?

filipecabaco: good call will update

filipecabaco: not only this one but the others 👍

filipecabaco: kept the test workflow focused on exs

edgurgel (approved):

PR #1372, opened 2025-05-13
ci: test impact of larger runner on tests

kevcodez:

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,10 +20,10 @@ on:
 jobs:
   tests:
     name: Tests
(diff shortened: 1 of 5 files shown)


kevcodez:
Findings:
- Ton of test time comes from random sleeps. Some up to 59 seconds. More Compute resources don't help
- We use PG14 for DB migration tests and then PG15 for the test runs, causing us to pull two different images, further slowing down tests, would recommend aligning

Issue #1369, opened 2025-05-09
Janitor's 10-minute delay causes silent failures in local development


jumski:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The Janitor process in Realtime has a 10-minute startup delay (`janitor_run_after_in_ms: 600000`) before creating partitions for the `realtime.messages` table. This causes silent failures when using `realtime.send()` in local development after database resets, as the function expects partitions to exist but doesn't create them on demand.

## To Reproduce

**Scenario 1: Fresh Install or First Startup**
1. Set up Supabase locally
2. Immediately try to use `realtime.send()` in a function or query
3. The function appears to succeed but no messages are stored
4. Check `realtime.messages` and find no data

**Scenario 2: After Database Reset**
1. Run `supabase db reset` (which wipes out existing partitions)
2. Try to use `realtime.send()` in a function or query
3. The function appears to succeed but no messages are stored
4. Behavior persists until the Janitor's next scheduled run (up to 4 hours later)

Example SQL that fails silently:
```sql
SELECT realtime.send(
  jsonb_build_object('data', 'test'),
  'test_event',
  'test_topic',
  true
);
```

Note: Simply restarting Supabase without a database reset doesn't cause this issue, as partitions persist in the database. The Janitor proactively creates partitions for today and the next 3 days, so day transitions are handled automatically once it has run initially.

## Expected behavior

Either:
1. The Janitor should run with minimal delay in development environments
2. The `realtime.send()` function should create the necessary partition if it doesn't exist
3. At minimum, `realtime.send()` should return an error rather than failing silently

## System information

- OS: Linux
- Version of Supabase CLI: v2.21.1

## Additional context

I found this issue while writing pgTAP tests that needed to verify Realtime notifications were working. Messages were silently failing to be stored because partitions didn't exist.

I've verified that the Janitor is enabled in the Supabase CLI environment (`RUN_JANITOR=true`), but the 10-minute delay is causing problems for development workflows where database resets are common.

While this 10-minute delay might be reasonable in production where services run continuously, it's particularly problematic in development environments where developers frequently reset their databases.

The current workaround is to manually create the partition:
```sql
DO $$
DECLARE
  today date := current_date;
[...]

PR #1383, opened 2025-05-23
fix: ensure that messages partitions are created for self hosted environment.  

edgurgel:

Ensure that messages partitions are created for self hosted environment.  It also logs on success or on error so that's easier to know if seeding worked ✅ 

It fixes https://github.com/supabase/realtime/issues/1369

It only ensures that migrations are executed.

The user would need to wait a significant amount of time for the Janitor to run.

## What is the new behavior?

Seeds now run `Janitor.MaintenanceTask`

## Additional context

### DIFF
--- a/lib/realtime/tenants/janitor.ex
+++ b/lib/realtime/tenants/janitor.ex
@@ -8,11 +8,7 @@ defmodule Realtime.Tenants.Janitor do
 
   import Realtime.Logs
 
-  alias Realtime.Api.Tenant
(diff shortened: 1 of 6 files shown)


filipecabaco (approved):

filipecabaco (changes requested):
mix.exs needs updating

edgurgel:
Oops! Done!

filipecabaco (approved):

PR #1393, opened 2025-05-29
fix: unassign db_conn from socket to prevent dead connections

filipecabaco:

unassign db_conn from socket to prevent dead connections

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -93,8 +93,7 @@ defmodule RealtimeWeb.RealtimeChannel do
         presence_key: presence_key(params),
         self_broadcast: !!params["config"]["broadcast"]["self"],
         tenant_topic: tenant_topic,
-        channel_name: sub_topic,
-        db_conn: db_conn
+        channel_name: sub_topic
       }
 
       # Start presence and add user
@@ -343,13 +342,13 @@ defmodule RealtimeWeb.RealtimeChannel do
       assigns: %{
         access_token: access_token,
         pg_sub_ref: pg_sub_ref,
-        db_conn: db_conn,
         channel_name: channel_name,
         pg_change_params: pg_change_params
       }
     } = socket
 
     socket = assign(socket, :access_token, refresh_token)
+    {:ok, db_conn} = Connect.lookup_or_start_connection(socket.assigns.tenant)
 
     with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
          socket = assign_authorization_context(socket, channel_name, access_token, claims),
@@ -541,13 +540,15 @@ defmodule RealtimeWeb.RealtimeChannel do
   defp confirm_token(%{assigns: assigns} = socket) do
     %{
       jwt_secret: jwt_secret,
-      access_token: access_token
+      access_token: access_token,
+      tenant: tenant_id
     } = assigns
 
     topic = Map.get(assigns, :topic)
-    db_conn = Map.get(assigns, :db_conn)
     socket = Map.put(socket, :policies, nil)
     jwt_jwks = Map.get(assigns, :jwt_jwks)
+
+    db_conn = Connect.lookup_or_start_connection(tenant_id)
 
     with jwt_secret_dec <- Crypto.decrypt!(jwt_secret),
          {:ok, %{"exp" => exp} = claims} when is_integer(exp) <-
@@ -705,11 +706,7 @@ defmodule RealtimeWeb.RealtimeChannel do
     assign(socket, :authorization_context, authorization_context)
   end
 
-  defp maybe_assign_policies(
-         topic,
-         db_conn,
-         %{assigns: %{private?: true}} = socket
-       )
+  defp maybe_assign_policies(topic, db_conn, %{assigns: %{private?: true}} = socket)
        when not is_nil(topic) and not is_nil(db_conn) do
     authorization_context = socket.assigns.authorization_context
 
--- a/lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
+++ b/lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
@@ -9,6 +9,7 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
   alias Phoenix.Socket
   alias Realtime.GenCounter
   alias Realtime.RateCounter
+  alias Realtime.Tenants.Connect
   alias Realtime.Tenants.Authorization
   alias Realtime.Tenants.Authorization.Policies
   alias Realtime.Tenants.Authorization.Policies.BroadcastPolicies
@@ -23,9 +24,11 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
         self_broadcast: self_broadcast,
         tenant_topic: tenant_topic,
         authorization_context: authorization_context,
-        db_conn: db_conn
+        tenant: tenant_id
       }
     } = socket
+
+    {:ok, db_conn} = Connect.lookup_or_start_connection(tenant_id)
 
     case run_authorization_check(socket, db_conn, authorization_context) do
       {:ok,
--- a/lib/realtime_web/channels/realtime_channel/presence_handler.ex
+++ b/lib/realtime_web/channels/realtime_channel/presence_handler.ex
@@ -12,6 +12,7 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
   alias Phoenix.Tracker.Shard
   alias Realtime.GenCounter
   alias Realtime.RateCounter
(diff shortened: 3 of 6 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -343,13 +342,13 @@ defmodule RealtimeWeb.RealtimeChannel do
...
         access_token: access_token,
         pg_sub_ref: pg_sub_ref,
-        db_conn: db_conn,
         channel_name: channel_name,
         pg_change_params: pg_change_params
       }
     } = socket
 
     socket = assign(socket, :access_token, refresh_token)
+    {:ok, db_conn} = Connect.lookup_or_start_connection(socket.assigns.tenant)
```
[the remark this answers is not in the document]
filipecabaco: yes we do

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -23,10 +24,12 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
         self_broadcast: self_broadcast,
         tenant_topic: tenant_topic,
         authorization_context: authorization_context,
-        db_conn: db_conn
+        tenant: tenant_id
       }
     } = socket
 
+    {:ok, db_conn} = Connect.lookup_or_start_connection(tenant_id)
```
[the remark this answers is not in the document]
filipecabaco: yes 👍

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -362,6 +360,7 @@ defmodule RealtimeWeb.RealtimeChannel do
 
     with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
          socket = assign_authorization_context(socket, channel_name, access_token, claims),
+         {:ok, db_conn} <- Connect.lookup_or_start_connection(socket.assigns.tenant),
```
edgurgel, line 392: Should the else here also cover the connect errors like we do on join?

```elixir
      {:error, :tenant_database_unavailable} ->
        Logging.log_error_message(
          :error,
          "UnableToConnectToProject",
          "Realtime was unable to connect to the project database"
        )

      {:error, :too_many_connections} ->
        msg = "Too many connected users"
        Logging.log_error_message(:error, "ConnectionRateLimitReached", msg)


      {:error, :tenant_db_too_many_connections} ->
        msg = "Database can't accept more connections, Realtime won't connect"
        Logging.log_error_message(:error, "DatabaseLackOfConnections", msg)


      {:error, :initializing} ->
        Logging.log_error_message(
          :error,
          "InitializingProjectConnection",
          "Realtime is initializing the project connection"
        )

      {:error, :tenant_database_connection_initializing} ->
        Logging.log_error_message(
          :error,
          "InitializingProjectConnection",
          "Connecting to the project database"
        )
```

We could probably have a generic

```elixir
defp connect(tenant) do
   case Connect.lookup_or_start_connection(tenant) do
     {:ok, db_conn} -> {:ok, db_conn}

     {:error, :tenant_database_unavailable} = error ->
        Logging.log_error_message(
          :error,
          "UnableToConnectToProject",
          "Realtime was unable to connect to the project database"
        )
        error

      {:error, :too_many_connections} = error->
        msg = "Too many connected users"
        Logging.log_error_message(:error, "ConnectionRateLimitReached", msg)
        error


      {:error, :tenant_db_too_many_connections} = error ->
        msg = "Database can't accept more connections, Realtime won't connect"
        Logging.log_error_message(:error, "DatabaseLackOfConnections", msg)
        error


      {:error, :initializing} = error ->
        Logging.log_error_message(
          :error,
          "InitializingProjectConnection",
          "Realtime is initializing the project connection"
        )
        error

      {:error, :tenant_database_connection_initializing} = error ->
        Logging.log_error_message(
          :error,
          "InitializingProjectConnection",
          "Connecting to the project database"
        )
        error
        error -> error
   end
end
```

then at least it will log appropriately everywhere it fails to connect I guess?

WDYT?

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -23,24 +24,26 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
     } = socket
 
-    case run_authorization_check(socket, db_conn, authorization_context) do
-      {:ok,
-       %{assigns: %{ack_broadcast: ack_broadcast, policies: %Policies{broadcast: %BroadcastPolicies{write: true}}}} =
-           socket} ->
-        socket = increment_rate_counter(socket)
-        send_message(self_broadcast, tenant_topic, payload)
-        if ack_broadcast, do: {:reply, :ok, socket}, else: {:noreply, socket}
+    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id) do
```
edgurgel: This with without an else will return `{:error, _reason}` if `Connect.lookup_or_start_connection/1` fails but the `@spec` says that this function must return `{:reply` or `{:noreply`. 

I was expecting dialyzer to complain there 🤔 

We may need some tests on BroadcastHandler and PresenceHandler where Connect fails to give a db connection?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -67,15 +68,17 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
          payload,
          %{assigns: %{private?: true, policies: %Policies{presence: %PresencePolicies{write: nil}}}} = socket
        ) do
-    %{assigns: %{db_conn: db_conn, authorization_context: authorization_context}} = socket
+    %{assigns: %{authorization_context: authorization_context, tenant: tenant_id}} = socket
 
-    case run_authorization_check(socket, db_conn, authorization_context) do
-      {:ok, socket} ->
-        handle_presence_event("track", payload, socket)
+    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id) do
```
edgurgel: This with without else will return `{:error, _reason}` but I think the callers expect `{:error, socket}` instead

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
edgurgel: Another option is to really pass `db_conn` here as an argument and leave for the caller (RealtimeChannel) to call `Connect` as it already has all the error handling done ? I don't have a strong opinion about this tbh.

I'm just worried about not logging that Connect failed to set-up a connection and not surfacing this to users.

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: this could lead to multiple logs which will make it a bit confusing right?

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
filipecabaco: maybe just passing the conn is better as that will allow us to have a single point of calling

filipecabaco:
@edgurgel added the move from handlers to RealtimeChannel so we centralize those calls.

For now we just have a "catch all" error and we can refine afterwards when we have a better understanding of what is actually the errors being caught

I also did the test by brute force change the syn state which won't trigger the Connect module exit flows

lib/realtime_web/channels/realtime_channel.ex
edgurgel: True. We will just end up pushing this to the websocket. Not sure if that's a problem or not but I thought I would mention anyway

```
    {:error, error} ->
shutdown_response(socket, inspect(error))
```

edgurgel (approved):
LGTM

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: wondering what we should do there 🤔 will check tomorrow and see a better output as we would be potentially showing too much information to external users...

edgurgel:
Sorry for the conflicts! But at least it should be pretty straightforward to fix those 🙏

test/realtime_web/channels/realtime_channel/presence_handler_test.exs
```diff
@@ -114,14 +111,14 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandlerTest do
 
       for _ <- 1..100, reduce: socket do
         socket ->
-          assert {:reply, :ok, socket} =
+          assert {:reply, :ok, _} =
```
edgurgel: This is the issue! socket needed to be passed around to be `reduced` and kept the policies that were cached etc!

I've pushed a fix!

filipecabaco: 🤦‍♂️🤦‍♂️🤦‍♂️

PR #1395, opened 2025-05-29
fix: add private toggle on inspector to improve debugging capabilities

filipecabaco:

add private toggle on inspector to improve debugging capabilities

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -20,7 +20,8 @@ Hooks.payload = {
     filter,
     bearer,
     enable_presence,
-    enable_db_changes
+    enable_db_changes,
+    private_channel
   ) {
     // Instantiate our client with the Realtime server and params to connect with
     {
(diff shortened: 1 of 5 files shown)


chasers:
You need to pick that field up in the app.js hook also

filipecabaco:
will fix

assets/js/app.js
```diff
@@ -43,7 +44,7 @@ Hooks.payload = {
     // Channels can be named anything
     // All clients on the same Channel will get messages sent to that Channel
     this.channel = this.realtimeSocket.channel(channelName, {
-      config: { broadcast: { self: true } },
+      config: { broadcast: { self: true, private: private_channel } },
```
chasers, line 47: Pretty sure this comes over as a string so you probably need to convert it to a bool

chasers (changes requested):

chasers (approved):

PR #1407, opened 2025-06-03
fix: register locally after syn

filipecabaco:

In Connect and PostgresCDC, after syn registration, it also registers into a Registry so we can use it in the Node for easier debugging and support operations

### DIFF
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -137,13 +137,6 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
     {:noreply, prepare_replication(state)}
   end
 
-  @impl true
-  def terminate(reason, state) do
-    Logger.error("Terminating ReplicationPoller: #{inspect(reason)}")
-    Registry.unregister(__MODULE__.Registry, state.tenant)
-    {:stop, reason, state}
-  end
-
   def slot_name_suffix do
     case Application.get_env(:realtime, :slot_name_suffix) do
       nil -> ""
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -45,6 +45,7 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
       tenant: tenant
     }
 
+    Registry.register(__MODULE__.Registry, tenant, %{})
     {:ok, state, {:continue, {:connect, args}}}
   end
 
@@ -134,6 +135,13 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
   def handle_info(:retry, %{retry_ref: retry_ref} = state) do
     cancel_timer(retry_ref)
     {:noreply, prepare_replication(state)}
+  end
+
+  @impl true
+  def terminate(reason, state) do
+    Logger.error("Terminating ReplicationPoller: #{inspect(reason)}")
+    Registry.unregister(__MODULE__.Registry, state.tenant)
+    {:stop, reason, state}
   end
 
   def slot_name_suffix do
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -45,7 +45,7 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
       tenant: tenant
     }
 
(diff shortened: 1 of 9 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -354,6 +354,7 @@ defmodule Realtime.Tenants.Connect do
   def terminate(reason, %{tenant_id: tenant_id}) do
     Logger.info("Tenant #{tenant_id} has been terminated: #{inspect(reason)}")
     Realtime.MetricsCleaner.delete_metric(tenant_id)
+    Registry.unregister(__MODULE__.Registry, tenant_id)
```
chasers: Why again do we want to unregister specifically?

filipecabaco: no need, removed it

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -45,6 +45,7 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
       tenant: tenant
     }
 
+    Registry.register(__MODULE__.Registry, tenant, %{})
```
edgurgel: Do we care if this errors out with `{:error, {:already_registered, pid()}}` ?

lib/realtime/tenants/connect/register_process.ex
```diff
@@ -2,14 +2,17 @@ defmodule Realtime.Tenants.Connect.RegisterProcess do
...
+  alias Realtime.Tenants.Connect
   @behaviour Realtime.Tenants.Connect.Piper
 
   @impl true
   def run(acc) do
     %{tenant_id: tenant_id, db_conn_pid: conn} = acc
 
-    case :syn.update_registry(Realtime.Tenants.Connect, tenant_id, fn _pid, meta -> %{meta | conn: conn} end) do
-      {:ok, _} -> {:ok, acc}
+    with {:ok, _} <- :syn.update_registry(Connect, tenant_id, fn _pid, meta -> %{meta | conn: conn} end),
```
edgurgel, line 12: Should we first register locally? 

If local register fails we won't spend time sending syn metadata to all nodes I think?

test/realtime/tenants/connect/register_process_test.exs
```diff
@@ -1,20 +1,22 @@
...
       # Warm cache to avoid Cachex and Ecto.Sandbox ownership issues
       Cachex.put!(Realtime.Tenants.Cache, {{:get_tenant_by_external_id, 1}, [tenant.external_id]}, {:cached, tenant})
-
-      conn = start_supervised!({Connect, [tenant_id: tenant.external_id]})
+      {:ok, conn} = Database.connect(tenant, "realtime_test")
       %{tenant_id: tenant.external_id, db_conn_pid: conn}
     end
 
-    test "registers the process in :syn", %{tenant_id: tenant_id, db_conn_pid: conn} do
+    test "registers the process in syn and Registry and updates metadata", %{tenant_id: tenant_id, db_conn_pid: conn} do
```
edgurgel, line 15: Should we also check if `Connect.Registry` sees the new process?

test/realtime/tenants/connect/register_process_test.exs
```diff
@@ -1,20 +1,22 @@
...
+      # Fake the process registration in :syn
+      :syn.register(Realtime.Tenants.Connect, tenant_id, self(), %{conn: nil})
       assert {:ok, _} = RegisterProcess.run(%{tenant_id: tenant_id, db_conn_pid: conn})
+      assert {_, %{conn: ^conn}} = :syn.lookup(Realtime.Tenants.Connect, tenant_id)
     end
 
     test "handles undefined process error" do
```
edgurgel, line 39: Do we want to add a new test when `Connect.Registry` already has a process registered?

lib/extensions/postgres_cdc_rls/replication_poller.ex
filipecabaco: we should yes as that will show some errors that can be useful for us to understand what is happening

lib/realtime/tenants/connect/register_process.ex
filipecabaco: honestly not sure... 😓 I thought about it but this felt more "natural" in the sense that we would have a better understanding of what seen is doing...

edgurgel: Sure we can revisit this if we need to

edgurgel (approved):

PR #1413, opened 2025-06-05
fix: shutdown connect on relevant settings changes

filipecabaco:

shutdown connect on relevant settings changes


test/realtime/api_test.exs
```diff
@@ -117,30 +123,41 @@ defmodule Realtime.ApiTest do
...
       :ok = Phoenix.PubSub.subscribe(Realtime.PubSub, "realtime:operations:" <> tenant.external_id)
-
       assert {:ok, %Tenant{}} = Api.update_tenant(tenant, %{max_events_per_second: 100})
       refute_receive :disconnect, 500
     end
+
+    test "valid data and jwt_secret change will restart the database connection", %{tenants: [tenant | _]} do
+      {:ok, old_pid} = Connect.lookup_or_start_connection(tenant.external_id)
+      assert {:ok, %Tenant{}} = Api.update_tenant(tenant, %{jwt_secret: "potato"})
+      Process.sleep(100)
```
edgurgel: Instead of sleeping we can monitor and assert DOWN like we do here:

https://github.com/supabase/realtime/blob/42fc73da9061b0107114a93600d244d9551b3dec/test/realtime/tenants/connect_test.exs#L22-L25

It won't wait more than the time needed for the monitor message to arrive

test/realtime/api_test.exs
```diff
@@ -117,30 +123,41 @@ defmodule Realtime.ApiTest do
...
+      refute Process.alive?(old_pid)
+      assert {:ok, new_pid} = Connect.lookup_or_start_connection(tenant.external_id)
+      refute old_pid == new_pid
+    end
+
+    test "valid data and max_concurrent_users change will restart the database connection", %{tenants: [tenant | _]} do
+      {:ok, old_pid} = Connect.lookup_or_start_connection(tenant.external_id)
+      assert {:ok, %Tenant{}} = Api.update_tenant(tenant, %{max_concurrent_users: 100})
+      Process.sleep(100)
```
edgurgel: same here

edgurgel (approved):
Add a couple of comments but they are not blockers

lib/realtime/api.ex
```diff
@@ -112,23 +124,15 @@ defmodule Realtime.Api do
 
   """
   def update_tenant(%Tenant{} = tenant, attrs) do
-    tenant
-    |> Tenant.changeset(attrs)
-    |> tap(&maybe_trigger_disconnect/1)
-    |> Repo.update()
-  end
+    changeset = Tenant.changeset(tenant, attrs)
+    updated = Repo.update(changeset)
```
edgurgel, line 128: Do we want to only processed with the `maybe_`s if this was an `{:ok, _}` ?

test/realtime/api_test.exs
```diff
@@ -9,13 +9,21 @@ defmodule Realtime.ApiTest do
...
-    tenant_fixture(%{max_concurrent_users: 10_000_000})
-    tenant_fixture(%{max_concurrent_users: 25_000_000})
+    tenant1 = Containers.checkout_tenant(run_migrations: true)
+    tenant2 = Containers.checkout_tenant(run_migrations: true)
+    {:ok, tenant1} = Api.update_tenant(tenant1, %{max_concurrent_users: 10_000_000})
+    {:ok, tenant2} = Api.update_tenant(tenant2, %{max_concurrent_users: 20_000_000})
+
+    # Warm cache to avoid Cachex and Ecto.Sandbox ownership issues
+    Cachex.put!(Cache, {{:get_tenant_by_external_id, 1}, [tenant1.external_id]}, {:cached, tenant1})
+    Cachex.put!(Cache, {{:get_tenant_by_external_id, 1}, [tenant2.external_id]}, {:cached, tenant2})
```
edgurgel: Given that it's running on async: false this can probably be removed completely

filipecabaco: that will be changed 😂 forgot

lib/realtime/api.ex
filipecabaco: good call

lib/realtime/api.ex
```diff
@@ -225,4 +230,32 @@ defmodule Realtime.Api do
...
+  defp maybe_trigger_disconnect(%Ecto.Changeset{data: %{external_id: external_id}} = changeset)
+       when requires_disconnect(changeset) do
+    Phoenix.PubSub.broadcast!(Realtime.PubSub, "realtime:operations:" <> external_id, :disconnect)
+  end
+
+  defp maybe_trigger_disconnect(_changeset), do: nil
+
+  defp maybe_restart_db_connection(%Ecto.Changeset{data: %{external_id: external_id}} = changeset)
+       when requires_restarting_db_connection(changeset) do
+    case Connect.lookup_or_start_connection(external_id) do
```
edgurgel: Instead of `Connect.lookup_or_start_connection` should we use `Connect.whereis` so that we don't start Connect if the tenant is inactive to then shut it down? 🤔 WDYT?

And to be fair it's probably safe to just call `Connect.shutdown(external_id)` as it will do whereis and then send a shutdown message if that's the case. 🤔

lib/realtime/api.ex
```diff
@@ -112,23 +124,15 @@ defmodule Realtime.Api do
...
+    updated = Repo.update(changeset)
 
-  defp maybe_trigger_disconnect(%Ecto.Changeset{
-         changes: changes,
-         valid?: true,
-         data: %{external_id: external_id}
-       })
-       when is_map_key(changes, :jwt_jwks) or is_map_key(changes, :jwt_secret) do
-    Phoenix.PubSub.broadcast!(Realtime.PubSub, "realtime:operations:" <> external_id, :disconnect)
+    maybe_invalidate_cache(changeset)
```
edgurgel: Is there a race condition here?

What if the invalidation takes a long time and the DB reconnects before the Tenant cache has been discarded 🤔 This way it might still connect to the database with the old  settings?

test/realtime/api_test.exs
filipecabaco: for some reason it's breaking with async true as such going to avoid it for now

test/realtime/api_test.exs
filipecabaco: actually it breaks 😓 without the sleep it seems that the `Connect.lookup_or_start_connection` returns the old pid 🤔

lib/realtime/api.ex
filipecabaco: I have no clue honestly ... we might just need to trust cachex? 😅

test/realtime/api_test.exs
filipecabaco: I pushed the changes to make it work...

lib/realtime/api.ex
```diff
@@ -225,4 +238,25 @@ defmodule Realtime.Api do
     |> Map.put(:events_per_second_rolling, avg)
     |> Map.put(:events_per_second_now, current)
   end
+
+  defp maybe_invalidate_cache(%Ecto.Changeset{changes: changes, valid?: true, data: %{external_id: external_id}})
+       when changes != %{} do
+    Tenants.Cache.invalidate_tenant_cache(external_id)
```
edgurgel: Should this be calling `distributed_invalidate_tenant_cache` ? Otherwise a remote node won't have the new settings when `Connect.lookup_or_start_connection` is called the next time. 

And that's what I mean by race condition:

* Tell all nodes to invalidate cache (broadcast)
* Tell Connect to shutdown (broadcast)

There is no real guarantee which one of the two above will be processed first on each node ^

Then at some point `Connect.lookup_or_start_connection` will be called and it's not 100% clear that the new settings will be available if this makes sense.

🤔

test/realtime/api_test.exs
edgurgel: Right that's what I meant. If you are leaving `async: false` then just remove the `Cachex.put` calls here. It shouldn't be needed

test/realtime/api_test.exs
edgurgel: Oh you are right. It's because `syn` needs to react to the monitor as well. :tableflip:

lib/realtime/api.ex
filipecabaco: yes it should

filipecabaco: honestly I have no clue what to do because there's no good way of guaranteeing anything 😅

edgurgel: I think a way to be 100% sure is to __not__ use a broadcast (cast, fire and forget) but use an RPC call that waits for a reply for every node. Otherwise we won't really know if a node has the new settings or not. 

:erpc has `multicall` for this use case

WDYT?

filipecabaco: that's a very good call as it will also free up more PubSub message passing ... will change in another PR how distributed cache works 👍

edgurgel (approved):

lib/realtime/api.ex
edgurgel: Sounds good!

PR #1417, opened 2025-06-10
Fix missing log and telemetry metadata

edgurgel:

Fix a few places where tenant/external_id/project information was missing on logs and telemetry.

Missing metadata

## What is the new behavior?

Metadata is set (when available) on:
* `Rpc.enhanced_call` error log statements
*  `Rpc.enhanced_call` telemetry
* Authorization -> Database.transaction error logs
* Connect error log statements

## Additional context

### DIFF
--- a/lib/realtime/database.ex
+++ b/lib/realtime/database.ex
@@ -164,7 +164,7 @@ defmodule Realtime.Database do
     metadata = Keyword.put(metadata, :target, node(db_conn))
     args = [db_conn, func, opts, metadata]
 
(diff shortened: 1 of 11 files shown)


lib/realtime/database.ex
```diff
@@ -205,7 +205,6 @@ defmodule Realtime.Database do
       ssl: ssl
     } = settings
 
-    Logger.metadata(application_name: application_name)
```
edgurgel, line 208: This was setting the metadata on the calling process ⛔

filipecabaco (approved):

PR #1421, opened 2025-06-16
fix: use multicall to invalidate cache

filipecabaco:

To prevent usage of PubSub for cache invalidation we're moving to use :erpc.multicall

Due to this, we can also simplify the supervision tree for the Caching layer


edgurgel (approved):
LGTM

PR #1425, opened 2025-06-18
fix: GenRpc broadcaster to use multicast

edgurgel:

Change the GenRpc broadcaster to use eval_everywhere so that we don't have to wait for all nodes to ack that they called Phoenix.local_broadcast. This way we have a more equivalent version of the Phoenix broadcaster

Using `multicall` and waiting for responses

## What is the new behavior?

Fire and forget. We still get errors logged if there is an issue with the gen_rpc tcp connection

## Additional context

### DIFF
--- a/lib/realtime/gen_rpc.ex
+++ b/lib/realtime/gen_rpc.ex
@@ -10,14 +10,24 @@ defmodule Realtime.GenRpc do
 
   @type result :: any | {:error, :rpc_error, reason :: any}
 
-  # Here we run the async_call on all nodes using gen_rpc except the local node
-  # This is because gen_rpc does not have a bypass for local node on multicall
-  # For the local node we use rpc instead
-  defp async_call({node, _}, mod, func, args) when node == node(), do: :rpc.async_call(node, mod, func, args)
-  defp async_call(node, mod, func, args), do: :gen_rpc.async_call(node, mod, func, args)
+  @doc """
+  Fire and forget apply(mod, func, args) on all nodes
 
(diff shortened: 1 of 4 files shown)


filipecabaco (approved):

lib/realtime/gen_rpc.ex
```diff
@@ -10,14 +10,24 @@ defmodule Realtime.GenRpc do
...
+  """
+  @spec multicast(module, atom, list(any), keyword()) :: :ok
+  def multicast(mod, func, args, opts \\ []) when is_atom(mod) and is_atom(func) and is_list(args) and is_list(opts) do
+    key = Keyword.get(opts, :key, nil)
+
+    nodes = rpc_nodes(Node.list(), key)
+
+    # Use erpc for the local node because :gen_rpc tries to connect with the local node
+    :ok = :erpc.cast(Node.self(), mod, func, args)
+    :gen_rpc.eval_everywhere(nodes, mod, func, args)
```
abc3, line 28: out of curiosity, why is it better than the builtin `:rpc.eval_everywhere/4` or `:erpc.multicast/4`?

abc3: Oh, it uses dedicated tcp connections instead of relying on erldist. Nice!

edgurgel: Yup! The main idea here is to leave erldist for process registry (syn, pg, etc) and move the bulk of broadcasts and calls to gen_rpc if everything goes well 🤞

PR #1433, opened 2025-06-24
fix: Reload endpoint updated to also shutdown Connect module

filipecabaco:

Reload endpoint updated to also shutdown Connect module

### DIFF
--- a/test/realtime_web/controllers/tenant_controller_test.exs
+++ b/test/realtime_web/controllers/tenant_controller_test.exs
@@ -8,6 +8,7 @@ defmodule RealtimeWeb.TenantControllerTest do
   alias Realtime.Api.Tenant
   alias Realtime.Crypto
   alias Realtime.Database
+  alias Realtime.PostgresCdc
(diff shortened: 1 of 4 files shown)


test/realtime_web/controllers/tenant_controller_test.exs
```diff
@@ -269,8 +270,22 @@ defmodule RealtimeWeb.TenantControllerTest do
...
+
+      {:ok, connect_pid} = Connect.lookup_or_start_connection(tenant.external_id)
+
+      assert Process.alive?(manager_pid)
+      assert Process.alive?(connect_pid)
+
       %{status: status} = post(conn, ~p"/api/tenants/#{tenant.external_id}/reload")
       assert status == 204
+      Process.sleep(1000)
+      refute Process.alive?(manager_pid)
```
edgurgel: Instead of sleeping we could monitor and assert_receive DOWN... etc and wait just the minimum amount of time to notice the process death 💀

edgurgel (approved):

PR #1434, opened 2025-06-24
fix: expose db event metrics via prometheus

filipecabaco:

currently we're not exposing db_event metrics via prometheus endpoint

### DIFF
--- a/lib/realtime/monitoring/prom_ex/plugins/tenant.ex
+++ b/lib/realtime/monitoring/prom_ex/plugins/tenant.ex
@@ -135,7 +135,7 @@ defmodule Realtime.PromEx.Plugins.Tenant do
           tags: [:tenant]
         ),
         sum(
-          [:realtime, :channel, :presence_events],
+          [:realtime, :channel, :db_events],
           event_name: [:realtime, :rate_counter, :channel, :db_events],
           measurement: :sum,
           description: "Sum of db messages sent on a Realtime Channel.",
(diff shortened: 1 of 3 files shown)


lib/realtime/monitoring/prom_ex/plugins/tenant.ex
```diff
@@ -127,11 +127,18 @@ defmodule Realtime.PromEx.Plugins.Tenant do
...
-          description: "Rate limit of messages per second sent on a Realtime Channel.",
+        sum(
+          [:realtime, :channel, :presence_events],
+          event_name: [:realtime, :rate_counter, :channel, :presence_events],
+          measurement: :sum,
+          description: "Sum of presence messages sent on a Realtime Channel.",
+          tags: [:tenant]
+        ),
+        sum(
+          [:realtime, :channel, :presence_events],
```
edgurgel: should this be [realtime, channel, db_events] instead?

filipecabaco: the power of copy paste... I will also try to add some tests as this is very meh not be tested

filipecabaco: actually checked again and looks good, I just moved up all the "events" to the same spot and ordered sum's, last values; distribition's etc... just readability thingies

will still add tests

filipecabaco: I retract that... LOL

PR #1438, opened 2025-06-26
fix: Listen and ReplicationConnection supervisor setup

edgurgel:

Use the same amount of partitions that Connect has given that they map 1:1

Also move ReplicationConnection and Listen supervisors above Connect as they need to be up and running before Connect can start

## Additional context


filipecabaco (approved):

README.md
```diff
@@ -160,6 +160,7 @@ If you're using the default tenant, the URL is `ws://realtime-dev.localhost:4000
...
 | JANITOR_CHUNK_SIZE                   | number  | Number of tenants to process per chunk. Each chunk will be processed by a Task                                                                                                                                                                                                                                                  |
 | MIGRATION_PARTITION_SLOTS            | number  | Number of dynamic supervisor partitions used by the migrations process                                                                                                                                                                                                                                                          |
+| CONNECT_PARTITION_SLOTS              | number  | Number of dynamic supervisor partitions used by the Connect, ReplicationConnect and Listen processes                                                                                                                                                                                                                            |
```
edgurgel, line 163: @filipecabaco we were both right! The env already exists but it was not documented:

https://github.com/supabase/realtime/blob/f2bef3168eebea212273a1745dfa42736f78b6c5/config/runtime.exs#L21-L22

PR #1442, opened 2025-06-30
fix: Connect keeps region in syn state

filipecabaco:

Connect keeps region in syn state

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -104,7 +104,7 @@ defmodule Realtime.Tenants.Connect do
   @doc """
   Connects to a tenant's database and stores the DBConnection in the process :syn metadata
   """
-  @spec connect(binary(), keyword()) :: {:ok, DBConnection.t()} | {:error, term()}
+  @spec connect(binary(), binary(), keyword()) :: {:ok, DBConnection.t()} | {:error, term()}
   def connect(tenant_id, region, opts \\ []) do
     supervisor =
       {:via, PartitionSupervisor, {Realtime.Tenants.Connect.DynamicSupervisor, tenant_id}}
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -52,23 +52,23 @@ defmodule Realtime.Tenants.Connect do
   @doc """
   Returns the database connection for a tenant. If the tenant is not connected, it will attempt to connect to the tenant's database.
   """
-  @spec lookup_or_start_connection(binary(), keyword()) ::
+  @spec lookup_or_start_connection(binary(), binary(), keyword()) ::
           {:ok, pid()}
           | {:error, :tenant_database_unavailable}
           | {:error, :initializing}
           | {:error, :tenant_database_connection_initializing}
           | {:error, :rpc_error, term()}
-  def lookup_or_start_connection(tenant_id, opts \\ []) when is_binary(tenant_id) do
+  def lookup_or_start_connection(tenant_id, region, opts \\ []) when is_binary(tenant_id) do
     case get_status(tenant_id) do
       {:ok, conn} ->
         {:ok, conn}
 
       {:error, :tenant_database_unavailable} ->
-        call_external_node(tenant_id, opts)
+        call_external_node(tenant_id, region, opts)
 
       {:error, :tenant_database_connection_initializing} ->
         Process.sleep(100)
-        call_external_node(tenant_id, opts)
+        call_external_node(tenant_id, region, opts)
 
       {:error, :initializing} ->
         {:error, :tenant_database_unavailable}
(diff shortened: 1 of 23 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -52,23 +52,23 @@ defmodule Realtime.Tenants.Connect do
...
   """
-  @spec lookup_or_start_connection(binary(), keyword()) ::
+  @spec lookup_or_start_connection(binary(), binary(), keyword()) ::
           {:ok, pid()}
           | {:error, :tenant_database_unavailable}
           | {:error, :initializing}
           | {:error, :tenant_database_connection_initializing}
           | {:error, :rpc_error, term()}
-  def lookup_or_start_connection(tenant_id, opts \\ []) when is_binary(tenant_id) do
+  def lookup_or_start_connection(tenant_id, region, opts \\ []) when is_binary(tenant_id) do
```
chasers: Should Connect.get_status take the whole tenant struct instead? 

Then lookup the region there?

Perhaps all these other things should not have to care about the region at all.

filipecabaco: it's more the tests that create the issues 🤔 and ideally we avoid passing the tenant as some of this stuff is called from RPC calls so it would mean sending a struct

chasers: Looks like we already have a Nodes.get_region which is called via Nodes.get_node_for_tenant

lib/realtime/tenants/connect.ex
```diff
@@ -359,13 +360,16 @@ defmodule Realtime.Tenants.Connect do
   end
 
   ## Private functions
-  defp call_external_node(tenant_id, opts) do
+  defp call_external_node(tenant_id, region, opts) do
     rpc_timeout = Keyword.get(opts, :rpc_timeout, @rpc_timeout_default)
 
     with tenant <- Tenants.Cache.get_tenant_by_external_id(tenant_id),
          :ok <- tenant_suspended?(tenant),
          {:ok, node} <- Realtime.Nodes.get_node_for_tenant(tenant) do
```
chasers: This could return `{:ok, region, node}` to pass to `connect/3`

lib/realtime/tenants/connect.ex
filipecabaco: it's a `defp` but didn't see it honestly. will probably have it all under `Tenants` so it's bounded to the context of the tenant

lib/realtime/tenants/connect.ex
filipecabaco: good call will check

lib/realtime/tenants/connect.ex
```diff
@@ -105,11 +105,11 @@ defmodule Realtime.Tenants.Connect do
   Connects to a tenant's database and stores the DBConnection in the process :syn metadata
   """
   @spec connect(binary(), keyword()) :: {:ok, DBConnection.t()} | {:error, term()}
```
edgurgel: should we update the spec as well?

mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.58.0",
+      version: "2.59.1",
```
edgurgel: 2.58.2 I think? Might need to rebase/merge with main first?

test/realtime/api_test.exs
```diff
@@ -137,7 +137,9 @@ defmodule Realtime.ApiTest do
     end
 
     test "valid data and jwt_secret change will restart the database connection", %{tenant: tenant} do
+      region = Realtime.Tenants.region(tenant)
```
edgurgel: Probably not needed anymore?

Same for the rest of the file

test/realtime/extensions/cdc_rls/cdc_rls_test.exs
```diff
@@ -33,7 +34,8 @@ defmodule Realtime.Extensions.CdcRlsTest do
         max_bytes_per_second: 100_000
       },
       is_new_api: false,
-      log_level: :info
+      log_level: :info,
+      region: Tenants.region(tenant)
```
edgurgel: Why is this change needed? 🤔

edgurgel (approved):
Added a few comments but LGTM

mix.exs
filipecabaco: no, the presence disable was set as minor vs patch 👀

test/realtime/api_test.exs
filipecabaco: yes my bad 😅

test/realtime/extensions/cdc_rls/cdc_rls_test.exs
filipecabaco: for nothing will remove

mix.exs
edgurgel: Oh I see 😆

PR #1443, opened 2025-07-01
fix: set idle_shutdown to all RateCounters to be 1 hour

edgurgel:

We were leaving them there forever

RateCounters and respective GenCounters are left as active processes forever never shutting them down regardless of activity

## What is the new behavior?

Shut them down after 1 hour without any increments.

## Additional context


chasers:
Thank you!

PR #1446, opened 2025-07-02
fix: Add more claims to InvalidJWT errors

filipecabaco:

To better debug issues with Invalid JWTs we're adding more claims to the metadata, namely exp and iss.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.59.5",
+      version: "2.39.6",
       elixir: "~> 1.17.3",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -51,7 +51,7 @@ config :tailwind,
 # Configures Elixir's Logger
 config :logger, :console,
   format: "$time $metadata[$level] $message\n",
-  metadata: [:request_id, :project, :external_id, :application_name, :sub, :error_code]
+  metadata: [:request_id, :project, :external_id, :application_name, :sub, :exp, :iss, :error_code]
```
edgurgel: Do you actually need this though? I think this is only necessary to always expose the metadata if set through `Logger.metadata`.

When it's sent explicitly logger will log any metadata I think. 

For example here we log "measurements" and it's not added here

https://github.com/supabase/realtime/blob/main/lib/realtime/telemetry/logger.ex#L33-L34

filipecabaco: you are right, this would only be needed for development where seeing this is not that relevant to be fair... removed it

PR #1452, opened 2025-07-09
fix: Add metadata to logs by default

filipecabaco:

Add metadata to logs by default

### DIFF
--- a/lib/realtime_web/channels/realtime_channel/logging.ex
+++ b/lib/realtime_web/channels/realtime_channel/logging.ex
@@ -9,45 +9,28 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
   @doc """
   Checks if the log level set in the socket is less than or equal to the given level of the message to be logged.
   """
-  @spec maybe_log(
-          socket :: Phoenix.Socket.t(),
-          level :: Logger.level(),
-          code :: binary(),
-          msg :: binary(),
-          metadata :: keyword()
-        ) :: :ok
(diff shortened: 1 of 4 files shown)


lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -9,45 +9,28 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
...
+  def maybe_log_warning(socket, code, msg), do: maybe_log(socket, :warning, code, msg)
 
-  def maybe_log_info(%{assigns: %{log_level: log_level}}, msg, metadata \\ []) do
-    metadata = if metadata == [], do: Logger.metadata()
-
-    if Logger.compare_levels(log_level, :info) != :gt do
-      Logger.info(inspect(msg), metadata)
-    end
+  def maybe_log_info(%{assigns: %{log_level: log_level, tenant: tenant}}, msg) do
+    Logger.metadata(external_id: tenant, project: tenant)
```
edgurgel, line 23: Same as the other PR. I think this should be set at when the process start like we do on `RealtimeChannel.join`. This will keep trying to update the process metadata which will just be a waste operation I think. We probably don't need to touch the metadata ?

filipecabaco: we do try that and then it ends up not being there as this error shows where we had the metadata set but does not show :/

edgurgel (approved):

PR #1453, opened 2025-07-09
fix: Add specific error for increased message rates

filipecabaco:

Add specific error for increased message rate errors but also refactored some logging approach to start simplifying that code


lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -58,25 +57,25 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
...
-  defp update_metadata_with_token_claims(metadata, token) do
+  defp update_metadata_with_token_claims(token) do
     case Joken.peek_claims(token) do
       {:ok, claims} ->
         sub = Map.get(claims, "sub")
         exp = Map.get(claims, "exp")
         iss = Map.get(claims, "iss")
-        metadata ++ [sub: sub, exp: exp, iss: iss]
+
+        Logger.metadata(sub: sub, exp: exp, iss: iss)
```
edgurgel: This sets the metadata at the process level. Do we really want that? 

From this point onwards this process will add this metadata to all the next logs even if unrelated to tokens.

I think this was correct

lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -43,13 +43,12 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
...
-          token :: Joken.bearer_token(),
-          metadata :: keyword()
+          socket :: Phoenix.Socket.t()
         ) :: {:error, %{reason: binary()}}
-  def log_error_with_token_metadata(code, msg, token, metadata \\ []) do
-    if metadata == [], do: Logger.metadata()
-    metadata = update_metadata_with_token_claims(metadata, token)
-    log_error_message(:error, code, msg, metadata)
+  def log_error_with_token_metadata(code, msg, %{assigns: %{access_token: access_token, tenant: tenant_id}} = socket) do
+    Logger.metadata(external_id: tenant_id, project: tenant_id)
```
edgurgel: `Logger.metadata` AFAIK updates the process dictionary to do this. Do we really need to do this at this stage?

It feels like it's enough to do this at the `RealtimeChannel.join` level as it will be set for the process from the point onwards. No need to set it for every log statement

filipecabaco: it's more on the side that when we try at the process level it seems that it does not "stick" plus we have this scattered across the code at the moment 😞 wanted to have a unified way to tackle this

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -370,7 +369,7 @@ defmodule RealtimeWeb.RealtimeChannel do
   def handle_in(_, _, %{assigns: %{rate_counter: %{avg: avg}, limits: %{max_events_per_second: max}}} = socket)
       when avg > max do
     message = "Too many messages per second"
-
+    Logging.log_error(socket, "MessagePerSecondRateLimitReached", message)
```
edgurgel: I'm not even sure this block of code is reached ATM. We can probably delete this block.

The rate counter is updated from a message generated by `RealtimeChannel.MessageDispatcher` [here](https://github.com/supabase/realtime/blob/6a0ee730ab6f7f2f3878900afceabfa798a676d6/lib/realtime_web/channels/realtime_channel/message_dispatcher.ex):

https://github.com/supabase/realtime/blob/6a0ee730ab6f7f2f3878900afceabfa798a676d6/lib/realtime_web/channels/realtime_channel.ex#L199-L211

We can probably update this block I mentioned above to NOT log now that the RateCounter will do the logging for us.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -592,10 +591,9 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   defp shutdown_response(socket, message) when is_binary(message) do
-    %{assigns: %{channel_name: channel_name, access_token: access_token}} = socket
-    metadata = log_metadata(access_token)
+    %{assigns: %{channel_name: channel_name}} = socket
     push_system_message("system", socket, "error", message, channel_name)
-    log_warning("ChannelShutdown", message, metadata)
+    Logging.maybe_log_warning(socket, "ChannelShutdown", message)
```
edgurgel, line 589: nice

test/integration/rt_channel_test.exs
```diff
@@ -1204,21 +1204,25 @@ defmodule Realtime.Integration.RtChannelTest do
...
+                             "response" => %{"reason" => "InvalidJWTToken: Token has expired 0 seconds ago"}
+                           },
+                           topic: ^realtime_topic
+                         },
+                         500
+        end)
 
       assert_receive %Message{event: "phx_close"}
+      assert log =~ "#{tenant.external_id}"
+      assert log =~ "#{tenant.external_id}"
```
edgurgel: 🤔

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: removing it and let's see what happens but you are right due to fastlane this is probably not touched

edgurgel (approved):

PR #1460, opened 2025-07-14
fix: measurements of broadcast from database lag

filipecabaco:

measurements of broadcast from database lag

### DIFF
--- a/lib/realtime/tenants/replication_connection.ex
+++ b/lib/realtime/tenants/replication_connection.ex
@@ -33,8 +33,6 @@ defmodule Realtime.Tenants.ReplicationConnection do
 
   @type t :: %__MODULE__{
           tenant_id: String.t(),
-          table: String.t(),
-          schema: String.t(),
           opts: Keyword.t(),
           step:
             :disconnected
@@ -54,8 +52,6 @@ defmodule Realtime.Tenants.ReplicationConnection do
           latency_committed_at: integer()
         }
   defstruct tenant_id: nil,
-            table: nil,
-            schema: "public",
             opts: [],
             step: :disconnected,
             publication_name: nil,
@@ -87,7 +83,8 @@ defmodule Realtime.Tenants.ReplicationConnection do
   end
 
   @default_init_timeout 30_000
-
+  @table "messages"
+  @schema "realtime"
   @doc """
   Starts the replication connection for a tenant and monitors a given pid to stop the ReplicationConnection.
   """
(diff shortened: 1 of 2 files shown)


lib/realtime/tenants/replication_connection.ex
```diff
@@ -298,51 +306,35 @@ defmodule Realtime.Tenants.ReplicationConnection do
...
-            case BatchBroadcast.broadcast(nil, tenant, %{messages: [to_broadcast]}, true) do
-              :ok -> :ok
-              error -> log_error("UnableToBatchBroadcastChanges", error)
-            end
-
-            {:noreply, state}
-        end
+    %{relations: relations, tenant_id: tenant_id, latency_committed_at: latency_committed_at} = state
+
+    with %{columns: columns} <- Map.get(relations, relation_id),
```
edgurgel, line 291: Image

payload, event and private are nullable. Should we allow them to not exist? Or should we update the table to not allow nullable and have default values?

I guess `private` and `event` really should be provided but payload could probably be an empty json? Not sure 🤔 

This might be fine to error out then. Just thought I would bring this up

lib/realtime/tenants/replication_connection.ex
```diff
@@ -280,6 +283,14 @@ defmodule Realtime.Tenants.ReplicationConnection do
   end
 
   @impl true
+  def handle_info(%Decoder.Messages.Begin{commit_timestamp: commit_timestamp}, state) do
+    latency_committed_at =
+      NaiveDateTime.utc_now()
+      |> NaiveDateTime.diff(commit_timestamp, :millisecond)
+
```
edgurgel: 👀

lib/realtime/tenants/replication_connection.ex
filipecabaco: honestly unsure 😅 we had them with `fetch!` so they would blow up anyway we just handle it now 🤔

edgurgel: Yeah I think it's better now 👍

lib/realtime/tenants/replication_connection.ex
```diff
@@ -378,4 +355,23 @@ defmodule Realtime.Tenants.ReplicationConnection do
...
+    |> Map.new(fn
+      {nil, %{name: name}} -> {name, nil}
+      {value, %{name: name, type: "jsonb"}} -> {name, Jason.decode!(value)}
+      {value, %{name: name, type: "bool"}} -> {name, value == "t"}
+      {value, %{name: name}} -> {name, value}
+    end)
+  end
+
+  defp get_or_error(map, key, error_type) do
+    case Map.get(map, key) do
```
edgurgel, line 367: I think it could simply be `Map.get(map, key, {:error, error_type})` 🤔

edgurgel: oh actually no the {:ok, part wouldn't be there

PR #1462, opened 2025-07-14
feat: use fastlane for broadcast events

edgurgel:

Introduces the fastlane approach that Phoenix supports. 

Fastlane essentially means that we send an encoded message straight to the Socket Transport process instead of going through the Channel process. It also serializes the message only once for each node/serializer instead of encoding the same payload for every channel process on every node.

We have 3 main types of events: presence, broadcast and postgres changes.

Postgres changes already uses fastlane with a custom MessageDispatcher:

https://github.com/supabase/realtime/blob/c5acf26de4d8114582d7303377db54a311549080/lib/extensions/postgres_cdc_rls/message_dispatcher.ex#L18

Presence is not managed by us so no fastlane there at the moment but can be done later with a custom adapter by the looks of it.

Broadcast is handled by us but it's not using fastlane.

## What is the new behavior?

Broadcast messages are now capable of using fastlane (both through Phoenix and `gen_rpc` for the distributed bit).

We still send 1 message per channel process so that the rate counter is updated and evaluated like before. This can be improved later with:

* Increment the counter once with the number of subscribers. 
* Change the rate counter limiter to happen at the broadcast level only not at the receiving end 🤔


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -833,4 +855,53 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+  @doc """
+  Inspired by Phoenix.Channel.Server.dispatch/3
+
+  """
+  def dispatch(subscribers, from, %Phoenix.Socket.Broadcast{} = msg) do
+    # fastlane_pid is the actual socket transport pid
+    # This reduce caches the serialization and bypasses the channel process going straight to the
+    # transport process
+
+    # Credo doesn't like that we don't use the result aggregation
```
filipecabaco: better to ignore the rule in credo itself maybe

filipecabaco (approved):

filipecabaco (approved):

PR #1463, opened 2025-07-15
fix: add channel tracking capabilities to sockets

filipecabaco:

* Adds a new Tracker process that:
  * uses ets to track how many channels each transport pid has
  * every X ms (setup via DISCONNECT_SOCKET_ON_NO_CHANNELS_INTERVAL_IN_MS) checks which pids have <0 counters and kills the associated pid
  * realtime channel increments on join
  * realtime channel decrements on terminate
* Moves some elements from rt_channel_test to other modules as we need to start think about ways of reducing the cognitive load of this file


lib/realtime_web/channels/user_socket.ex
```diff
@@ -33,22 +33,23 @@ defmodule RealtimeWeb.UserSocket do
...
-             jwt_jwks: jwt_jwks,
-             max_concurrent_users: max_conn_users,
-             max_events_per_second: max_events_per_second,
-             max_bytes_per_second: max_bytes_per_second,
-             max_joins_per_second: max_joins_per_second,
-             max_channels_per_client: max_channels_per_client,
-             postgres_cdc_default: postgres_cdc_default,
-             suspend: false
-           } <- Tenants.Cache.get_tenant_by_external_id(external_id),
+      with %Tenant{jwt_secret: jwt_secret, jwt_jwks: jwt_jwks, postgres_cdc_default: postgres_cdc_default} = tenant <-
```
edgurgel: `suspend: false` should be here I think? I think the test didn't break because I didn't assert on the log statement. 

What is probably happening here is that down on that match below when `suspend: true` it's crashing with a MatchError causing the socket to close but with an unexpected error.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -457,9 +459,10 @@ defmodule RealtimeWeb.RealtimeChannel do
   end
 
   @impl true
-  def terminate(reason, _state) do
+  def terminate(reason, %{transport_pid: transport_pid}) do
     Logger.debug("Channel terminated with reason: #{reason}")
     :telemetry.execute([:prom_ex, :plugin, :realtime, :disconnected], %{})
+    Tracker.untrack(transport_pid)
```
edgurgel, line 481: I don't think we have a guarantee that this will run if the RealtimeChannel process crashes 🤔

lib/realtime_web/channels/user_socket.ex
filipecabaco: damn yes, will tackle it

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: the only way would be for us to actively monitor every channel in the tracker process which will be bad 😅 

or am I thinking wrong?

lib/realtime_web/channels/user_socket.ex
filipecabaco: fixed and added an extra condition to the test to check for the suspend log

test/integration/rt_channel_test.exs
```diff
@@ -1586,10 +1586,15 @@ defmodule Realtime.Integration.RtChannelTest do
...
 
-      refute_receive _any
+      log =
+        capture_log(fn ->
+          {:ok, _} = Realtime.Api.update_tenant(tenant, %{suspend: true})
+          {:error, %Mint.WebSocket.UpgradeFailureError{}} = get_connection(tenant, "anon")
+          refute_receive _any
+        end)
+
+      assert log =~ "RealtimeDisabledForTenant"
```
edgurgel, line 1621: Thank you!

lib/realtime_web/channels/realtime_channel/tracker.ex
```diff
@@ -0,0 +1,86 @@
...
+  def init(opts) do
+    check_interval_in_ms = Keyword.fetch!(opts, :check_interval_in_ms)
+    Process.send_after(self(), :check_channels, check_interval_in_ms)
+    {:ok, %{check_interval_in_ms: check_interval_in_ms}}
+  end
+
+  @impl true
+  def handle_info(:check_channels, state) do
+    chunked_killing()
+    :ets.select_delete(@table, @zero_count_delete)
```
edgurgel, line 68: This is possible but not sure if we want to worry about this but between `chunked_killing` running and` select_delete` running we might miss some pids as the ets table might've been updated already and we removed rows that we didn't kill.

lib/realtime_web/channels/realtime_channel/tracker.ex
```diff
@@ -0,0 +1,86 @@
+defmodule RealtimeWeb.RealtimeChannel.Tracker do
```
edgurgel, line 1: Will this Tracker solve the issue where websockets are set-up and a channel is NEVER joined? Or will it won't detect websockets that joined and left getting to `0`?

lib/realtime_web/channels/realtime_channel/tracker.ex
filipecabaco: let's aim for best effort for now. we could actually remove right after killing 🤔 but let's see how it goes

lib/realtime_web/channels/realtime_channel/tracker.ex
filipecabaco: it won't 🤔 will move the track to UserSocket which will then allow us to do that

edgurgel: One thing that we need to ensure is that if a socket just started and say it took 5 seconds to join then it might get killed because we don't seem to track when the socket actually connected 🤔 

The ideal would be to ensure that the socket has no channels for X minutes or something? 🤔 

The more I think about it it feels like we need one process per socket to track this kind of stuff 😭

filipecabaco: still think that will be a huge bottleneck and this needs to be a best effort. this will tackle most use cases IMO 

this also allow us to set the time to delete which is ok for now.

started tracking as the first think of the RealtimeChannel which will handle most of the issues of when a channel fails to proceed

config/runtime.exs
```diff
@@ -48,6 +51,8 @@ janitor_run_after_in_ms = System.get_env("JANITOR_RUN_AFTER_IN_MS", "600000") |>
 janitor_children_timeout = System.get_env("JANITOR_CHILDREN_TIMEOUT", "5000") |> String.to_integer()
 # 4 hours by default
 janitor_schedule_timer = System.get_env("JANITOR_SCHEDULE_TIMER_IN_MS", "14400000") |> String.to_integer()
+# defaults to 30 seconds
+no_channel_timeout_in_ms = System.get_env("NO_CHANNEL_TIMEOUT_IN_MS", "600000") |> String.to_integer()
```
edgurgel: Is this being used? 🤔

config/runtime.exs
```diff
@@ -31,6 +31,9 @@ metrics_rpc_timeout_in_ms =
 rebalance_check_interval_in_ms =
   System.get_env("REBALANCE_CHECK_INTERVAL_IN_MS", to_string(:timer.minutes(10))) |> String.to_integer()
 
+disconnect_socket_on_no_channels_interval_in_ms =
```
edgurgel, line 34: Should this default to 10 minutes?

test/integration/rt_channel_test.exs
```diff
@@ -2129,32 +2109,85 @@ defmodule Realtime.Integration.RtChannelTest do
...
+      topic = "realtime:#{random_string()}"
+      :ok = WebsocketClient.join(socket, topic, %{config: config})
+      assert_receive %Message{topic: ^topic, event: "phx_reply", payload: %{"status" => "error"}}, 500
     end
+
+    assert [{_pid, count}] = Tracker.list_pids()
+    assert count == 0
+  end
+
+  test "failed connections but one succeeds properly tracks",
```
edgurgel, line 2181: Could we make the test check if the websocket connection is killed? I think if we start a Tracker with a low interval it should work?

lib/realtime_web/channels/realtime_channel/tracker.ex
```diff
@@ -0,0 +1,86 @@
...
+
+  defp chunked_killing(cont \\ nil) do
+    result = if cont, do: :ets.select(cont), else: :ets.select(@table, @zero_count_match, 1000)
+
+    case result do
+      :"$end_of_table" ->
+        :ok
+
+      {pids, cont} ->
+        Enum.each(pids, fn pid -> if Process.alive?(pid), do: Process.exit(pid, :kill) end)
```
edgurgel, line 82: Should we log how many processes we are killing here so we can observe it in action?

edgurgel: We could potentially later on find a way to send a "system" message before killing. Not for this PR though

PR #1464, opened 2025-07-16
fix: region rebalancing for Postgres Changes

edgurgel:

Follow-up from https://github.com/supabase/realtime/pull/1455  to also rebalance our Postgres Changes pool of processes to the correct region

If the Postgres Changes pool of processes start in the wrong region they might have a much higher latency that can impact performance. 

## What is the new behavior?

Every X time it will check if it's running at the correct region. If it's not then it stops. The next time it runs it will start at the right place.

## Additional context


filipecabaco (approved):

kimsoo527 (commented):
Incredible performance

PR #1465, opened 2025-07-17
Fix/gen rpc authorization

edgurgel:

Change Authorization calls to use `:gen_rpc` instead of `:erpc`

It uses `:erpc` for Authorization remote calls

## What is the new behavior?

`:gen_rpc` is used instead.

## Additional context

### DIFF
--- a/test/realtime/tenants/authorization_remote_test.exs
+++ b/test/realtime/tenants/authorization_remote_test.exs
@@ -93,36 +93,29 @@ defmodule Realtime.Tenants.AuthorizationRemoteTest do
 
       Process.sleep(100)
 
-      log =
-        capture_log(fn ->
-          t1 =
-            Task.async(fn ->
-              assert {:error, :increase_connection_pool} =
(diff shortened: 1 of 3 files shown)


test/realtime/tenants/authorization_remote_test.exs
```diff
@@ -93,36 +93,29 @@ defmodule Realtime.Tenants.AuthorizationRemoteTest do
...
+                   Authorization.get_read_authorizations(
+                     %Policies{},
+                     context.db_conn,
+                     context.authorization_context
+                   )
         end)
 
-      external_id = context.tenant.external_id
-      assert log =~ "project=#{external_id} external_id=#{external_id} [error] ErrorExecutingTransaction"
+      t2 =
```
edgurgel, line 106: For some reason during tests the logs from the other node are not getting redirected. I tested running 2 nodes and the logs work fine outside of tests.

filipecabaco (approved):

PR #1472, opened 2025-07-22
fix: realtime channel error cases

edgurgel:

Fix a few issues with Connect & Authorization usage

* Don't reset Authorization policies on `:confirm_token` as the actual token has not changed and only the expiration can be affected here.
* Avoid resetting Authorization policies twice on new "access_token"
* New access token was not being passed to `authorization_context` when a new access token was provided
* Capture Connect RPC errors that were not handled before

Added comments to the code to explain the changes

### DIFF
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -64,9 +64,8 @@ defmodule RealtimeWeb.RealtimeChannel do
          :ok <- limit_channels(socket),
          :ok <- limit_max_users(socket.assigns),
          {:ok, claims, confirm_token_ref, access_token, _} <- confirm_token(socket),
-         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id),
          socket = assign_authorization_context(socket, sub_topic, access_token, claims),
-         {:ok, socket} <- maybe_assign_policies(sub_topic, db_conn, socket) do
+         {:ok, socket} <- maybe_assign_policies(sub_topic, socket) do
       tenant_topic = Tenants.tenant_topic(tenant_id, sub_topic, !socket.assigns.private?)
 
       # fastlane subscription
@@ -424,20 +423,18 @@ defmodule RealtimeWeb.RealtimeChannel do
   def handle_in("access_token", %{"access_token" => refresh_token}, socket) when is_binary(refresh_token) do
     %{
       assigns: %{
-        access_token: access_token,
         pg_sub_ref: pg_sub_ref,
         channel_name: channel_name,
-        pg_change_params: pg_change_params,
-        tenant: tenant
+        pg_change_params: pg_change_params
       }
     } = socket
 
-    socket = assign(socket, :access_token, refresh_token)
+    # Update token and reset policies
+    socket = assign(socket, %{access_token: refresh_token, policies: nil})
 
     with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
-         socket = assign_authorization_context(socket, channel_name, access_token, claims),
-         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant),
-         {:ok, socket} <- maybe_assign_policies(channel_name, db_conn, socket) do
+         socket = assign_authorization_context(socket, channel_name, refresh_token, claims),
+         {:ok, socket} <- maybe_assign_policies(channel_name, socket) do
       Helpers.cancel_timer(pg_sub_ref)
       pg_change_params = Enum.map(pg_change_params, &Map.put(&1, :claims, claims))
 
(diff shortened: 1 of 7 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -62,9 +62,8 @@ defmodule RealtimeWeb.RealtimeChannel do
          :ok <- limit_max_users(socket.assigns),
          :ok <- start_db_rate_counter(tenant_id),
          {:ok, claims, confirm_token_ref, access_token, _} <- confirm_token(socket),
-         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id),
          socket = assign_authorization_context(socket, sub_topic, access_token, claims),
-         {:ok, socket} <- maybe_assign_policies(sub_topic, db_conn, socket) do
+         {:ok, socket} <- maybe_assign_policies(sub_topic, socket) do
```
edgurgel: Moved the `Connect` call to happen inside `maybe_assign_policies` only for private channels.

Public channels won't block here to connect.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -412,20 +411,18 @@ defmodule RealtimeWeb.RealtimeChannel do
...
 
-    socket = assign(socket, :access_token, refresh_token)
+    # Update token and reset policies
+    socket = assign(socket, %{access_token: refresh_token, policies: nil})
 
     with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
-         socket = assign_authorization_context(socket, channel_name, access_token, claims),
-         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant),
-         {:ok, socket} <- maybe_assign_policies(channel_name, db_conn, socket) do
+         socket = assign_authorization_context(socket, channel_name, refresh_token, claims),
```
edgurgel, line 438: We were using `access_token` and not `refresh_token` here. Wrong authorization context

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -412,20 +411,18 @@ defmodule RealtimeWeb.RealtimeChannel do
...
         channel_name: channel_name,
-        pg_change_params: pg_change_params,
-        tenant: tenant
+        pg_change_params: pg_change_params
       }
     } = socket
 
-    socket = assign(socket, :access_token, refresh_token)
+    # Update token and reset policies
+    socket = assign(socket, %{access_token: refresh_token, policies: nil})
```
edgurgel, line 435: Resetting the policies here.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -620,22 +617,14 @@ defmodule RealtimeWeb.RealtimeChannel do
...
 
   defp confirm_token(%{assigns: assigns} = socket) do
-    %{
-      jwt_secret: jwt_secret,
-      access_token: access_token,
-      tenant: tenant_id
-    } = assigns
+    %{jwt_secret: jwt_secret, access_token: access_token} = assigns
 
-    topic = Map.get(assigns, :topic)
```
edgurgel, line 604: This was always `nil` causing the call below on `maybe_assign_policies` to set it to `nil`. That's how we were resetting policies before.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -620,22 +617,14 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-    socket = Map.put(socket, :policies, nil)
     jwt_jwks = Map.get(assigns, :jwt_jwks)
 
     with jwt_secret_dec <- Crypto.decrypt!(jwt_secret),
          {:ok, %{"exp" => exp} = claims} when is_integer(exp) <-
            ChannelsAuthorization.authorize_conn(access_token, jwt_secret_dec, jwt_jwks),
-         exp_diff when exp_diff > 0 <- exp - Joken.current_time(),
-         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id),
-         {:ok, socket} <- maybe_assign_policies(topic, db_conn, socket) do
```
edgurgel, line 613: Stopped calculating authorization here as confirm token is just checking if the token is valid. 

We don't need to re-evaluate the policies at this stage. Just when we join and when the access token changes.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -412,20 +411,18 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        tenant: tenant
+        pg_change_params: pg_change_params
       }
     } = socket
 
-    socket = assign(socket, :access_token, refresh_token)
+    # Update token and reset policies
+    socket = assign(socket, %{access_token: refresh_token, policies: nil})
 
     with {:ok, claims, confirm_token_ref, _, socket} <- confirm_token(socket),
```
edgurgel, line 437: `confirm_token` was also calling `maybe_assign_policies` but we already call it in the next line. The confirm_token one was just resetting the policies due to the bug on `topic` assigns (see `confirm_token` comment)

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: doing this will create issues if the person is using broadcast from database and public channels

also this was being done to check if tenant db was healthy

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: this was done to recheck in case RLS policies change in between checks

filipecabaco: ideally we would "capture RLS changes" but this was the other way to handle it

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Fair point! Will keep the connect on join!

edgurgel: Done [here](https://github.com/supabase/realtime/pull/1472/commits/2f51df45798f652abad25296f1f3cdbc17623d08)

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Yeah but how often RLS policies change compared to access token refreshing? 🤔  

I think if we want to do this then yeah probably better to detect elsewhere and broadcast to all websockets to refresh it?

Like from time to time list the realtime.messages policy and when changed broadcast? Again I don't think policies change that often though

filipecabaco: depends because if the user is using anon directly the token actually has a 10 year expiration date at the moment so it won't happen that often but then you are right as those scenarios are not applicable here as they are not using private channels 🤔 ...

that tracking would be good but then it's another side process... but better than one flow per process doing it to be fair 😆

filipecabaco (approved):

PR #1487, opened 2025-08-06
fix: move rate limit message to happen inside the RateCounter

edgurgel:

Changes so that RateCounter can detect that a limit has been crossed so that we log once instead of whenever an event (in this case a RealtimeChannel join) happens.

I've also changed to only increment a join when it's not being rate limited. 

We can apply the same solution to other limits we currently have like limit of events, presence events, etc

We log every time a `RealtimeChannel` join has failed due to the rate limit of joins been reached

## What is the new behavior?

We log once when the limit it crossed and we won't log again until the average has decreased enough to be below the limit.

## Additional context

### DIFF
--- a/lib/realtime/rate_counter/rate_counter.ex
+++ b/lib/realtime/rate_counter/rate_counter.ex
@@ -34,6 +34,7 @@ defmodule Realtime.RateCounter do
             tick_ref: nil,
             idle_shutdown: @idle_shutdown,
             idle_shutdown_ref: nil,
+            limit: %{log: false},
             telemetry: %{
               event_name: [@app_name] ++ [:rate_counter],
               measurements: %{sum: 0},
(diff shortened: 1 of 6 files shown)


filipecabaco (approved):

lib/realtime/rate_counter/rate_counter.ex
```diff
@@ -49,6 +50,12 @@ defmodule Realtime.RateCounter do
           tick_ref: reference(),
           idle_shutdown: integer() | :infinity,
           idle_shutdown_ref: reference(),
+          limit: %{
+            log: boolean(),
+            value: integer(),
+            triggered: boolean(),
+            log_fn: (-> term())
```
filipecabaco, line 57: could this hurt us by being a function?

edgurgel: I don't think so? We could change to be `log: [:error, "message here", metadata]` and call Logger from inside. I think a function here is fine though.

PR #1493, opened 2025-08-11
fix: rate limit presence events

filipecabaco:

rate limit presence events


lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -23,30 +25,36 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
   def sync(%{assigns: %{private?: false}} = socket) do
-    %{assigns: %{tenant_topic: topic, channel_name: channel_name}} = socket
-    msg = "Syncing presence state for #{channel_name}"
-    Logging.maybe_log_info(socket, msg)
-    count(socket)
-    push(socket, "presence_state", presence_dirty_list(topic))
-    {:noreply, socket}
+    %{assigns: %{tenant_topic: topic}} = socket
+
+    Logging.maybe_log_info(socket, :sync_presence)
```
edgurgel: Should we only log if we `push` the presence state? That's what we are doing with a private channel below.

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -23,30 +25,36 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
+
+    with :ok <- limit_presence_event(socket) do
+      push(socket, "presence_state", presence_dirty_list(topic))
+      {:noreply, socket}
+    else
+      {:error, :rate_limit_exceeded} -> {:stop, :normal, socket}
```
edgurgel: Should we push a system message in this case? We do this on "Too many messages per second". Maybe pushing this back to RealtimeChannel and letting it handle it?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -23,30 +25,36 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
+      {:noreply, socket}
+    else
+      %Policies{presence: %PresencePolicies{read: false}} ->
+        Logger.info("Presence track message ignored on #{topic}")
+        {:noreply, socket}
+
+      _ ->
+        Logging.maybe_log_info(socket, :sync_presence)
+        push(socket, "presence_state", presence_dirty_list(topic))
+        socket
```
edgurgel: Which else case are we expecting here that should still publish if the `read: true` case should be the only place where we are allowed to push for a private channel? 🤔

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -92,11 +92,11 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
 
       {:error, :rls_policy_error, error} ->
         log_error("RlsPolicyError", error)
-        {:error, socket}
+        {:reply, :error, socket}
```
edgurgel: Should we let RealtimeChannel handle this by sending a proper system message with a shutdown response etc? 

https://github.com/supabase/realtime/blob/96866172c1556ba020ad92fe8f23c864b9181a87/lib/realtime_web/channels/realtime_channel.ex#L354-L364

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -115,41 +115,43 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
        ) do
-    {:error, socket}
+    {:reply, :error, socket}
   end
 
   defp handle_presence_event("untrack", _, _, socket) do
     %{assigns: %{presence_key: presence_key, tenant_topic: tenant_topic}} = socket
-    {Presence.untrack(self(), tenant_topic, presence_key), socket}
+    :ok = Presence.untrack(self(), tenant_topic, presence_key)
+    {:reply, :ok, socket}
```
edgurgel: What will this change on the websocket level?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -158,4 +160,20 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
     |> Shard.dirty_list(topic)
     |> Phoenix.Presence.group()
   end
+
+  defp limit_presence_event(socket) do
+    %{assigns: %{presence_rate_counter: presence_counter, tenant: tenant_id}} = socket
+    GenCounter.add(presence_counter.id)
```
edgurgel: Should we only count when the limit has not been reached (inside the `else`)? This way we don't count if it's already not going to progress?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -115,41 +115,43 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
       {:error, {:already_tracked, pid, _, _}} ->
         case Presence.update(pid, tenant_topic, presence_key, payload) do
-          {:ok, _} -> {:ok, socket}
-          {:error, _} -> {:error, socket}
+          {:ok, _} -> {:reply, :ok, socket}
+          {:error, _} -> {:reply, :error, socket}
         end
 
+      {:error, :rate_limit_exceeded} ->
+        {:reply, :error, socket}
```
edgurgel: Why replying error here and above we are doing a stop on line 36? 🤔

lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -18,6 +25,17 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
     {:error, %{reason: msg}}
   end
 
+  @doc """
+  Logs an error message
```
edgurgel:
```suggestion
  Logs a warning message
```

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -10,80 +10,72 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
   alias RealtimeWeb.Presence
   alias RealtimeWeb.RealtimeChannel.Logging
 
+  defguard is_private?(socket) when socket.assigns.private?
+
+  defguard can_read_presence?(socket) when is_private?(socket) and socket.assigns.policies.presence.read
+
+  defguard can_write_presence?(socket)
+           when is_private?(socket) and
+                  (is_nil(socket.assigns.policies.presence.write) or socket.assigns.policies.presence.write)
```
edgurgel: why is this different from can_read_presence?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -10,80 +10,72 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
-  def sync(%{assigns: %{presence_enabled?: false}} = socket), do: {:noreply, socket}
-
-  def sync(%{assigns: %{private?: false}} = socket) do
-    %{assigns: %{tenant_topic: topic, channel_name: channel_name}} = socket
-    msg = "Syncing presence state for #{channel_name}"
-    Logging.maybe_log_info(socket, msg)
-    count(socket)
-    push(socket, "presence_state", presence_dirty_list(topic))
-    {:noreply, socket}
+  @spec sync(Socket.t()) :: {:ok, Socket.t()} | {:error, :rls_policy_error | :unable_to_set_policies}
```
edgurgel: Should we add the reason `:rate_limit_exceeded`

lib/realtime_web/channels/realtime_channel/presence_handler.ex
edgurgel: Any reason to count even if it's already rate limited? 🤔

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -10,80 +10,72 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
-  def handle(_, %{assigns: %{presence_enabled?: false}} = socket), do: {:reply, :ok, socket}
-  def handle(payload, %{assigns: %{private?: false}} = socket), do: handle(payload, nil, socket)
+  @spec handle(map(), Socket.t()) :: {:ok, Socket.t()} | {:error, :unauthorized} | {:error, :rate_limit_exceeded}
+  def handle(_, %{assigns: %{presence_enabled?: false}} = socket), do: {:ok, socket}
+  def handle(payload, socket) when not is_private?(socket), do: handle(payload, nil, socket)
 
-  @spec handle(map(), pid() | nil, Socket.t()) :: {:reply, :error | :ok, Socket.t()}
-  def handle(_, _, %{assigns: %{presence_enabled?: false}} = socket), do: {:reply, :ok, socket}
+  @spec handle(map(), pid() | nil, Socket.t()) ::
+          {:ok, Socket.t()} | {:error, :unauthorized} | {:error, :rate_limit_exceeded}
```
edgurgel: missing `:unable_to_track_presence` reason I think

edgurgel: oh and `rls_policy_error` and `unable_to_set_policies`

test/integration/rt_channel_test.exs
```diff
@@ -706,7 +706,6 @@ defmodule Realtime.Integration.RtChannelTest do
 
       WebsocketClient.send_event(socket, topic, "presence", payload)
 
-      assert_receive %Phoenix.Socket.Message{topic: ^topic, event: "phx_reply", payload: %{"status" => "ok"}}
```
edgurgel, line 709: Do you know why this has changed?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
edgurgel: It doesn't look like we change the socket. Should we return `:ok` then for the non-error case?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -353,17 +347,30 @@ defmodule RealtimeWeb.RealtimeChannel do
   def handle_in("presence", payload, %{assigns: %{private?: true}} = socket) do
     %{tenant: tenant_id} = socket.assigns
 
-    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id) do
-      PresenceHandler.handle(payload, db_conn, socket)
+    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant_id),
+         {:ok, socket} <- PresenceHandler.handle(payload, db_conn, socket) do
+      {:noreply, socket}
```
edgurgel: So do we want to change `track` to not reply ok here? Can this impact a client code that expects this? 🤔

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -353,17 +347,30 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-        log_error("UnableToHandlePresence", error)
+        log_error(socket, "UnableToHandlePresence", error)
         {:noreply, socket}
     end
   end
 
   def handle_in("presence", payload, %{assigns: %{private?: false}} = socket) do
-    PresenceHandler.handle(payload, socket)
+    with {:ok, socket} <- PresenceHandler.handle(payload, socket) do
+      {:noreply, socket}
```
edgurgel: Same question as above

lib/realtime_web/channels/realtime_channel/presence_handler.ex
filipecabaco: moving it to the `else` of the check

lib/realtime_web/channels/realtime_channel/presence_handler.ex
filipecabaco: can read presence checks read policy, this checks write policy

lib/realtime_web/channels/realtime_channel/presence_handler.ex
filipecabaco: can do

test/integration/rt_channel_test.exs
filipecabaco: this should not be here as this could be an error actually as this event would mean it reconnected 👀

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: good catch, it would indeed break as this is replying to an `acked` event from the client lib

lib/realtime_web/channels/realtime_channel/presence_handler.ex
edgurgel: Oh sorry but why `is_nil` is allowed here?

edgurgel: Oh I see we call Auth down there. 

nitpick: I wonder if we should leave here without the nil check and do the `nil` check when needed?

```elixir
defguard can_write_presence?(socket) when is_private?(socket) and socket.assigns.policies.presence.write)
```

Will add a comment where this is used ⏬

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -10,80 +10,75 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
 
-  defp handle_presence_event(
-         "track",
-         payload,
-         db_conn,
-         %{assigns: %{private?: true, policies: %Policies{presence: %PresencePolicies{write: nil}} = policies}} = socket
-       ) do
-    %{assigns: %{authorization_context: authorization_context}} = socket
+  defp handle_presence_event("track", payload, db_conn, socket)
+       when can_write_presence?(socket) and is_nil(socket.assigns.policies.presence.write) do
```
edgurgel: nit: Maybe here it could be only `when is_nil(socket.assigns.policies.presence.write)` which means we still need to calculate just for this case?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -10,80 +10,75 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
...
-
-  def sync(%{assigns: %{private?: false}} = socket) do
-    %{assigns: %{tenant_topic: topic, channel_name: channel_name}} = socket
-    msg = "Syncing presence state for #{channel_name}"
-    Logging.maybe_log_info(socket, msg)
-    count(socket)
-    push(socket, "presence_state", presence_dirty_list(topic))
-    {:noreply, socket}
+  @spec sync(Socket.t()) ::
+          :ok | {:error, :rls_policy_error | :unable_to_set_policies | :rate_limit_exceeded}
```
edgurgel: can this function return rls_policy_error or unable_to_set_policies?

edgurgel (approved):
Added a few comments but overall LGTM. Nice one

lib/realtime_web/channels/realtime_channel/presence_handler.ex
filipecabaco: no! copy pasta strikes again

lib/realtime_web/channels/realtime_channel/presence_handler.ex
filipecabaco: did this change as it also simplifies a lot the guard of the function

PR #1501, opened 2025-08-21
fix: improve join payload handling

filipecabaco:

improves error handling on join by doing basic checks on types received by realtime. it's the scafold for future validations with more informative errors to the users


lib/realtime_web/channels/realtime_channel/payloads/join.ex
```diff
@@ -0,0 +1,140 @@
...
+    field :enabled, :boolean, default: true
+    field :key, :string, default: UUID.uuid1()
+  end
+
+  def changeset(presence, attrs) do
+    cast(presence, attrs, [:enabled, :key], message: &Join.error_message/2)
+  end
+end
+
+defmodule RealtimeWeb.Channels.RealtimeChannel.Payloads.Join.Config do
```
edgurgel: We don't seem to use `Channels` as a namespace anywhere. Also it would be nice to have a shorter module name 😆 

RealtimeWeb.RealtimeChannel.Join
RealtimeWeb.RealtimeChannel.Join.Config
RealtimeWeb.RealtimeChannel.Join.Config.Presence 

Removing Channels and Payloads maybe?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -531,40 +546,24 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-       when is_binary(user_token) do
-    assign(socket, :access_token, user_token)
-  end
+    access_token =
+      cond do
+        access_token != nil and !String.starts_with?(access_token, "sb_") -> access_token
+        user_token != nil and !String.starts_with?(user_token, "sb_") -> user_token
+        tenant_token != nil and !String.starts_with?(tenant_token, "sb_") -> tenant_token
+        true -> header
+      end
```
edgurgel: nitpick: Do we need to check if it's a string as well? Our previous code was enforcing it with `when is_binary`

I think we can simply swap the "!= nil" for `is_binary(...)` ?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -631,28 +630,30 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-          }
-        end)
-
-      _ ->
-        []
-    end
+  defp pg_change_params(true, %Join{config: %Config{postgres_changes: postgres_changes}}, channel_pid, claims, _)
+       when not is_nil(postgres_changes) do
+    postgres_changes
+    |> Enum.reject(&is_nil/1)
```
edgurgel: Should this even be possible? Having a `nil` as part of `postgres_changes`? What happens in this case with the new code? 

```
 "postgres_changes": [
        {
          "event": "*",
          "schema": "public",
          "table": "*"
        },
       null
      ],
```

test/realtime_web/channels/realtime_channel/payloads/join_test.exs
```diff
@@ -0,0 +1,90 @@
...
+  describe "validate/1" do
+    test "valid payload allows join" do
+      key = random_string()
+      access_token = random_string()
+
+      config = %{
+        "config" => %{
+          "private" => false,
+          "broadcast" => %{"ack" => false, "self" => false},
+          "presence" => %{"enabled" => true, "key" => key},
```
edgurgel: Should we add a test when `presence->enabled` is not provided?

lib/realtime_web/channels/realtime_channel/payloads/join.ex
```diff
@@ -0,0 +1,140 @@
...
+        errors = Ecto.Changeset.traverse_errors(changeset, &elem(&1, 0))
+        {:error, :invalid_join_payload, errors}
+    end
+  end
+
+  def presence_enabled?(%__MODULE__{config: %Config{presence: %Presence{enabled: enabled}}}), do: enabled
+  def presence_enabled?(_), do: true
+
+  def presence_key(%__MODULE__{config: %Config{presence: %Presence{key: key}}}), do: key
+  def presence_key(_), do: UUID.uuid1()
```
edgurgel: Do we need to account for empty string on `key` ? I think realtime-js sends this today:

```
      "presence": {
        "key": ""
      },
```

It might work already but maybe we need a test?

phx_join.schema.json
```diff
@@ -0,0 +1,139 @@
+{
```
edgurgel, line 1: Are we going to use this for something?

lib/realtime_web/channels/realtime_channel/payloads/join.ex
```diff
@@ -0,0 +1,140 @@
...
+    embeds_one :presence, Presence
+    embeds_many :postgres_changes, PostgresChange
+    field :private, :boolean, default: false
+  end
+
+  def changeset(config, attrs) do
+    config
+    |> cast(attrs, [:private], message: &Join.error_message/2)
+    |> cast_embed(:broadcast, invalid_message: "unable to parse, expected a map")
+    |> cast_embed(:presence, invalid_message: "unable to parse, expected a map ")
```
edgurgel: supernit: Extra space at the end of string 😆

edgurgel (commented):
Nice improvement!

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: avoided to change this logic honestly, kept what we had just "cleaner"

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: good call will change

lib/realtime_web/channels/realtime_channel/payloads/join.ex
filipecabaco: will do 👍 and will break into multiple files. wanted to keep it closer to the folder structure

test/realtime_web/channels/realtime_channel/payloads/join_test.exs
filipecabaco: can do 👍

lib/realtime_web/channels/realtime_channel/payloads/join.ex
filipecabaco: good call will do

phx_join.schema.json
filipecabaco: will post it on docs for client libs, other payloads will follow so we can let users validate payloads in client libs

lib/realtime_web/channels/realtime_channel/payloads/join.ex
filipecabaco: eagle eye!

edgurgel (approved):

test/realtime_web/channels/payloads/join_test.exs
```diff
@@ -0,0 +1,96 @@
+defmodule RealtimeWeb.Channels.Payloads.JoinTest do
+  use ExUnit.Case
```
edgurgel:
```suggestion
  use ExUnit.Case, async: true
```

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -54,6 +55,11 @@ defmodule RealtimeWeb.RealtimeChannel do
       |> assign(:private?, !!params["config"]["private"])
       |> assign(:policies, nil)
 
+    case Join.validate(params) do
+      {:ok, _join} -> nil
+      {:error, :invalid_join_payload, errors} -> log_error(socket, "InvalidJoinPayload", errors)
```
edgurgel: Should it also log the params? Otherwise we will see errors but we won't see the original params that caused the errors I think?

filipecabaco: logging them out with redacted values for user_token and access_token

PR #1505, opened 2025-08-26
feat: rate counter for failing auth due to PG connection errors

edgurgel:

* Change RateCounter to allow for not only average (`avg`) but `sum` limits
* Change Authorization to rate limit based on the amount of Connection errors we get when calling verifying channel authorisation

The approach I took was: once it reaches the amount of errors equal to the connection pool size in the last 30 seconds then rate limit until time has passed enough so that the sum of errors is below the connection pool size.

Connection pool is 5, if in the last 30 seconds 5 errors happened, the immediate next Auth call will fail instantly. We could think about adding a `Process.sleep` here to avoid a loop of retries. In any case the channel join limit will kick in.

The message I chose was:

```
IncreaseConnectionPool: Too many database timeouts
```

## Additional context


README.md
```diff
@@ -232,8 +232,8 @@ This is the list of operational codes that can help you understand your deployme
...
 | ConnectionRateLimitReached         | The number of connected clients as reached its limit                                                                                                                                                  |
-| ClientJoinRateLimitReached         | The rate of joins per second from your clients as reached the channel limits                                                                                                                          |
-| MessagePerSecondRateLimitReached   | The rate of messages per second from your clients as reached the channel limits                                                                                                                       |
+| ClientJoinRateLimitReached         | The rate of joins per second from your clients sas reached the channel limits                                                                                                                         |
+| MessagePerSecondRateLimitReached   | The rate of messages per second from your clients sas reached the channel limits                                                                                                                      |
```
filipecabaco, line 236:
```suggestion
| MessagePerSecondRateLimitReached   | The rate of messages per second from your clients as reached the channel limits                                                                                                                      |
```

README.md
```diff
@@ -232,8 +232,8 @@ This is the list of operational codes that can help you understand your deployme
...
 | ChannelRateLimitReached            | The number of channels you can create has reached its limit                                                                                                                                           |
 | ConnectionRateLimitReached         | The number of connected clients as reached its limit                                                                                                                                                  |
-| ClientJoinRateLimitReached         | The rate of joins per second from your clients as reached the channel limits                                                                                                                          |
-| MessagePerSecondRateLimitReached   | The rate of messages per second from your clients as reached the channel limits                                                                                                                       |
+| ClientJoinRateLimitReached         | The rate of joins per second from your clients sas reached the channel limits                                                                                                                         |
```
filipecabaco, line 235:
```suggestion
| ClientJoinRateLimitReached         | The rate of joins per second from your clients as reached the channel limits                                                                                                                         |
```

lib/realtime/tenants/authorization.ex
```diff
@@ -91,11 +108,13 @@ defmodule Realtime.Tenants.Authorization do
...
-    case get_write_policies_for_connection(db_conn, authorization_context, policies) do
-      {:ok, %Policies{} = policies} -> {:ok, policies}
-      {:ok, {:error, %Postgrex.Error{} = error}} -> {:error, :rls_policy_error, error}
-      {:error, %ConnectionError{reason: :queue_timeout}} -> {:error, :increase_connection_pool}
-      {:error, error} -> {:error, error}
+    rate_counter = rate_counter(authorization_context.tenant_id)
+
+    if !rate_counter.limit.triggered do
+      get_write_policies_for_connection(db_conn, authorization_context, policies)
+      |> handle_policies_result(rate_counter)
```
filipecabaco, line 115:
```suggestion
      db_conn
      |> get_write_policies_for_connection( authorization_context, policies)
      |> handle_policies_result(rate_counter)
```

test/realtime/database_test.exs
```diff
@@ -274,7 +274,7 @@ defmodule Realtime.DatabaseTest do
              end) =~ "project=123 external_id=123 [error] ErrorExecutingTransaction:"
     end
 
-    test "noproc error" do
+    test "noproc" do
```
filipecabaco, line 277: it would be better to something more descriptive

README.md
edgurgel: I swear I had written "has" no sas 😆

edgurgel:
Sorry i opened the PR with the wrong branch 🤦 

https://github.com/supabase/realtime/pull/1508

test/realtime/database_test.exs
edgurgel: Fixed in the right PR!

lib/realtime/tenants/authorization.ex
edgurgel: Fixed in the right PR!

PR #1506, opened 2025-08-26
fix: improve deployment to be more platform agnostic

filipecabaco:

currently we have fly and aws very hardcoded so we need to reduce that by changing the way we setup realtime to be less locked to this ecosystems

the first we are focusing on read replicas which is very hard coded to our infrastructure.

for read replicas you can set:
```
export DB_HOST_REPLICA_HOST_us-east-1='127.0.0.1'
export DB_HOST_REPLICA_TARGET_REGIONS_us-east-1='us-east-1,us-east-2,us-east-3'
```
which will setup the `us-east-1` read replica and inform realtime that the regions `us-east-1,us-east-2,us-east-3` should use this read replica. if not, they will use the default repo.


config/runtime.exs
```diff
@@ -11,62 +25,44 @@ database = System.get_env("DB_NAME", "postgres")
...
-  System.get_env("METRICS_RPC_TIMEOUT_IN_MS", "15000") |> String.to_integer()
-
-rebalance_check_interval_in_ms =
-  System.get_env("REBALANCE_CHECK_INTERVAL_IN_MS", to_string(:timer.minutes(10))) |> String.to_integer()
+tenant_cache_expiration = Env.get_integer("TENANT_CACHE_EXPIRATION_IN_MS", :timer.seconds(30))
+migration_partition_slots = Env.get_integer("MIGRATION_PARTITION_SLOTS", System.schedulers_online() * 2)
+connect_partition_slots = Env.get_integer("CONNECT_PARTITION_SLOTS", System.schedulers_online() * 2)
+metrics_cleaner_schedule_timer_in_ms = Env.get_integer("METRICS_CLEANER_SCHEDULE_TIMER_IN_MS", :timer.minutes(30))
+metrics_rpc_timeout_in_ms = Env.get_integer("METRICS_RPC_TIMEOUT_IN_MS", :timer.seconds(15))
+rebalance_check_interval_in_ms = Env.get_integer("REBALANCE_CHECK_INTERVAL_IN_MS", :timer.seconds(10))
```
edgurgel, line 43: this one should be `:timer.minutes(10)`

edgurgel, line 43:
```suggestion
rebalance_check_interval_in_ms = Env.get_integer("REBALANCE_CHECK_INTERVAL_IN_MS", :timer.minutes(10))
```

config/runtime.exs
```diff
@@ -11,62 +25,44 @@ database = System.get_env("DB_NAME", "postgres")
...
-janitor_max_children = System.get_env("JANITOR_MAX_CHILDREN", "5") |> String.to_integer()
-janitor_chunk_size = System.get_env("JANITOR_CHUNK_SIZE", "10") |> String.to_integer()
-# defaults to 10 minutes
-janitor_run_after_in_ms = System.get_env("JANITOR_RUN_AFTER_IN_MS", "600000") |> String.to_integer()
-# defaults to 5 seconds
-janitor_children_timeout = System.get_env("JANITOR_CHILDREN_TIMEOUT", "5000") |> String.to_integer()
-# 4 hours by default
-janitor_schedule_timer = System.get_env("JANITOR_SCHEDULE_TIMER_IN_MS", "14400000") |> String.to_integer()
-# defaults to 10 minutes
+  Env.get_integer("DISCONNECT_SOCKET_ON_NO_CHANNELS_INTERVAL_IN_MS", :timer.seconds(30))
```
edgurgel, line 60: I don't think this `disconnect_socket_on_no_channels_interval_in_ms` is being used at all. Should we just remove it?

config/runtime.exs
```diff
@@ -103,58 +99,91 @@ config :realtime,
   max_gen_rpc_clients: max_gen_rpc_clients,
   no_channel_timeout_in_ms: no_channel_timeout_in_ms
 
-if config_env() == :test || !run_janitor? do
-  config :realtime, run_janitor: false
-else
+if config_env() != :test && run_janitor? do
   config :realtime,
```
edgurgel, line 103: Where we are setting `run_janitor: false` when `run_janitor?` is `false`?

config/runtime.exs
```diff
@@ -191,12 +218,39 @@ if config_env() != :test do
...
+  config :libcluster, debug: false, topologies: cluster_topologies
+  secret_key_base = System.fetch_env!("SECRET_KEY_BASE")
+  if app_name == "", do: raise("APP_NAME not available")
+
+  config :realtime, RealtimeWeb.Endpoint,
+    server: true,
+    url: [host: "#{app_name}.supabase.co", port: 443],
+    http: [
+      compress: true,
+      port: Env.get_integer("PORT", 443),
```
edgurgel, line 237: port was `4000` before here

config/runtime.exs
```diff
@@ -191,12 +218,39 @@ if config_env() != :test do
...
+# Setup Production
+
+if config_env() == :prod do
+  config :libcluster, debug: false, topologies: cluster_topologies
+  secret_key_base = System.fetch_env!("SECRET_KEY_BASE")
+  if app_name == "", do: raise("APP_NAME not available")
+
+  config :realtime, RealtimeWeb.Endpoint,
+    server: true,
+    url: [host: "#{app_name}.supabase.co", port: 443],
```
edgurgel, line 234: it was `80` here

README.md
```diff
@@ -184,8 +184,21 @@ If you're using the default tenant, the URL is `ws://realtime-dev.localhost:4000
...
+### Read Replica setup
+
+Realtime supports setting up read replicas per region using environment variables. For example, if you want to set up a read replica for the `us-east-1` region, you can set the following environment variables:
+
+```bash
+export DB_HOST_REPLICA_HOST_us-east-1='127.0.0.1'
+export DB_HOST_REPLICA_TARGET_REGIONS_us-east-1='us-east-1,us-east-2,us-east-3'
+```
+
+This will inform realtime that nodes connected to the `us-east-1` region will be able to connect to the replica database for the `us-east-1` region. If the region is not specified, the we will set the main repository as the default.
```
edgurgel, line 201: how us-east-1 and us-east-2 etc are used in this example

```
If the region is not specified, the we will set the main repository as the default.
```

*then ?

lib/realtime/nodes.ex
```diff
@@ -27,11 +27,6 @@ defmodule Realtime.Nodes do
   def platform_region_translator(nil), do: nil
 
   def platform_region_translator(tenant_region) when is_binary(tenant_region) do
-    platform = Application.get_env(:realtime, :platform)
-    region_mapping(platform, tenant_region)
-  end
-
-  defp region_mapping(:aws, tenant_region) do
```
edgurgel, line 34: Should this be configured through the ENV vars as well? (Not in this PR but in the future?)

filipecabaco: yes it will be just wanted to make it in a different PR but it will be similar to the existing one 👀

filipecabaco:
https://github.com/supabase/realtime/pull/1511

PR #1516, opened 2025-09-01
feat: add support for generating certs for securing inter-node comms

darora:
This is anticipated to be a temporary feature that will be deprecated
and removed once we move to a platform that offers easy init container
support.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.45.0",
+      version: "2.46.0",
       elixir: "~> 1.17.3",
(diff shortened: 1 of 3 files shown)


mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.45.0",
+      version: "2.45.1",
```
edgurgel: Given that it's a `feat:` this should be 2.46.0

edgurgel (approved):
Added a comment on the version. Other than that LGTM

mix.exs
darora: fair, addressed

PR #1526, opened 2025-09-08
feat: replay realtime.messages

edgurgel:

Introduce a way to load `realtime.messages` when joining a channel.

```typescript
const replay = { limit: 10, since: 1 }
const channel = client.channel('my-channel', { config: { broadcast: { ack: true, replay: replay } } })
```

A new index was created on `inserted_at DESC, topic, private` 

The hardcoded limit is 25 for now.

## Additional context


edgurgel:
> we should also use this opportunity to change granularity of inserted_at otherwise let's go!

Yup can do (in a separate PR)!

PR #1528, opened 2025-09-09
fix: handle wal bloat

filipecabaco:

Verify that replication connection is able to reconnect when faced with WAL bloat issues

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -327,24 +327,28 @@ defmodule Realtime.Tenants.Connect do
     {:stop, :shutdown, state}
   end
 
+  @replication_recovery_backoff 1000
+
   # Handle replication connection termination
   def handle_info(
         {:DOWN, replication_connection_reference, _, _, _},
         %{replication_connection_reference: replication_connection_reference} = state
       ) do
-    Logger.warning("Replication connection has died")
-    Process.send_after(self(), :recover_replication_connection, 100)
+    log_warning("ReplicationConnectionDown", "Replication connection has been terminated")
+    Process.send_after(self(), :recover_replication_connection, @replication_recovery_backoff)
+    state = %{state | replication_connection_pid: nil, replication_connection_reference: nil}
     {:noreply, state}
   end
 
-  @replication_connection_query "SELECT * from pg_stat_activity where application_name='realtime_replication_connection'"
+  @replication_connection_query "SELECT 1 from pg_stat_activity where application_name='realtime_replication_connection'"
   def handle_info(:recover_replication_connection, state) do
     with %{num_rows: 0} <- Postgrex.query!(state.db_conn_pid, @replication_connection_query, []),
          {:ok, state} <- start_replication_connection(state) do
       {:noreply, state}
     else
       _ ->
-        Process.send_after(self(), :recover_replication_connection, 100)
+        log_error("ReplicationConnectionRecoveryFailed", "Replication connection recovery failed")
+        Process.send_after(self(), :recover_replication_connection, @replication_recovery_backoff)
         {:noreply, state}
     end
   end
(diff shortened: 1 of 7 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -354,7 +333,20 @@ defmodule Realtime.Tenants.Connect do
         %{replication_connection_reference: replication_connection_reference} = state
       ) do
     Logger.warning("Replication connection has died")
-    {:stop, :shutdown, state}
+    Process.send_after(self(), :recover_replication_connection, 100)
+    {:noreply, state}
+  end
+
+  @replication_connection_query "SELECT * from pg_stat_activity where application_name='realtime_replication_connection'"
```
edgurgel: nitpick: `SELECT 1 FROM pg_stat_activity WHERE application_name='realtime_replication_connection'`  ? 

Given that we don't care about the result

lib/realtime/tenants/connect.ex
```diff
@@ -354,7 +333,20 @@ defmodule Realtime.Tenants.Connect do
         %{replication_connection_reference: replication_connection_reference} = state
       ) do
     Logger.warning("Replication connection has died")
-    {:stop, :shutdown, state}
+    Process.send_after(self(), :recover_replication_connection, 100)
```
edgurgel: Should we turn this `100` into a module attribute? Any reason for 100 and not say 1000 or 500?

lib/realtime/tenants/connect.ex
```diff
@@ -354,7 +333,20 @@ defmodule Realtime.Tenants.Connect do
...
+  def handle_info(:recover_replication_connection, state) do
+    with %{num_rows: 0} <- Postgrex.query!(state.db_conn_pid, @replication_connection_query, []),
+         {:ok, state} <- start_replication_connection(state) do
+      {:noreply, state}
+    else
+      _ ->
+        Process.send_after(self(), :recover_replication_connection, 100)
+        {:noreply, state}
+    end
   end
```
edgurgel: Should we log something when we trying to reconnect?

lib/realtime/tenants/connect.ex
```diff
@@ -354,7 +333,20 @@ defmodule Realtime.Tenants.Connect do
         %{replication_connection_reference: replication_connection_reference} = state
       ) do
     Logger.warning("Replication connection has died")
-    {:stop, :shutdown, state}
+    Process.send_after(self(), :recover_replication_connection, 100)
+    {:noreply, state}
```
edgurgel, line 340: Should we set the `replication_connection_reference` and `replication_connection_pid` to `nil` here?

filipecabaco: good call changed

lib/realtime/tenants/connect.ex
filipecabaco: yes added it 👍

lib/realtime/tenants/connect.ex
filipecabaco: no particular reason, made it 1000 to prevent overloading the db

lib/realtime/tenants/connect.ex
filipecabaco: good call changed it

lib/realtime/tenants/replication_connection.ex
```diff
@@ -144,8 +144,8 @@ defmodule Realtime.Tenants.ReplicationConnection do
         port: connection_opts.port,
         socket_options: connection_opts.socket_options,
         ssl: connection_opts.ssl,
-        backoff_type: :stop,
```
edgurgel, line 147: do we still want this change? What does it do?

edgurgel (approved):

lib/realtime/tenants/replication_connection.ex
filipecabaco: nothing 😆 this was a thing we thought we needed but this is for DBConnection so it's useless for Postgrex.ReplicationConnection

PR #1533, opened 2025-09-15
fix: move DB setup to happen after Connect.init

edgurgel:

This change reduces the impact of slow DB setup impacting other tenants trying to connect at the same time that landed on the same partition.

I've also removed `StartCounters` as it's not needed anymore. We start the RateCounters where they are handled. 

`Connect.init/1` can take minutes blocking other `Connect` processes from starting up if they share the same partition supervisor.

## What is the new behavior?

Change `Connect.init/1` to be quick and delay connecting to `handle_continue`

## Additional context


lib/realtime/tenants/connect/check_connection.ex
```diff
@@ -2,16 +2,14 @@ defmodule Realtime.Tenants.Connect.CheckConnection do
...
 
   @behaviour Realtime.Tenants.Connect.Piper
   @impl true
   def run(acc) do
     %{tenant: tenant} = acc
 
-    case Database.check_tenant_connection(tenant) do
+    case Realtime.Database.check_tenant_connection(tenant) do
       {:ok, conn} ->
-        Process.link(conn)
```
edgurgel, line 14: No need to link as `Database.check_tenant_connection` calls `Postgrex.start_link` which links the process.

lib/realtime/tenants/connect/check_connection.ex
```diff
@@ -2,16 +2,14 @@ defmodule Realtime.Tenants.Connect.CheckConnection do
...
         db_conn_reference = Process.monitor(conn)
```
edgurgel, line 13: In fact this Process.monitor is also not doing much given that they are linked. Most likely won't have time to react to the DOWN message as the linked process will crash `Connect`

filipecabaco (commented):
does this mean that , for some amount of time, users will be able to start connecting and then we will disconnect them?

part of the reason we had this to be more blockng / sync was to actually be sure that the database was fully ready and up to move forward with accepting socket connections so I'm wondering what will be the side effects of having this more async vs just having more partitions

edgurgel:
> does this mean that , for some amount of time, users will be able to start connecting and then we will disconnect them?
> part of the reason we had this to be more blockng / sync was to actually be sure that the database was fully ready and up to move forward with accepting socket connections so I'm wondering what will be the side effects of having this more async vs just having more partitions

We only return a `conn` when `syn` has the `conn` metadata which is after the DB connection is set-up. This is how it works today already and this PR has not changed this. 

This PR just also changes so that we also capture the shutdown while connecting. We won't block on the Supervisor call anymore but we will block waiting for the `syn` message that the `conn` is ready (which we already do today).

Previously if we had 5 websockets from different nodes calling `Connect.lookup_or..`:

* The first one would block calling Supervisor.start_child
* The other 4 websockets would block waiting for the `syn` message saying that `conn` metadata has been set.

This PR changes so that all 5 websockets wait on `syn`

`

filipecabaco (approved):

filipecabaco:
makes sense ship it 👍

PR #1541, opened 2025-09-22
fix: rate limit connect module

filipecabaco:

On bad connection, we rate limit the Connect module so we prevent abuses and too much logging of errors

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.51.4",
+      version: "2.51.5",
       elixir: "~> 1.17.3",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,


lib/realtime/tenants/connect.ex
```diff
@@ -58,21 +57,28 @@ defmodule Realtime.Tenants.Connect do
           | {:error, :tenant_db_too_many_connections}
           | {:error, :rpc_error, term()}
   def lookup_or_start_connection(tenant_id, opts \\ []) when is_binary(tenant_id) do
```
edgurgel, line 59: I think we should only increment when there is an error. `lookup_or_start_connection` is very cheap when the connection is already established and it's not a problem to be called. It's a simple `syn` lookup in the end.

If we only increment when there is an error we will only limit the ones that are constantly failing.

`tenant_database_connection_initializing` happens when there is no connection so not an error in this case.

`initializing` is also fine it might just mean that the connection took some time to connect.

`tenant_database_unavailable` is returned from an error 👍 

WDYT?

lib/realtime/tenants.ex
```diff
@@ -328,6 +328,32 @@ defmodule Realtime.Tenants do
     %RateCounter.Args{id: {:channel, :authorization_errors, external_id}, opts: opts}
   end
 
+  @connect_per_second_default 2
```
edgurgel: Technically that's not per second if we are using `measurement: :sum`. More like total?

Why 2?

lib/realtime/tenants.ex
```diff
@@ -328,6 +328,32 @@ defmodule Realtime.Tenants do
...
+  @doc "RateCounter arguments for counting connect per second."
+  @spec connect_per_second_key(Tenant.t() | String.t()) :: RateCounter.Args.t()
+  def connect_per_second_key(%Tenant{external_id: external_id}) do
```
edgurgel: I think all the other functions are `_rate` when they return RateCounter.Args

lib/realtime/tenants/connect.ex
```diff
@@ -58,21 +57,28 @@ defmodule Realtime.Tenants.Connect do
...
+    RateCounter.new(rate_args)
 
-      {:error, :tenant_database_unavailable} ->
-        {:error, :tenant_database_unavailable}
+    with {:ok, %{limit: %{triggered: false}}} <- RateCounter.get(rate_args),
+         {:ok, conn} <- get_status(tenant_id) do
+      {:ok, conn}
+    else
+      {:ok, %{limit: %{triggered: true}}} ->
+        GenCounter.add(rate_args.id)
```
edgurgel: If it's already triggered I don't think we should count again otherwise they might never have the chance to try again.

also why return `tenant_db_too_many_connections` when `triggered: true` ? Users might think they don't have enough connections but that's not the case, right?

lib/realtime/tenants/connect.ex
filipecabaco: makes sense 👍

lib/realtime/tenants.ex
filipecabaco: oh you are right, changing it and the 2 total was just taken out of nowhere 😆 just low number per second that felt "normal" a we want to avoid a constant barrage of users trying to connect

lib/realtime/tenants.ex
filipecabaco: changing

lib/realtime/tenants/connect.ex
edgurgel: Thoughts?

lib/realtime/tenants.ex
edgurgel: Ok cool so we are looking 10 on avg in 10 seconds (buckets)? 

So if we get 100 total errors in 10 seconds we will limit.

I think it was fine to leave it as `sum` so that if in 10 seconds we got 10 errors we block.

filipecabaco: then I will leave it as sum of 10 so that will make it less noisy and feels that it fits better the current scenario where this happens

lib/realtime/tenants/connect.ex
filipecabaco: good call will change to not count here and will change the error to be more rate limit specific

filipecabaco: needed to actually have a count when we actually attempt to connect otherwise it did not trigger and it does make sense as we actually don't check the result of this call so we end up not knowing.

having this way not only works but ensures we avoid the most expensive part of the process - calling external nodes

edgurgel (approved):
LGTM

PR #1545, opened 2025-09-27
build: automatically cancel old tests/build on new push

kevcodez:
Currently, whenever you push any commit to your branch, the old builds are still running and a new build is started. Once a new commit is added, the old test results no longer matter and it's just a waste of CI resources. Also reduces confusion with multiple builds running in parallel for the same branch/possibly blocking any merges.

With this little change, we ensure that whenever a new commit is added, the previous build is immediately canceled/stopped and only the build (latest commit) runs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,6 +16,10 @@ on:
     branches:
       - main
 
+concurrency:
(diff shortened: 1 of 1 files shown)


edgurgel (approved):
LGTM

PR #1547, opened 2025-09-30
fix: set max socket message queue length

filipecabaco:

We're using the event to update the counter to check if the websocket has too many messages. If so we will consider abusive behaviour and will kill the socket to prevent further damage.

### DIFF
--- a/lib/realtime/monitoring/erl_sys_mon.ex
+++ b/lib/realtime/monitoring/erl_sys_mon.ex
@@ -24,8 +24,8 @@ defmodule Realtime.ErlSysMon do
     {:ok, []}
   end
 
-  def handle_info({:monitor, pid, _type, _meta} = msg, state) when is_pid(pid) do
-    log_process_info(msg, pid)
+  def handle_info({:monitor, pid, type, _meta} = msg, state) when is_pid(pid) do
+    log_process_info(msg, pid, type)
     {:noreply, state}
   end
 
@@ -34,7 +34,7 @@ defmodule Realtime.ErlSysMon do
     {:noreply, state}
   end
 
-  defp log_process_info(msg, pid) do
+  defp log_process_info(msg, pid, type) do
     pid_info =
       pid
       |> Process.info(:dictionary)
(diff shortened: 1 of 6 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -238,16 +237,24 @@ defmodule RealtimeWeb.RealtimeChannel do
     {:noreply, socket}
   end
 
+  @websocket_message_queue_length_limit 1000
   def handle_info(:update_rate_counter, socket) do
     count(socket)
 
     {:ok, rate_counter} = RateCounter.get(socket.assigns.rate_counter)
+    {:message_queue_len, len} = Process.info(self(), :message_queue_len) |> dbg()
```
edgurgel: The process here is not the one that will receive the messages from fastlane. And it's not the one that can reach the max heap size limit.

I think we should move this check to happen at the MessageDispatcher level instead and remove this `update_rate_counter`.

And we can check the `fastlane_pid` (the actual cowboy websocket process) for their `message_queue_len`

filipecabaco: makes sense will tackle first the change to be in message dispatcher. will also try to find a way to simplify the code as currently is getting a bit trickier to read

lib/realtime/monitoring/erl_sys_mon.ex
```diff
@@ -56,4 +57,14 @@ defmodule Realtime.ErlSysMon do
     _ ->
       Logger.warning("#{__MODULE__} message: " <> inspect(msg))
   end
+
+  defp defensive_exit?(pid, pid_info, type) do
+    {{:"$initial_call", {mod, _, _}}, _} = pid_info
+
+    # credo:disable-for-next-line
+    cond do
+      mod == Realtime.Integration.WebsocketClient && type == :long_message_queue -> Process.exit(pid, :kill)
```
edgurgel: This module is the one we use for testing right? We might need to look for the cowboy websocket one 

Also for long_message_queue we might want to not kill on the current threshold (1000) but maybe if it goes beyond a certain number. Might be worth looking at some prod data?

filipecabaco: yes it is and indeed we do just a test drive of the idea and more to start the discussion you mentioned on how should we handle killing?

will check prod data to see what I find

filipecabaco:
Currently this will bring more complexity with less gains so we're not moving forward with it

PR #1566, opened 2025-10-08
fix: shard user scopes in syn

filipecabaco:

shard user scopes in syn


lib/realtime/tenants.ex
```diff
@@ -21,7 +21,9 @@ defmodule Realtime.Tenants do
   """
   @spec list_connected_tenants(atom()) :: [String.t()]
   def list_connected_tenants(node) do
-    :syn.group_names(:users, node)
+    UsersCounter.scopes()
+    |> Enum.map(fn scope -> :syn.group_names(scope, node) end)
+    |> List.flatten()
```
edgurgel: nitpick: `Enum.flat_map`

PR #1567, opened 2025-10-08
fix: refine join payload checking

filipecabaco:

refine join payload checking to handle certain scenarios where we can consider valid:
* numeric presence key
* `[nil]` postgres changes to be considered `[]`


lib/realtime_web/channels/payloads/presence.ex
```diff
@@ -8,10 +8,20 @@ defmodule RealtimeWeb.Channels.Payloads.Presence do
...
   end
 
   def changeset(presence, attrs) do
-    cast(presence, attrs, [:enabled, :key], message: &Join.error_message/2)
+    presence
+    |> cast(attrs, [:enabled], message: &Join.error_message/2)
+    |> cast_to_string(attrs, :key, &UUID.uuid1/0)
+  end
+
+  defp cast_to_string(changeset, attrs, field, default_fun) do
```
edgurgel: But will this keep the previous behaviour? Will casting to string cause any downstream change?

Could we simply have no type here on the field and on the validation check if integer OR string?

filipecabaco: good call it might break users expectations

filipecabaco: pushed the fix

PR #1574, opened 2025-10-14
fix: validate size of track message

filipecabaco:

Validates the size of the track message sent and fails if exceeds the value set by `tenant.max_payload_size_in_kb`

### DIFF
--- a/test/integration/rt_channel_test.exs
+++ b/test/integration/rt_channel_test.exs
@@ -714,36 +714,6 @@ defmodule Realtime.Integration.RtChannelTest do
       assert get_in(join_payload, ["t"]) == payload.payload.t
     end
 
-    test "presence track closes on high payload size", %{tenant: tenant} do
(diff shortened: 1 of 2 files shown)


lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -124,6 +125,9 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
       {:error, :rate_limit_exceeded} ->
         {:error, :rate_limit_exceeded}
 
+      {:error, :payload_size_exceeded} ->
+        {:error, :payload_size_exceeded}
```
edgurgel, line 134: Do we need a new error code for RealtimeChannel?

filipecabaco: yes we do, added

edgurgel (approved):
LGTM

PR #1576, opened 2025-10-14
fix: provide microsecond granularity by using binary pgoutput mode

filipecabaco:

* When `pgoutput` uses binary mode goes to https://doxygen.postgresql.org/proto_8c_source.html#l00767 
* We end up in `OidSendFunctionCall` which gets the write function to call and transform into a binary payload based on OID - https://doxygen.postgresql.org/fmgr_8c.html#ad028dec5dc28e883308be264689f4559
* The functions are referenced in https://doxygen.postgresql.org/dir_19ba549c088f18020d0eadfcd07e87be.html 
* It does `<data_tupe>.c` and calls `<data_type>_send` as the function to use
    * Jsonb - https://doxygen.postgresql.org/jsonb_8c.html#afa4708e251f3078084e9986a46f228e9
        * First byte is JSONB version (1), remainder is the text value
    * Timestamp - https://doxygen.postgresql.org/backend_2utils_2adt_2timestamp_8c.html#ab0f80b3ae44bdcc50d0c0abb0c1d7939
        * 64 bytes integer with microseconds since `2000-01-01`
    * Boolean - https://doxygen.postgresql.org/bool_8c.html#a71cfb4f6c28259924b231dae4781b3fc
        * 1 is true, 0 is false
    * UUID - https://doxygen.postgresql.org/uuid_8c.html#a3cd205c61c108aaf08d6746c1d321e81 
        * 16 byte binary ( https://doxygen.postgresql.org/uuid_8h.html#a9692a0205a857ed2cc29558470c2ed77 ) 
    * Text - https://doxygen.postgresql.org/varlena_8c.html#a0ea0ddd71c614e3fbd3571e70c53ed77
        * Sends binary data

### DIFF
--- a/lib/realtime/adapters/postgres/decoder.ex
+++ b/lib/realtime/adapters/postgres/decoder.ex
@@ -318,16 +318,16 @@ defmodule Realtime.Adapters.Postgres.Decoder do
         <<0>> ->
           false
 
-        <<1, binary::binary-size(column_length - 1)>> ->
-          binary
+        <<uuid_binary::binary-16>> ->
+          UUID.binary_to_string!(uuid_binary)
 
         <<microseconds::signed-big-64>> ->
           @start_date
           |> NaiveDateTime.from_iso8601!()
           |> NaiveDateTime.add(microseconds, :microsecond)
 
-        <<uuid_binary::binary-16>> ->
-          UUID.binary_to_string!(uuid_binary)
+        <<1, binary::binary-size(column_length - 1)>> ->
+          binary
 
         data when is_binary(data) ->
           data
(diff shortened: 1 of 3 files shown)


lib/realtime/monitoring/prom_ex/plugins/tenant.ex
```diff
@@ -245,7 +245,7 @@ defmodule Realtime.PromEx.Plugins.Tenant do
           unit: :second,
```
edgurgel: You might need to set this to `unit: {:microsecond, :millisecond}` I think? We usually provide milliseconds like this:

https://github.com/supabase/realtime/blob/b8dd5f384f30303cfca7a4039737f8f3d796bff4/lib/realtime/monitoring/prom_ex/plugins/tenant.ex#L119

PR #1607, opened 2025-11-04
fix: measure egress of output bytes for all sockets for a given tenant

filipecabaco:

measure egress of output bytes for all sockets for a given tenant


filipecabaco:
https://github.com/supabase/realtime/pull/1637

PR #1620, opened 2025-11-18
feat: tenant crud operations are region aware

filipecabaco:

Routes CRUD operations to master node when such is set via environment variable

### DIFF
--- a/lib/realtime/api.ex
+++ b/lib/realtime/api.ex
@@ -273,7 +273,18 @@ defmodule Realtime.Api do
 
   defp remote_call(operation, args, tenant_id) do
     master_region = Application.get_env(:realtime, :master_region)
-    {:ok, master_node} = Nodes.node_from_region(master_region, self())
-    GenRpc.call(master_node, Realtime.Repo, operation, args, tenant_id: tenant_id)
+
+    with {:ok, master_node} <- Nodes.node_from_region(master_region, self()),
+         {:ok, result} <- wrapped_call(master_node, operation, args, tenant_id) do
+      result
+    end
+  end
+
+  defp wrapped_call(master_node, operation, args, tenant_id) do
+    case GenRpc.call(master_node, Realtime.Repo, operation, args, tenant_id: tenant_id) do
+      {:error, :rpc_error, reason} -> {:error, reason}
+      {:error, reason} -> {:error, reason}
+      result -> {:ok, result}
+    end
   end
 end
--- a/test/integration/region_aware_routing_test.exs
+++ b/test/integration/region_aware_routing_test.exs
@@ -5,6 +5,7 @@ defmodule Realtime.Integration.RegionAwareRoutingTest do
   alias Realtime.Api
   alias Realtime.Api.Tenant
   alias Realtime.GenRpc
+  alias Realtime.Nodes
 
   setup do
     # Configure test runner as non-master region (eu-west-1) with master_region = us-east-1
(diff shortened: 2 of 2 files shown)


lib/realtime/api.ex
```diff
@@ -243,4 +243,32 @@ defmodule Realtime.Api do
...
+      else: remote_call(operation, [queryable], tenant_id)
+  end
+
+  defp local_call(operation, args, _tenant_id) do
+    apply(Realtime.Repo, operation, args)
+  end
+
+  defp remote_call(operation, args, tenant_id) do
+    master_region = Application.get_env(:realtime, :master_region)
+    {:ok, master_node} = Nodes.node_from_region(master_region, tenant_id)
```
edgurgel: What if we just use `self()` for the `node_from_region` key (second argument)? This way we don't need to pass `tenant_id` at all.

 `self()` is a good key as the cardinality is pretty high. That's what [PG2 pubsub](https://github.com/phoenixframework/phoenix_pubsub/blob/43ab10bfcec56f82e915c246c2dd9dba438fe8da/lib/phoenix/pubsub/pg2.ex#L45) uses for their pub sub worker selection for example.

lib/realtime/api.ex
```diff
@@ -243,4 +243,32 @@ defmodule Realtime.Api do
...
+    if local_call?(),
+      do: local_call(operation, [changeset], tenant_id),
+      else: remote_call(operation, [changeset], tenant_id)
+  end
+
+  defp region_aware_write(queryable, operation, tenant_id) when is_struct(queryable, Ecto.Query) do
+    if local_call?(),
+      do: local_call(operation, [queryable], tenant_id),
+      else: remote_call(operation, [queryable], tenant_id)
+  end
```
edgurgel: Could these two be a single function?

```elixir
defp region_aware_write(%struct{}, operation, tenant_id) when struct in [Changeset, Ecto.Query] do
...
```

I think this should work 🤔

test/integration/region_aware_routing_test.exs
```diff
@@ -0,0 +1,166 @@
...
+
+    # Start peer node as master region (us-east-1)
+    # The master node will automatically register itself in RegionNodes on startup
+    {:ok, master_node} =
+      Clustered.start(nil,
+        extra_config: [
+          {:realtime, :region, "us-east-1"},
+          {:realtime, :master_region, "us-east-1"}
+        ]
+      )
```
edgurgel, line 24: I'm just trying to understand this whole setup so sorry for any stupid questions!

Could we instead of changing the local region and the syn region for the main/local node just:

* Set the `master_region` to `eu-west-1`
* Start the new remote node with region and master_region set to `eu-west-1`

This way you don't need to change the local node that is running. Less things to go wrong I think? No need to syn.leave and syn.join twice and no need to change the `region`

The only reason why I am asking is that changing region and messing with syn RegionNodes for the main node can cause flakyness.

WDYT?

edgurgel (commented):
Added some comments!

Do we need to update this [function as well](https://github.com/supabase/realtime/blob/0de9c160bc06ea637618f6addad12d12e13f7502/lib/realtime/tenants.ex#L490-L496) to call `Api.update_tenant`?

test/integration/region_aware_routing_test.exs
filipecabaco: will try it out 👍 the testing of this feature is really weird so will try anything that simplifies it 💀

filipecabaco: actually we need to keep it as is so we can test that we are routing the requests to the remote node which is impersonating the master write 🤔

edgurgel: This is what I meant: https://github.com/supabase/realtime/pull/1623

I also changed to use `call_original` from Mimic to actually perform the gen rpc call. This way it actually goes through gen rpc

lib/realtime/api.ex
```diff
@@ -243,4 +256,24 @@ defmodule Realtime.Api do
...
+    if local_call?(),
+      do: local_call(operation, [argument], tenant_id),
+      else: remote_call(operation, [argument], tenant_id)
+  end
+
+  defp local_call(operation, args, _tenant_id), do: apply(Realtime.Repo, operation, args)
+
+  defp remote_call(operation, args, tenant_id) do
+    master_region = Application.get_env(:realtime, :master_region)
+    {:ok, master_node} = Nodes.node_from_region(master_region, self())
```
edgurgel: Do we want to treat the error case and log something and return a `{:error, _}` ? We might want a test for this as well. 

The functions that end up calling this like `create_tenant` expects {:ok, tenant} or `{:error, changeset}` I think?

We can write a unit test for this case by stubbing `Nodes.nodes_from_region` to return `{:error, :not_available}`

Also GenRpc.call can return `{:error, :rpc_error, reason}` so we might want to account for that as well.

filipecabaco: yes we do will add

edgurgel (approved):
LGTM

PR #1624, opened 2025-11-21
feat: filter postgres changes by action

edgurgel:
Follow-up from https://github.com/supabase/walrus/pull/77

Filter postgres changes based on action

### DIFF
--- a/lib/extensions/postgres_cdc_rls/subscriptions.ex
+++ b/lib/extensions/postgres_cdc_rls/subscriptions.ex
@@ -8,7 +8,7 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
 
   @type conn() :: Postgrex.conn()
   @type filter :: {binary, binary, binary}
-  @type subscription_params :: {binary, binary, [filter]}
+  @type subscription_params :: {action_filter :: binary, schema :: binary, table :: binary, [filter]}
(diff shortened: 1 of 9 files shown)


lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -36,26 +36,32 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
         do update set
         claims = excluded.claims,
         created_at = now()
       returning
          id"
 
     transaction(conn, fn conn ->
-      Enum.map(subscription_list, fn %{id: id, claims: claims, subscription_params: params = {schema, table, filters}} ->
-        case query(conn, sql, [publication, schema, table, id, claims, filters]) do
+      Enum.map(subscription_list, fn %{
```
filipecabaco: later we should extract this to a function as this is a bit cluttered but not important now

edgurgel:
> did we already ship the walrus change?

The walrus PR has been merged if that's the question 🤔

lib/extensions/postgres_cdc_rls/subscriptions.ex
edgurgel: Good point. Will refactor!

filipecabaco:
yes it was 😁 didn't remember if it did happen approving to unblock

PR #1625, opened 2025-11-24
fix: replicate pg_output issue

filipecabaco:

If the string size matches the uuid length we were parsing it to be uuid which created issues. To prevent it we now handle payload based on the relations provided and we further simplified the logic to not use send in our message handling

### DIFF
--- a/lib/realtime/adapters/postgres/decoder.ex
+++ b/lib/realtime/adapters/postgres/decoder.ex
@@ -149,12 +149,6 @@ defmodule Realtime.Adapters.Postgres.Decoder do
 
   @doc """
   Parses logical replication messages from Postgres
-
-  ## Examples
-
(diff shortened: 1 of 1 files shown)


lib/realtime/adapters/postgres/decoder.ex
edgurgel: Doc needs to be updated (if we want to have docs ofc)

lib/realtime/adapters/postgres/decoder.ex
```diff
@@ -288,52 +233,53 @@ defmodule Realtime.Adapters.Postgres.Decoder do
...
-        <<1>> ->
-          true
+      case type do
+        "bool" ->
+          data == <<1>>
 
-        <<0>> ->
-          false
+        "jsonb" ->
+          <<1, binary::binary-size(column_length - 1)>> = data
```
edgurgel: Looking at the docs for jsonb we just need to discard the 1(version) so this should be enough:

```elixir
<<1, rest::binary>> = data
rest
```

edgurgel (approved):
Added a couple of comments! LGTM

PR #1628, opened 2025-11-25
feat: nodejs integration tests

edgurgel:

* Add postgres changes tests to deno integration tests
* Add broadcast integration tests to the new nodejs integration tests.


filipecabaco:
tackled by new e2e

edgurgel:
But that's using bun right?

filipecabaco:
yes but we have integration tests for nodejs in supabase-js

edgurgel:
Yeah but the tests here are for the `main` branch not whatever the supabase cli has as stable/latest. It's a way to check if any backend change caused issues with existing libraries.

We already run `deno` tests this PR was to add the nodejs version of it: https://github.com/supabase/realtime/blob/main/test/integration/tests.ts

filipecabaco:
got it let's reopen then

PR #1632, opened 2025-11-27
fix: measure traffic at the websocket level

filipecabaco:

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

### DIFF
--- a/test/integration/measure_traffic_test.exs
+++ b/test/integration/measure_traffic_test.exs
@@ -1,7 +1,5 @@
 defmodule Realtime.Integration.MeasureTrafficTest do
-  use RealtimeWeb.ConnCase,
-    async: false,
-    parameterize: [%{serializer: Phoenix.Socket.V1.JSONSerializer}, %{serializer: RealtimeWeb.Socket.V2Serializer}]
+  use RealtimeWeb.ConnCase, async: false
 
   alias Phoenix.Socket.Message
   alias Realtime.Integration.WebsocketClient
@@ -71,8 +69,8 @@ defmodule Realtime.Integration.MeasureTrafficTest do
       :ok
     end
 
-    test "measure traffic for broadcast events", %{tenant: tenant, serializer: serializer} do
-      {socket, _} = get_connection(tenant, serializer)
+    test "measure traffic for broadcast events", %{tenant: tenant} do
+      {socket, _} = get_connection(tenant)
       config = %{broadcast: %{self: true}}
       topic = "realtime:any"
 
@@ -108,8 +106,8 @@ defmodule Realtime.Integration.MeasureTrafficTest do
       assert input_bytes > 0
     end
 
-    test "measure traffic for presence events", %{tenant: tenant, serializer: serializer} do
-      {socket, _} = get_connection(tenant, serializer)
+    test "measure traffic for presence events", %{tenant: tenant} do
+      {socket, _} = get_connection(tenant)
       config = %{broadcast: %{self: true}, presence: %{enabled: true}}
       topic = "realtime:any"
 
@@ -138,8 +136,8 @@ defmodule Realtime.Integration.MeasureTrafficTest do
       assert input_bytes > 0, "Expected input_bytes to be greater than 0, got #{input_bytes}"
     end
 
-    test "measure traffic for postgres changes events", %{tenant: tenant, serializer: serializer, db_conn: db_conn} do
-      {socket, _} = get_connection(tenant, serializer)
+    test "measure traffic for postgres changes events", %{tenant: tenant, db_conn: db_conn} do
+      {socket, _} = get_connection(tenant)
       config = %{broadcast: %{self: true}, postgres_changes: [%{event: "*", schema: "public"}]}
       topic = "realtime:any"
 
@@ -185,8 +183,8 @@ defmodule Realtime.Integration.MeasureTrafficTest do
       assert input_bytes > 0, "Expected input_bytes to be greater than 0, got #{input_bytes}"
     end
 
-    test "measure traffic for db events", %{tenant: tenant, serializer: serializer, db_conn: db_conn} do
-      {socket, _} = get_connection(tenant, serializer)
+    test "measure traffic for db events", %{tenant: tenant, db_conn: db_conn} do
+      {socket, _} = get_connection(tenant)
       config = %{broadcast: %{self: true}, db: %{enabled: true}}
       topic = "realtime:any"
       channel_name = "any"
--- a/test/realtime/monitoring/prom_ex/plugins/tenant_test.exs
+++ b/test/realtime/monitoring/prom_ex/plugins/tenant_test.exs
@@ -300,27 +300,21 @@ defmodule Realtime.PromEx.Plugins.TenantTest do
     test "channel input bytes", context do
       external_id = context.tenant.external_id
 
-      pattern = ~r/realtime_channel_input_bytes{tenant="#{external_id}"}\s(?<number>\d+)/
-
-      metric_value = metric_value(pattern)
       FakeUserCounter.fake_input_bytes(external_id)
       FakeUserCounter.fake_input_bytes(external_id)
 
       Process.sleep(200)
-      assert metric_value(pattern) == metric_value + 20
+      assert metric_value("realtime_channel_input_bytes", tenant: external_id) == 20
     end
 
     test "channel output bytes", context do
       external_id = context.tenant.external_id
 
-      pattern = ~r/realtime_channel_output_bytes{tenant="#{external_id}"}\s(?<number>\d+)/
-
-      metric_value = metric_value(pattern)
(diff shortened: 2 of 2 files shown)


lib/realtime_web/channels/user_socket.ex
```diff
@@ -4,7 +4,29 @@ defmodule RealtimeWeb.UserSocket do
...
+
+  @impl true
+  def handle_info(:measure_traffic, {channels, %{assigns: assigns} = socket}) do
+    ref = Map.get(assigns, :measure_traffic_ref)
+    tenant_external_id = Map.get(assigns, :tenant)
+    prev_bytes = Map.get(assigns, :prev_traffic_bytes, %{send_oct: 0, recv_oct: 0})
+    prev_bytes = collect_traffic_telemetry(socket.transport_pid, tenant_external_id, prev_bytes)
+
+    Process.cancel_timer(ref)
+    ref = Process.send_after(self(), :measure_traffic, measure_traffic_interval_in_ms())
```
edgurgel: No need to change but we could avoid storing the prev bytes by simply sending as part of the message you send to yourself `{:measure_traffic, prev_bytes}` and the first message can send `{measure_traffic, %{send_oct: 0, recv_oct: 0}}`

I also don't know if we need to store the `ref` and cancel the timer here as there is no other way for this message to arrive so the timer should always not be there I think?

This way we avoid touching the `assigns` and leaving t he handle info a bit cleaner

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -23,6 +23,11 @@ defmodule RealtimeWeb.RealtimeChannelTest do
   setup do
     tenant = Containers.checkout_tenant(run_migrations: true)
     Cachex.put!(Realtime.Tenants.Cache, {{:get_tenant_by_external_id, 1}, [tenant.external_id]}, {:cached, tenant})
+
+    measure_traffic_interval_in_ms = Application.get_env(:realtime, :measure_traffic_interval_in_ms)
+    Application.put_env(:realtime, :measure_traffic_interval_in_ms, 100)
+    on_exit(fn -> Application.put_env(:realtime, :measure_traffic_interval_in_ms, measure_traffic_interval_in_ms) end)
```
edgurgel: BTW this test file is `async: true` so changing env would probably move this to be `async: false` :sadpanda:

test/integration/rt_channel_test.exs
```diff
@@ -2332,6 +2346,194 @@ defmodule Realtime.Integration.RtChannelTest do
...
+          start: {Agent, :start_link, [fn -> %{} end, [name: name]]}
+        })
+
+      RateCounterHelper.stop(tenant.external_id)
+      on_exit(fn -> :telemetry.detach({__MODULE__, tenant.external_id}) end)
+      :telemetry.attach_many({__MODULE__, tenant.external_id}, events, &__MODULE__.handle_telemetry/4, name)
+
+      measure_traffic_interval_in_ms = Application.get_env(:realtime, :measure_traffic_interval_in_ms)
+      Application.put_env(:realtime, :measure_traffic_interval_in_ms, 10)
+      on_exit(fn -> Application.put_env(:realtime, :measure_traffic_interval_in_ms, measure_traffic_interval_in_ms) end)
```
edgurgel: BTW this test file is async: true so changing env would probably move this to be async: false :sadpanda:

Maybe we just create a new integration file just for measure traffic and make it async false?

that's what I did for the distributed integration test that uses `Clustered`: https://github.com/supabase/realtime/blob/main/test/integration/distributed_realtime_channel_test.exs

test/realtime_web/channels/realtime_channel_test.exs
edgurgel: Do we even need unit tests for this? 🤔

test/integration/rt_channel_test.exs
```diff
@@ -2332,6 +2346,194 @@ defmodule Realtime.Integration.RtChannelTest do
...
+
+      :ok
+    end
+
+    test "measure traffic for broadcast events", %{tenant: tenant, serializer: serializer} do
```
edgurgel: do we need to test each event type? 🤔 I would expect a single test being enough as long as the bytes go up with messages going back and forth. WDYT? Integration tests are slow so the least we add the better if possible

Broadcast and Broadcast from Database tests are at least redundant as they end up calling the same code in the end and they are not treated differently.

lib/realtime_web/channels/user_socket.ex
```diff
@@ -144,4 +166,39 @@ defmodule RealtimeWeb.UserSocket do
...
 
   defp max_heap_size(), do: Application.fetch_env!(:realtime, :websocket_max_heap_size)
+  defp measure_traffic_interval_in_ms(), do: Application.fetch_env!(:realtime, :measure_traffic_interval_in_ms)
+
+  defp collect_traffic_telemetry(nil, _tenant_external_id, prev_bytes), do: prev_bytes
+
+  defp collect_traffic_telemetry(transport_pid, tenant_external_id, prev_bytes) do
+    transport_pid
+    |> Process.info(:links)
+    |> then(fn {:links, links} -> links end)
```
edgurgel: Do we expect more than 1 port being here? If we do we would need to track prev bytes per port then. I wonder if we should always just get 1 port here and not call reduce so that's explicit that we only want to account for 1 TCP connection and if one day this returns more than 1 port we probably want to account for multiple ports. 

Another option is to accumulate the new bytes and with the result of `Enum.reduce` we call telemetry with the difference and return the result of Enum.reduce as new values. I hope this makes sense

filipecabaco: I don't expect it, it's to prevent it from blowing up

test/integration/rt_channel_test.exs
filipecabaco: just checking that we are indeed sending for all features we have more than anything as a regression test

test/realtime_web/channels/realtime_channel_test.exs
filipecabaco: no, removing it

lib/realtime_web/channels/user_socket.ex
```diff
@@ -4,9 +4,25 @@ defmodule RealtimeWeb.UserSocket do
...
+  def handle_info(
+        {:measure_traffic, previous_recv, previous_send},
+        {_, %{assigns: assigns, transport_pid: transport_pid}} = state
+      ) do
+    tenant_external_id = Map.get(assigns, :tenant)
+
+    %{send_delta: send_delta, recv_delta: recv_delta} =
+      collect_traffic_telemetry(transport_pid, tenant_external_id, previous_recv, previous_send)
+
+    Process.send_after(self(), {:measure_traffic, recv_delta, send_delta}, measure_traffic_interval_in_ms())
```
edgurgel: Should we be sending delta 🤔 ? We should be sending the actual value I think? 

Example:

* Websocket starts with {0, 0}.
* First stats are {10, 5}. We emit the delta which is 10 - 0, 5 - 0. Schedule message with {10, 5}
* Second stats are {25, 7}. We emit the delta which is 25 - 10, 7 - 5. Schedule message with {25, 7}

filipecabaco: you are right, pushed the fix

lib/realtime/monitoring/prom_ex/plugins/tenant.ex
```diff
@@ -197,6 +197,20 @@ defmodule Realtime.PromEx.Plugins.Tenant do
...
+          measurement: :size,
+          tags: [:tenant]
+        ),
+        sum(
+          [:realtime, :channel, :output_bytes],
+          event_name: [:realtime, :channel, :output_bytes],
+          description: "Sum of output bytes sent on sockets.",
+          measurement: :size,
+          tags: [:tenant]
+        ),
```
edgurgel, line 235: If we intend to have also a "global" metric on traffic we might want a `[:realtime, :global, :channel, :output_byes]` without the tenant tag. Not sure if that's what we want but I thought I would point it out.

test/integration/measure_traffic_test.exs
```diff
@@ -0,0 +1,235 @@
+defmodule Realtime.Integration.MeasureTrafficTest do
+  use RealtimeWeb.ConnCase,
+    async: false,
+    parameterize: [%{serializer: Phoenix.Socket.V1.JSONSerializer}, %{serializer: RealtimeWeb.Socket.V2Serializer}]
```
edgurgel: don't think we need to parameterize here as the serializer won't change any implementation

edgurgel (approved):
LGTM

lib/realtime/monitoring/prom_ex/plugins/tenant.ex
filipecabaco: maybe later not for now

PR #1633, opened 2025-11-27
fix: improve crud remote calls

filipecabaco:

This approach prevents sending more information than needed reducing the amount of data send via gen_rpc. We also remove some unnecessary calls to primary in the delete controller operation

### DIFF
--- a/lib/realtime/api.ex
+++ b/lib/realtime/api.ex
@@ -127,22 +127,20 @@ defmodule Realtime.Api do
   @doc """
   Updates a tenant.
   """
-  @spec update_tenant(binary() | Tenant.t(), map()) :: {:ok, Tenant.t()} | {:error, term()}
-  def update_tenant(tenant_id, attrs) when is_binary(tenant_id) do
+  @spec update_tenant_by_external_id(binary(), map()) :: {:ok, Tenant.t()} | {:error, term()}
+  def update_tenant_by_external_id(tenant_id, attrs) when is_binary(tenant_id) do
     if master_region?() do
       tenant_id
       |> get_tenant_by_external_id(use_replica?: false)
       |> update_tenant(attrs)
     else
-      call(:update_tenant, [tenant_id, attrs], tenant_id)
+      call(:update_tenant_by_external_id, [tenant_id, attrs], tenant_id)
     end
   end
 
-  def update_tenant(%Tenant{external_id: external_id} = tenant, attrs) do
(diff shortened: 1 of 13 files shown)


lib/realtime/api.ex
```diff
@@ -150,38 +163,23 @@ defmodule Realtime.Api do
...
-      _ -> false
+    if master_region?() do
+      from(t in Tenant, where: t.external_id == ^id)
+      |> Repo.delete_all()
+      |> case do
+        {num, _} when num > 0 -> true
+        _ -> false
+      end
+    else
+      call(:delete_tenant_by_external_id, [id], id)
```
edgurgel, line 166: Not a huge problem but I will mention anyway:

This way of doing the RPC call has the potential for a distributed call loop IF the nodes don't have a consistent configuration on master region. Say if you have something like:

* Node A (region=us-east-1, master_region=eu-west-2)
* Node B (region=eu-west-2, master_region=us-east-1)

Where both nodes think that the other region is the master region they will keep calling each other.

lib/realtime/api.ex
```diff
@@ -115,26 +115,34 @@ defmodule Realtime.Api do
...
+      call(:update_tenant, [tenant_id, attrs], tenant_id)
+    end
+  end
+
+  def update_tenant(%Tenant{external_id: external_id} = tenant, attrs) do
     changeset = Tenant.changeset(tenant, attrs)
-    updated = region_aware_write(changeset, :update, tenant.external_id)
+
+    updated =
+      if master_region?(), do: Repo.update(changeset), else: call(:update_tenant, [external_id, attrs], external_id)
```
edgurgel: I'm a bit confused here. Why is this one calling a remote call on another function and using the local changeset on the way back? 🤔 Could this be a problem?

Could we just change this to be a private function and make all users of `update_tenant` just pass tenant_id and attrs? I think we only have one or two places doing this anyway

lib/realtime/api.ex
filipecabaco: TBD as this feels like it needs to be controlled at CD level more than anywhere else 😓

lib/realtime/api.ex
edgurgel: This way only the function with external_id and attrs can be called and we only need to do a remote call there?

filipecabaco: will do

filipecabaco: will need to update how we call update in a couple of spots nothing big

PR #1646, opened 2025-12-04
fix: allow nullable jwt secret

filipecabaco:

allow nullable jwt secret

### DIFF
--- a/priv/repo/migrations/20251204170944_nullable_jwt_secrets.exs
+++ b/priv/repo/migrations/20251204170944_nullable_jwt_secrets.exs
@@ -1,0 +1,13 @@
+defmodule Realtime.Repo.Migrations.NullableJwtSecrets do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tenants) do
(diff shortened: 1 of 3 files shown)


priv/repo/migrations/20251204170944_nullable_jwt_secrets.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Realtime.Repo.Migrations.NullableJwtSecrets do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tenants) do
+      modify :jwt_secret, :string, null: true
+    end
+
+    create constraint(:tenants, :jwt_secret_or_jwt_jwks_required,
```
edgurgel, line 9: Do we want to add this to the [changeset](https://hexdocs.pm/ecto/Ecto.Changeset.html#check_constraint/3)?

filipecabaco: added 👍

PR #1657, opened 2025-12-10
fix: ensure replication connection publication only has correct tables

filipecabaco:

Ensures replication connection publication only has correct tables by checking during the connect steps what tables are associated with `supabase_realtime_messages_publication`


test/realtime/tenants/replication_connection_test.exs
```diff
@@ -335,7 +339,7 @@ defmodule Realtime.Tenants.ReplicationConnectionTest do
           assert received_payload == payload
         end)
 
-      assert logs =~ "Unexpected relation on schema 'public' and table 'test'"
+      assert logs =~ "Recreating"
```
edgurgel: Could we keep this test? We can start ReplicationConnection and then change the publication and still capture this case which is possible AFAIK.

PR #1659, opened 2025-12-15
fix: run migrations on tenants nearest region

filipecabaco:

Runs migrations in tenants closest region. Removed additional check on db status as the migration will provide that already.

### DIFF
--- a/test/integration/region_aware_migrations_test.exs
+++ b/test/integration/region_aware_migrations_test.exs
@@ -41,8 +41,6 @@ defmodule Realtime.Integration.RegionAwareMigrationsTest do
 
     Process.sleep(100)
 
-    on_exit(fn -> Clustered.stop() end)
-
     %{tenant: tenant, node: node}
   end
 


lib/realtime/tenants/migrations.ex
```diff
@@ -161,14 +163,31 @@ defmodule Realtime.Tenants.Migrations do
...
+    attrs = %__MODULE__{
+      tenant_external_id: tenant.external_id,
+      settings: settings,
+      migrations_ran: tenant.migrations_ran
+    }
+
+    region = Tenants.region(tenant)
+
+    node =
+      case Nodes.node_from_region(region, tenant.external_id) do
```
edgurgel: Any reason to not use `Nodes.get_node_for_tenant` ? 🤔

test/integration/region_aware_migrations_test.exs
```diff
@@ -0,0 +1,61 @@
...
+
+      arg = hd(args)
+      assert arg.tenant_external_id == tenant.external_id
+      assert arg.migrations_ran == tenant.migrations_ran
+      assert arg.settings == hd(tenant.extensions).settings
+
+      call_original(Realtime.GenRpc, :call, [node, mod, func, args, opts])
+    end)
+
+    assert :ok = Migrations.run_migrations(tenant)
```
edgurgel, line 65: Should we check if migrations_ran was updated?

test/integration/region_aware_migrations_test.exs
```diff
@@ -0,0 +1,61 @@
...
+      }
+    ]
+
+    tenant = tenant_fixture(%{extensions: settings})
+
+    {:ok, node} = Clustered.start(nil, extra_config: [{:realtime, :region, Tenants.region(tenant)}])
+
+    on_exit(fn -> Clustered.stop() end)
+
+    Process.sleep(1000)
```
edgurgel: do you need this sleep?

test/realtime_web/controllers/tenant_controller_test.exs
```diff
@@ -419,7 +419,7 @@ defmodule RealtimeWeb.TenantControllerTest do
 
       conn = get(conn, ~p"/api/tenants/#{tenant.external_id}/health")
       data = json_response(conn, 200)["data"]
-      Process.sleep(2000)
+      Process.sleep(4000)
```
edgurgel: I wonder why we need this sleep. The GET should not return without waiting for migrations to run right? 🤔

lib/realtime/tenants/migrations.ex
filipecabaco: TIL we had that, will use it

test/integration/region_aware_migrations_test.exs
filipecabaco: will do 👍

test/integration/region_aware_migrations_test.exs
filipecabaco: will try to remove but I think it was a bit on the flaky side without it

test/realtime_web/controllers/tenant_controller_test.exs
filipecabaco: will check but I think it was flaky without it

edgurgel: Oh yeah I remember being flaky but I just don't know why it would be flaky if the endpoint should not return until the migration has finished 😭

lib/realtime/tenants/migrations.ex
edgurgel: That's what we use on Connect if I'm not mistaken

lib/realtime/api.ex
```diff
@@ -129,6 +129,8 @@ defmodule Realtime.Api do
   """
   @spec update_tenant_by_external_id(binary(), map()) :: {:ok, Tenant.t()} | {:error, term()}
   def update_tenant_by_external_id(tenant_id, attrs) when is_binary(tenant_id) do
+    IO.inspect(node())
```
edgurgel: 👀

lib/realtime/tenants.ex
```diff
@@ -98,13 +98,11 @@ defmodule Realtime.Tenants do
 
       connected_cluster when is_integer(connected_cluster) ->
         tenant = Cache.get_tenant_by_external_id(external_id)
-        {:ok, db_conn} = Database.connect(tenant, "realtime_health_check")
-        Process.alive?(db_conn) && GenServer.stop(db_conn)
-        Migrations.run_migrations(tenant)
+        result? = Migrations.run_migrations(tenant)
```
edgurgel, line 101: Should we avoid calling run_migrations if `run_migrations?` return false? It would nice to avoid the rpc call if not necessary

lib/realtime/tenants/migrations.ex
```diff
@@ -161,14 +163,29 @@ defmodule Realtime.Tenants.Migrations do
...
+
+  def start_migration(attrs) do
     supervisor =
-      {:via, PartitionSupervisor, {Realtime.Tenants.Migrations.DynamicSupervisor, tenant.external_id}}
+      {:via, PartitionSupervisor, {Realtime.Tenants.Migrations.DynamicSupervisor, attrs.tenant_external_id}}
 
     spec = {__MODULE__, attrs}
 
-    if Tenants.run_migrations?(tenant) do
+    if Tenants.run_migrations?(attrs.migrations_ran) do
```
edgurgel: Maybe this should happen before the RPC call is made? maybe as part of `run_migrations`?

lib/realtime/tenants.ex
filipecabaco: added the check in run.migrations so it will do before the rpc call

edgurgel (approved):

PR #1664, opened 2025-12-22
feat: beacon

edgurgel:

Create new process group library called Beacon that broadcast only group counts to other nodes. Actual pids are only available to the local node. It also supports custom adapter so that we can use PubSub for the broadcasting (including regional broadcasting). 

The plan is to start using Beacon without any real changes and later on start using the group data for 3 things:

* Connected stats
* Max connected users rate limit
* Connect & PG Changes check for online users for a tenant

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.69.2",
+      version: "2.70.0",
       elixir: "~> 1.18",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,


beacon/lib/beacon/adapter/erl_dist.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Beacon.Adapter.ErlDist do
```
josevalim, line 1: One alternative is to use Erlang's PG, as in Phoenix PubSub: https://github.com/phoenixframework/phoenix_pubsub/blob/v2.2.0/lib/phoenix/pubsub/pg2.ex

The main reason to do so is because if you have heterogeneous nodes, not all of them will be running Beacon, and PG will make it so you message only nodes you care about.

beacon/lib/beacon/partition.ex
```diff
@@ -0,0 +1,148 @@
+defmodule Beacon.Partition do
```
josevalim, line 1: I may be wrong but it feels this module is slightly reimplementing Elixir's Registry? Have you consider using the Registry instead? The main reason is that it has been battle tested for performance, it has ETS partitioning for scalability, etc. :) And if a feature is missing, we can gladly add it!

beacon/lib/beacon/scope.ex
```diff
@@ -0,0 +1,200 @@
...
+      peers: %{}
+    ]
+  end
+
+  @spec start_link(atom, Keyword.t()) :: GenServer.on_start()
+  def start_link(scope, opts \\ []), do: GenServer.start_link(__MODULE__, [scope, opts])
+
+  @impl true
+  def init([scope, opts]) do
+    :ok = :net_kernel.monitor_nodes(true)
```
josevalim, line 76: Keep in mind there is an annoying pitfall with monitor nodes, in that the node may be up, but not the Beacon service running on the node! So my suggestion is to not rely on `nodeup` events and instead, when the `Beacon.Scope` process is ready, it broadcasts to everyone.

beacon/lib/beacon/supervisor.ex
```diff
@@ -0,0 +1,53 @@
...
+  @impl true
+  def init([scope, partitions, opts]) do
+    children =
+      for i <- 0..(partitions - 1) do
+        partition_name = partition_name(scope, i)
+
+        ^partition_name =
+          :ets.new(partition_name, [:set, :public, :named_table, read_concurrency: true])
+
+        %{id: i, start: {Beacon.Partition, :start_link, [scope, partition_name]}}
```
josevalim: I understand the idea here is to not lose data if the partition process crashes but I am not sure if I would do this. First of all, the process is not meant to crash. Second, if the process crashes, you don't know if the root cause was because the ETS table got corrupted! So reusing the ETS table will now make this process constantly crash, and eventually the whole node to blow up. IIRC, the Registry links itself to monitored processes, so in case there is a bug in the Registry, every process it is "watching" crashes too.

josevalim:
PS: I will do another pass on the code after the first around of feedback, just to make sure I didn't miss anything important. Distributed code always deserves a couple rounds of looks!

edgurgel:
Oi, @josevalim ! Feliz natal! 

First I would like to say I'm sorry for not providing enough context in this PR description. I will try to fix this now:

Today we use `syn` groups to essentially know how many WebSockets a tenant has globally. So the aim here globally is many thousands of groups that are constantly changing non-stop.

`syn` broadcasts "join" and "leave" messages to all nodes for every single WebSocket that joins. This can be very chatty specially because we actually don't care about the actual PIDs. We only care about the group size so if a WebSocket quickly joins and leaves then another one joins and leaves `syn` will broadcast 4 messages while the count has not really changed in this brief moment.

We decided to build Beacon because we wanted to know how many processes belong to each group globally. We also wanted to keep some features that `syn` has:

* Support a subset of nodes to be part of the cluster although this will be more useful in the future as right now it's all global;
* Be resilient to failures and not cause existing WebSockets to crash. 

We also wanted to have some control how messages are sent to the cluster as we wanted to use `Phoenix.PubSub` with our `gen_rpc` adapter that does not use erl dist, compresses messages and does "regional broadcasting" (send a message to a node in each region and they broadcast to the rest of the nodes in the region).

It obviously also must be able to quickly answer how many processes belong to a group (locally and globally).

beacon/lib/beacon/adapter/erl_dist.ex
edgurgel: Oh yeah that's a fair point! We actually want to use our `gen_rpc` `Phoenix.PubSub` adapter and this ErlDist is there just as the baseline for the library that we intend to extract out of this repo soon.

And we want to use PubSub for now because at least in the short term all nodes will have the same `:users` Beacon scope. This might not be true later on and we might have a different adapter that could potentially use `pg` for the membership as you mentioned.

beacon/lib/beacon/partition.ex
edgurgel: It definitely borrows the partitioned idea from Elixir's Registry! 

We broadcast the group counts from time to time so that every node eventually has the global counts available. So we devised an ETS table that would be very quick to reply how many processes belong to a certain group and could quickly grab all groups and their counts. That's why the ETS table object is:

`{group, MapSet.t(pid), counter}`

A simple `lookup_element` can grab the counter for a certain group and a select only on `group` and `counter` can get me all I need in a single pass.

It wouldn't be fair to expect the Elixir Registry to adapt to these very specific needs I think?

beacon/lib/beacon/scope.ex
edgurgel: I think this needs to be here for the case when a node disconnects and reconnects.

We get a `nodedown` and then a `nodeup.` In this case, from what I understand, both processes have already started but they lost contact. When `nodedown` shows up we don't know if the process will ever come back so we forget all about that node. Once `nodeup` comes up we need to announce that we belong to a scope to know if this "new" node also belongs to this scope. 

That's how `syn` and `pg` do it as well.

beacon/lib/beacon/supervisor.ex
edgurgel: Oh yup we did this because that's how `syn` does it and it has saved us a few times when a node had network issues and kept reconnecting for a while. `syn` would keep trying to sync sending back and forth messages etc until eventually it reached the configured max heap size. `syn`'s process responsible to control the groups kept restarting until the network finally stabilised. 

We had issues with new WebSockets joining but we the WebSockets that were up-and-running were not affected and everything else worked.

I didn't account for an ETS table being corruped because I never heard about it before TBH 😆 I will need to do some research on that! Thanks for bringing this up.

Essentially we would like to minimise damage if Beacon has a bug or it's misbehaving due to some issue we didn't predict (like using too much memory somehow as syn did).

edgurgel:
> Hi @edgurgel! This is looking great and I have added some comments.
> 
> Keep in mind that this could be somewhat implemented with Phoenix.Tracker: you have a custom group and one counter process per node that periodically updates its metadata with the number of entries it sees. This means you can reuse all of PubSub/Tracker, its messaging and updates, while still benefiting from periodical updates. The Tracker approach should be consideraly fewer lines of code, but perhaps there are additional reasons for going ahead with Beacon!

I have to admit I didn't even think about checking Phoenix.Tracker. I will definitely check it out next week!

beacon/lib/beacon/partition.ex
josevalim: Oh, the `counter` was the piece of the puzzle I was missing!

The biggest different between this and Elixir's Registry is that the Registry doesn't require talking to a process to subscribe, which could otherwise become a bottleneck (i.e. the Registry is decentralized). You can also at some point have very large MapSets, which may eventually lead to: https://github.com/discord/sorted_set_nif

In case you want to use a registry-like design, I would do it with two ETS tables:

1. One table keeps all entries in a group (i.e. no MapSet) of type duplicate_key, the other simply keeps group counters
2. The GenServer's job is to increment the counter and setup monitors, and remove entries from the ETS table and decrease the counter when the monitor crashes, but it no longer needs to keep or update a MapSet. Especially because the whole MapSet is copied in and out of ETS every time you want to do an operation on it

The main idea is to reduce how much work happens in a GenServer and use ETS, which is more performant than maps/mapsets. The downside of this approach is that a single process can register itself twice, but I don't think this is an issue, as you can simply ensure you don't register twice. :D

PS: I do think this could be added as part of Elixir's Registry, an indexed group count may be a somewhat common need (and it comes with the benefit that this would then also be available for Phoenix.PubSub).

beacon/lib/beacon/scope.ex
josevalim: Oh, this is a very good call out! Phoenix Tracker doesn't do it and I think it is fine there for two reasons:

1. The node will eventually broadcast itself to everyone anyway
2. Phoenix.Tracker doesn't immediately react to node down anyway, as you don't want a network partition of 5s to cause all numbers to drastically flutuate

In any case, this is fine as is. I'd only document your rationale above with some code comments. Document the messages it receives and how you intend to use them!

beacon/lib/beacon/supervisor.ex
josevalim: > I didn't account for an ETS table being corruped because I never heard about it before TBH

It can't corrupt by itself, it corrupts when you store bad data in it. For example, imagine you make a mistake and you put a List instead of a MapSet. Now it will be forever broken! In any case, it is completely fine for you to leave the code like this :)

josevalim:
> I have to admit I didn't even think about checking Phoenix.Tracker. I will definitely check it out next week!

FWIW, I believe Phoenix.Tracker will immediately notify changes to metadata of a given process, so it won't fullfil all of your criteria above. Perhaps a feature could be added to postpone that until the next broadcast though, but I am not sure how trivial or complex that would be.

edgurgel:
Thanks, @josevalim! I went with your advice and now each Partition holds two ETS tables. 

Because I'm already "paying" the price of doing a GenServer call I ended up using a `set` ETS table for entries and just do a lookup before inserting to ensure no duplicates.

If any crash happens the entries table is used as source of truth.

If we for whatever reason crash before the counters table is updated it won't be a problem.

beacon/lib/beacon/partition.ex
```diff
@@ -78,21 +93,15 @@ defmodule Beacon.Partition do
...
-          # Already being tracked
-          {:reply, :ok, state}
-        else
-          new_pids = MapSet.put(pids, pid)
-          :ets.insert(state.name, {group, new_pids, counter + 1})
-          ref = Process.monitor(pid, tag: {:DOWN, group})
-          monitors = Map.put(state.monitors, {group, pid}, ref)
-          {:reply, :ok, %{state | monitors: monitors}}
-        end
+    case :ets.lookup(state.entries_table, {group, pid}) do
```
josevalim: You can use :ets.insert_new, instead of doing a lookup + insert.

beacon/lib/beacon/partition.ex
```diff
@@ -22,54 +23,68 @@ defmodule Beacon.Partition do
...
+      |> Enum.reduce({%{}, %{}}, fn {{group, pid}}, {monitors_acc, counts_acc} ->
         ref = Process.monitor(pid, tag: {:DOWN, group})
-        {{group, pid}, ref}
-      end
+        monitors_acc = Map.put(monitors_acc, {group, pid}, ref)
+        counts_acc = Map.update(counts_acc, group, 1, &(&1 + 1))
+        {monitors_acc, counts_acc}
+      end)
+
+    :ets.insert(state.name, Enum.to_list(counts))
```
josevalim: There is a chance building this directly on ETS and using `:ets.update_counter` will be quite faster than maps for 1k+ entries. :)

josevalim:
@edgurgel I have dropped two last comments in the commit: https://github.com/supabase/realtime/pull/1664/commits/8a156f5a0a23ec9ec226ef88a1f01c54f325a073

Btw, if both tables are set, you will likely be fine with a single table anyway, as long as you namespace the counts. But two tables are also fine. Completely your call.

beacon/lib/beacon/partition.ex
edgurgel: Oh yup completely forgot about insert_new 🤦 . Fixed!

beacon/lib/beacon/partition.ex
edgurgel: I should've benchmarked 😆 . update_counter is faster and uses less memory. It only gets slower than single insert when the number of entries is huge. Updated!

edgurgel:
Thanks for the review again! I've updated the code based on your comments 🙇

> Btw, if both tables are set, you will likely be fine with a single table anyway, as long as you namespace the counts. But two tables are also fine. Completely your call.

Yeah good point. I thought about it but I want the counters table to have the `{group, counter}` information ready to be fetched to broadcast to other scopes so it's a bit easier to just grab the whole table instead of having to do a select based on the key shape.

Also because I need to do an insert + update_counter there is chance that the counter is not updated and a crash happens.

Then when rebuilding the counters it is easier when I can simply delete all objects from the counters table before rebuilding it.

josevalim:
All good to me!

filipecabaco (approved):

PR #1676, opened 2026-01-22
fix: Load aware node picker using avg 5 minutes

filipecabaco:

Load aware node picker using avg 5 minutes

### DIFF
--- a/test/integration/region_aware_migrations_test.exs
+++ b/test/integration/region_aware_migrations_test.exs
@@ -48,20 +48,23 @@ defmodule Realtime.Integration.RegionAwareMigrationsTest do
     assert tenant.migrations_ran == 0
 
     Realtime.GenRpc
-    |> Mimic.expect(:call, fn called_node, mod, func, args, opts ->
-      assert called_node == node
-      assert mod == Migrations
-      assert func == :start_migration
-      assert opts[:tenant_id] == tenant.external_id
+    |> Mimic.expect(:call, fn
+      called_node, Realtime.Nodes, func, args, opts ->
+        call_original(Realtime.GenRpc, :call, [called_node, Realtime.Nodes, func, args, opts])
 
-      arg = hd(args)
-      assert arg.tenant_external_id == tenant.external_id
-      assert arg.migrations_ran == tenant.migrations_ran
-      assert arg.settings == hd(tenant.extensions).settings
+      called_node, Migrations, func, args, opts ->
+        assert called_node == node
+        assert func == :start_migration
+        assert opts[:tenant_id] == tenant.external_id
 
-      assert opts[:timeout] == 50_000
+        arg = hd(args)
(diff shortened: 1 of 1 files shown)


lib/realtime/nodes.ex
```diff
@@ -92,20 +92,29 @@ defmodule Realtime.Nodes do
...
+  def launch_node(_tenant_id, region, default) do
+    cond do
+      region_nodes(region) == [] ->
         Logger.warning("Zero region nodes for #{region} using #{inspect(default)}")
         default
 
-      regions_nodes ->
-        member_count = Enum.count(regions_nodes)
-        index = :erlang.phash2(tenant_id, member_count)
+      uptime_ms() < Application.fetch_env!(:realtime, :node_balance_uptime_threshold_in_ms) ->
```
edgurgel: I was expecting to check the uptime of the node we are requesting load and not the one that is sending the request. Cause we dont want to use the load of a "new node" right? 🤔 

So they I thought was:

* Request load from 2 random nodes
* If they both returned {:ok, their_load5} we get the min and pick the node
* If one of them return {:error, :new_node} (or something similar) we just pick a random node of the two or from the whole region

WDYT?

filipecabaco: ok will review the approach as I want to keep it simple and avoid too much of a back and forward

lib/realtime/nodes.ex
```diff
@@ -92,20 +92,41 @@ defmodule Realtime.Nodes do
   the `default` node given.
   """
   @spec launch_node(String.t(), String.t() | nil, atom()) :: atom()
-  def launch_node(tenant_id, region, default) do
+  def launch_node(_tenant_id, region, default) do
```
edgurgel: Should we update this function to not receive a tenant_id?

config/test.exs
```diff
@@ -41,7 +41,8 @@ config :realtime,
   api_jwt_secret: System.get_env("API_JWT_SECRET", "secret"),
   metrics_jwt_secret: "test",
   prom_poll_rate: 5_000,
-  request_id_baggage_key: "sb-request-id"
+  request_id_baggage_key: "sb-request-id",
+  node_balance_uptime_threshold_in_ms: 100
```
edgurgel: BTW you could leave this a high number so that almost every single tests won't try to use the load-aware picker. 

And just on the test cases you want you can set this to 0. 

WDYT?

mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.73.1",
+      version: "2.73.2",
```
edgurgel: I would argue that this is a feature so minor bump but not too worried either way 🤷

edgurgel (approved):

PR #1682, opened 2026-01-26
chore: tackle flakytests

filipecabaco:

* Fix up region aware migration mocks
* Block tests until containers are ready to be used

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -11,9 +11,7 @@ import Config
 # are secured by default in production.
 presence = System.get_env("PRESENCE", "false") == "false"
 
-config :realtime,
-  presence: presence,
-  max_gen_rpc_clients: 5
+config :realtime, presence: presence
 
 config :realtime, RealtimeWeb.Endpoint,
   http: [port: System.get_env("PORT", "4000"), compress: true],
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -155,8 +155,7 @@ if config_env() != :test && run_janitor? do
     janitor_chunk_size: janitor_chunk_size,
     janitor_run_after_in_ms: janitor_run_after_in_ms,
     janitor_children_timeout: janitor_children_timeout,
-    janitor_schedule_timer: janitor_schedule_timer,
-    max_gen_rpc_clients: max_gen_rpc_clients
+    janitor_schedule_timer: janitor_schedule_timer
 end
 
 default_cluster_strategy =
@@ -298,7 +297,8 @@ if config_env() != :test do
     db_enc_key: System.get_env("DB_ENC_KEY"),
     region: region,
     prom_poll_rate: Env.get_integer("PROM_POLL_RATE", 5000),
-    slot_name_suffix: slot_name_suffix
+    slot_name_suffix: slot_name_suffix,
+    max_gen_rpc_clients: max_gen_rpc_clients
 end
 
 # Setup Production
--- a/lib/realtime/gen_rpc.ex
+++ b/lib/realtime/gen_rpc.ex
@@ -136,6 +136,7 @@ defmodule Realtime.GenRpc do
     end
   end
 
+  # Not using :gen_rpc.multicall here because we can't see the actual results on errors
   @doc """
   Evaluates apply(mod, func, args) on all nodes
 
@@ -152,7 +153,12 @@ defmodule Realtime.GenRpc do
     key = Keyword.get(opts, :key, nil)
 
     nodes = rpc_nodes([node() | Node.list()], key)
-
+    # Latency here is the amount of time that it takes for this node to gather the result.
+    # If one node takes a while to reply the remaining calls will have at least the latency reported by this node
+    # Example:
+    # Node A, B and C receive the calls in this order
+    # Node A takes 500ms to return on nb_yield
+    # Node B and C will report at least 500ms to return regardless how long it took for them to actually reply back
     results =
       nodes
       |> Enum.map(&{&1, :erlang.monotonic_time(), async_call(&1, mod, func, args)})
@@ -225,6 +231,9 @@ defmodule Realtime.GenRpc do
 
   defp default_rpc_timeout, do: Application.get_env(:realtime, :rpc_timeout, 5_000)
 
+  # Here we run the async_call on all nodes using gen_rpc except the local node
+  # This is because gen_rpc does not have a bypass for local node on multicall
+  # For the local node we use rpc instead
   defp async_call({node, _}, mod, func, args) when node == node(), do: :rpc.async_call(node, mod, func, args)
   defp async_call(node, mod, func, args), do: :gen_rpc.async_call(node, mod, func, args)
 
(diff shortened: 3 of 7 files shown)


config/runtime.exs
```diff
@@ -156,7 +155,8 @@ if config_env() != :test && run_janitor? do
     janitor_chunk_size: janitor_chunk_size,
     janitor_run_after_in_ms: janitor_run_after_in_ms,
     janitor_children_timeout: janitor_children_timeout,
-    janitor_schedule_timer: janitor_schedule_timer
+    janitor_schedule_timer: janitor_schedule_timer,
+    max_gen_rpc_clients: max_gen_rpc_clients
```
edgurgel: why move this under `run_janitor?` 🤔

lib/realtime/gen_rpc.ex
```diff
@@ -226,11 +220,11 @@ defmodule Realtime.GenRpc do
   # Using phash2 to ensure the same key and the same client per node
   defp rpc_node(node, key), do: {node, :erlang.phash2(key, max_clients()) + 1}
 
+  defp unwrap_reason({:unknown_error, {{:badrpc, reason}, _}}), do: reason
+  defp unwrap_reason(reason), do: reason
+
   defp default_rpc_timeout, do: Application.get_env(:realtime, :rpc_timeout, 5_000)
 
-  # Here we run the async_call on all nodes using gen_rpc except the local node
```
edgurgel, line 231: 🤔 Why remove these comments?

config/runtime.exs
filipecabaco: it was more the not test flag

lib/realtime/gen_rpc.ex
```diff
@@ -153,20 +153,14 @@ defmodule Realtime.GenRpc do
 
     nodes = rpc_nodes([node() | Node.list()], key)
 
-    # Latency here is the amount of time that it takes for this node to gather the result.
-    # If one node takes a while to reply the remaining calls will have at least the latency reported by this node
-    # Example:
-    # Node A, B and C receive the calls in this order
-    # Node A takes 500ms to return on nb_yield
-    # Node B and C will report at least 500ms to return regardless how long it took for them to actually reply back
```
edgurgel, line 161: why remove these comments? 🤔

lib/realtime/gen_rpc.ex
filipecabaco: will add back

lib/realtime/gen_rpc.ex
```diff
@@ -134,8 +136,6 @@ defmodule Realtime.GenRpc do
     end
   end
 
-  # Not using :gen_rpc.multicall here because we can't see the actual results on errors
```
edgurgel, line 137: ?

config/dev.exs
```diff
@@ -12,7 +12,8 @@ import Config
 presence = System.get_env("PRESENCE", "false") == "false"
 
 config :realtime,
-  presence: presence
+  presence: presence,
+  max_gen_rpc_clients: 5
```
edgurgel: why is this necessary? the config/runtime.exs already sets 5 as the default for all environments 🤔

test/integration/region_aware_migrations_test.exs
```diff
@@ -49,17 +49,21 @@ defmodule Realtime.Integration.RegionAwareMigrationsTest do
 
     Realtime.GenRpc
     |> Mimic.expect(:call, fn called_node, mod, func, args, opts ->
-      assert called_node == node
-      assert mod == Migrations
-      assert func == :start_migration
-      assert opts[:tenant_id] == tenant.external_id
+      case {mod, func} do
+        {Realtime.Nodes, :node_load} ->
+          call_original(Realtime.GenRpc, :call, [called_node, mod, func, args, opts])
```
edgurgel: This function doesn't exist here right? Is this from another PR?

test/realtime/gen_rpc_test.exs
```diff
@@ -365,7 +365,7 @@ defmodule Realtime.GenRpcTest do
         end)
 
       assert log =~
-               ~r/project=123 external_id=123 \[error\] ErrorOnRpcCall: %{\s+error: :econnrefused,\s+mod: Map,\s+func: :fetch,\s+target:\s+:"#{node}"/
+               ~r/project=123 external_id=123 \[error\] ErrorOnRpcCall: %{\s*error: :econnrefused,\s+mod: Map,\s+func: :fetch,\s+target:\s+:"#{node}"/
```
edgurgel: were these tests failing before? 🤔

test/support/containers.ex
```diff
@@ -23,6 +23,21 @@ defmodule Containers do
...
+    wait_for_pool(timeout)
+    pool_size = max_cases + 2
+
+    Enum.each(1..pool_size, fn _ ->
+      :poolboy.transaction(
+        Containers.Pool,
+        fn container -> container |> Container.port() |> check_container_ready() end,
+        timeout
+      )
+    end)
```
edgurgel: I don't think you have any guarantee that you will get all connections from the pool like this.

Also `Container.port/1` will not even reply before Postgres is ready anyway:

https://github.com/supabase/realtime/blob/main/test/support/containers/container.ex#L34-L40

The `handle_continue` guarantees that no `handle_call` will be processed before it has finished the `handle_continue` bit.

So when we do checkout we always call `Container.port/1` which means that it won't serve you a container that is not ready to reply to requests:

https://github.com/supabase/realtime/blob/71bdf43b49c3ac94bc8ccd8e813be4680e40d6a5/test/support/containers.ex#L106-L109

test/support/containers.ex
```diff
@@ -33,6 +48,22 @@ defmodule Containers do
     {:ok, %{existing_containers: existing_containers, ports: available_ports}, {:continue, {:pool, max_cases}}}
   end
 
+  defp wait_for_pool(timeout, started_at \\ System.monotonic_time(:millisecond)) do
```
edgurgel: We would get an error if Containers.Pool can't reply on poolboy. I don't think I've ever seen an error on tests where the pool is not up. when poolboy returns from the start_link it always replies 🤔 

https://github.com/supabase/realtime/blob/71bdf43b49c3ac94bc8ccd8e813be4680e40d6a5/test/test_helper.exs#L16

test/support/containers.ex
```diff
@@ -89,7 +120,7 @@ defmodule Containers do
           port
       end
 
-    check_container_ready(name)
+    check_container_ready(port)
```
edgurgel: I think this is redundant because Container itself checks to be ready:

https://github.com/supabase/realtime/blob/71bdf43b49c3ac94bc8ccd8e813be4680e40d6a5/test/support/containers/container.ex#L42-L51

config/runtime.exs
edgurgel: Right... but what are these config changing in the end? They all seem to set to 5 like before? 🤔

test/support/clustered.ex
```diff
@@ -120,4 +129,27 @@ defmodule Clustered do
...
+  defp wait_for_port(host, port, attempts, delay_ms) do
+    case :gen_tcp.connect(host, port, [:binary, active: false], 200) do
+      {:ok, socket} ->
+        :ok = :gen_tcp.close(socket)
+        :ok
+
+      {:error, _reason} ->
+        Process.sleep(delay_ms)
+        wait_for_port(host, port, attempts - 1, delay_ms)
+    end
```
edgurgel, line 153: nice

PR #1690, opened 2026-01-30
fix: normalize booleans on join payloads

filipecabaco:

To ensure better compatibility with existing users we will infer intent from non boolean values by normalizing and down casing boolean strings

e.g. `'False'` will be considered `false` boolean value


mix.exs
```diff
@@ -4,7 +4,7 @@ defmodule Realtime.MixProject do
   def project do
     [
       app: :realtime,
-      version: "2.74.1",
+      version: "2.74.2",
```
edgurgel, line 7: You don't need to update the mix.exs version

PR #1693, opened 2026-02-02
fix: adds shutdown endpoint

filipecabaco:

Adds a shutdown endpoint for easier maintenance tasks

### DIFF
--- a/lib/realtime_web/controllers/tenant_controller.ex
+++ b/lib/realtime_web/controllers/tenant_controller.ex
@@ -227,7 +227,6 @@ defmodule RealtimeWeb.TenantController do
   def reload(conn, %{"tenant_id" => tenant_id}) do
     case Api.get_tenant_by_external_id(tenant_id, use_replica?: false) do
       nil ->
-        log_error("TenantNotFound", "Tenant not found")
         {:error, :not_found}
 
       tenant ->
@@ -261,7 +260,6 @@ defmodule RealtimeWeb.TenantController do
   def shutdown(conn, %{"tenant_id" => tenant_id}) do
     case Api.get_tenant_by_external_id(tenant_id, use_replica?: false) do
       nil ->
-        log_error("TenantNotFound", "Tenant not found")
         {:error, :not_found}
 
       tenant ->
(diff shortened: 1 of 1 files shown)


lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -247,6 +247,42 @@ defmodule RealtimeWeb.TenantController do
...
+  )
+
+  def shutdown(conn, %{"tenant_id" => tenant_id}) do
+    case Api.get_tenant_by_external_id(tenant_id, use_replica?: false) do
+      nil ->
+        log_error("TenantNotFound", "Tenant not found")
+
+        conn
+        |> put_status(404)
+        |> render("not_found.json", tenant: nil)
```
edgurgel: I think you can simply return `{:error, :not_found}` and the fallback controller will deal with it:

https://github.com/supabase/realtime/blob/083d5351a9307222bd7f625fd6b6fd5f4e0271b3/lib/realtime_web/controllers/fallback_controller.ex#L13-L18

lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -247,6 +247,42 @@ defmodule RealtimeWeb.TenantController do
...
+
+      tenant ->
+        Connect.shutdown(tenant.external_id)
+        SocketDisconnect.distributed_disconnect(tenant.external_id)
```
edgurgel: Do we need to force disconnect?

filipecabaco: will remove for now and focus only on restart conns

lib/realtime_web/controllers/tenant_controller.ex
filipecabaco: will fix

lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -234,10 +228,7 @@ defmodule RealtimeWeb.TenantController do
     case Api.get_tenant_by_external_id(tenant_id, use_replica?: false) do
       nil ->
         log_error("TenantNotFound", "Tenant not found")
```
edgurgel: should we remove this line if we are logging on fallback now?

lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -277,10 +300,7 @@ defmodule RealtimeWeb.TenantController do
 
       {:error, :tenant_not_found} ->
         log_error("TenantNotFound", "Tenant not found")
```
edgurgel: same here

mix.exs
edgurgel, line 7: remember to remove this :feels_good:

PR #1697, opened 2026-02-03
fix: add proper timeout error; improve backoff logic on replication recovery

filipecabaco:

* Properly log out the timeout error on replication connection
* Improve backoff policy to include exponential backoff to provide more time to db instance to recover

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -36,6 +36,7 @@ defmodule Realtime.Tenants.Connect do
           replication_connection_pid: pid(),
           replication_connection_reference: reference(),
           backoff: Backoff.t(),
+          replication_connection_attempts: non_neg_integer(),
           check_connected_user_interval: non_neg_integer(),
           connected_users_bucket: list(non_neg_integer()),
           check_connect_region_interval: non_neg_integer()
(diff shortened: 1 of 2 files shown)


lib/realtime/tenants/connect.ex
```diff
@@ -460,4 +476,10 @@ defmodule Realtime.Tenants.Connect do
       log_error("StartReplicationFailed", error)
       {:error, state}
   end
+
+  defp calculate_recovery_backoff(attempts) do
```
edgurgel: Would you mind a comment on what's the expected growth on the backoff here?

lib/realtime/tenants/connect.ex
```diff
@@ -350,29 +364,40 @@ defmodule Realtime.Tenants.Connect do
...
          {:ok, state} <- start_replication_connection(state) do
-      {:noreply, state}
+      {:noreply, %{state | backoff: Backoff.reset(backoff)}}
     else
-      _ ->
-        log_error("ReplicationConnectionRecoveryFailed", "Replication connection recovery failed")
-        Process.send_after(self(), :recover_replication_connection, @replication_recovery_backoff)
-        {:noreply, state}
+      %{num_rows: _} ->
+        {:noreply, %{state | backoff: Backoff.reset(backoff)}}
```
edgurgel: So if we got num_rows different than 0 but this Connect proces does not own the ReplicationConnection process what does this mean in the end?

PR #1698, opened 2026-02-03
fix: provide more granular health check response

filipecabaco:

Provide finer granularity of our healtcheck to ensure users have a better view if ReplicationConnection isn't working

### DIFF
--- a/lib/realtime/tenants/connect.ex
+++ b/lib/realtime/tenants/connect.ex
@@ -13,7 +13,6 @@ defmodule Realtime.Tenants.Connect do
 
   alias Realtime.Api.Tenant
   alias Realtime.GenCounter
-  alias Realtime.GenRpc
   alias Realtime.RateCounter
   alias Realtime.Rpc
   alias Realtime.Tenants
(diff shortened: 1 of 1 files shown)


lib/realtime/tenants.ex
```diff
@@ -94,13 +103,21 @@ defmodule Realtime.Tenants do
...
            connected_cluster: connected_cluster,
            region: region,
            node: node
          }}
     end
   end
 
+  defp replication_connected?(external_id) do
+    case ReplicationConnection.whereis(external_id) do
+      pid when is_pid(pid) -> Process.alive?(pid)
```
edgurgel: A couple of problems here:

* `ReplicationConnection.whereis` uses Registry so this won't work when this is not the node where ReplicationConnection is running. An RPC call is needed based on `Connection`'s location

* Do we need to check if it's alive? I'm asking because `Process.alive?/1` can only be run on a local pid. If we are using Registry or Syn it won't return us a pid that is not alive AFAIK.

filipecabaco: oh damn I thought it was using syn! 💀 and good call on the process alive

edgurgel: Oh and not sure what is the impact of adding an RPC call to the health check. It might add a few hundred milliseconds or more

filipecabaco: check the implementation.

honestly it should be fine as long as we provide a better output for the health check I think it's worth it

lib/realtime/tenants/connect.ex
```diff
@@ -186,6 +187,25 @@ defmodule Realtime.Tenants.Connect do
     end
   end
 
+  @spec whereis_replication_connection(binary()) :: pid() | nil
+  def whereis_replication_connection(tenant_id) do
+    case :syn.lookup(__MODULE__, tenant_id) do
+      {pid, _} when is_pid(pid) -> GenRpc.call(node(pid), ReplicationConnection, :whereis, [tenant_id], [])
```
edgurgel: Do we still need an rpc call here?

filipecabaco: nope will remove

edgurgel (approved):

PR #1706, opened 2026-02-09
feat: implement peek on walrus query

filipecabaco:

To reduce overall load of the query we are now peeking before pulling so we avoid running the more expensive query and get faster feedback


filipecabaco:
bad branch name

PR #1707, opened 2026-02-09
feat: implement peek on walrus query

filipecabaco:

To reduce overall load of the query we are now peeking before pulling so we avoid running the more expensive query and get faster feedback

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL logical replication polling capability with optimized detection of empty changes
  * Enhanced replication polling behavior with immediate rescheduling and backoff reset for faster response times
  * Implemented Row-Level Security (RLS) support for replication change tracking

* **Tests**
  * Added integration tests validating logical replication functionality, RLS filtering, and sentinel response behavior

### DIFF
--- a/lib/extensions/postgres_cdc_rls/replication_poller.ex
+++ b/lib/extensions/postgres_cdc_rls/replication_poller.ex
@@ -95,6 +95,11 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
     record_list_changes_telemetry(time, tenant_id)
 
     case handle_list_changes_result(list_changes, subscribers_nodes_table, tenant_id, rate_counter_args) do
+      {:ok, :peek_empty} ->
+        Backoff.reset(backoff)
+        poll_ref = Process.send_after(self(), :poll, poll_interval_ms)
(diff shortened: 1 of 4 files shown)


v0idpwn (approved):

edgurgel (changes requested):
I wonder if we are better off to use `pg_logical_slot_get_changes`, and use that to return the "sentinel row" that will always have the number of changes we pulled? This way we won't interact twice to the slot

Claude suggested to use AS MATERIALIZED to not call it twice

```
        -- MATERIALIZED ensures pg_logical_slot_get_changes is called exactly once
        w2j AS MATERIALIZED (
          SELECT x.*, pub.w2j_add_tables
          FROM pub,
               pg_logical_slot_get_changes(
                 slot_name, null, max_changes,
                 'include-pk', 'true',
                 'include-transaction', 'false',
                 'include-timestamp', 'true',
                 'include-type-oids', 'true',
                 'format-version', '2',
                 'actions', pub.w2j_actions,
                 'add-tables', pub.w2j_add_tables
               ) x
        ),

```

edgurgel:
solved differently here https://github.com/supabase/realtime/pull/1803

PR #1708, opened 2026-02-09
feat: increase pooling rate on no changes

filipecabaco:

To reduce overall load of the system we will start to reduce pooling speed when no changes are not registered


edgurgel:
Solved differently here https://github.com/supabase/realtime/pull/1803

PR #1710, opened 2026-02-09
fix: per tenant presence limit settings

filipecabaco:

Set presence limits per tenant and keep the defaults being a env var value

### DIFF
--- a/lib/realtime_web/open_api_schemas.ex
+++ b/lib/realtime_web/open_api_schemas.ex
@@ -113,7 +113,7 @@ defmodule RealtimeWeb.OpenApiSchemas do
             },
             max_client_presence_events_per_window: %Schema{
               type: :number,
-              description: "Maximum client presence events per second (overrides environment default when set)",
+              description: "Maximum client presence events (overrides environment default when set)",
               nullable: true
             },
             client_presence_window_ms: %Schema{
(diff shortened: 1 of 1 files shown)


lib/realtime/api/tenant.ex
```diff
@@ -30,6 +30,8 @@ defmodule Realtime.Api.Tenant do
     field(:private_only, :boolean, default: false)
     field(:migrations_ran, :integer, default: 0)
     field(:broadcast_adapter, Ecto.Enum, values: [:phoenix, :gen_rpc], default: :gen_rpc)
+    field(:max_client_presence_events_per_second, :integer)
```
edgurgel: should we call this `:max_client_presence_events_per_window` ? IT's not per second in this sense if we have the window as configuration as well

lib/realtime_web/open_api_schemas.ex
```diff
@@ -111,6 +111,16 @@ defmodule RealtimeWeb.OpenApiSchemas do
               type: :number,
               description: "Maximum payload size in KB"
             },
+            max_client_presence_events_per_window: %Schema{
+              type: :number,
+              description: "Maximum client presence events per second (overrides environment default when set)",
```
edgurgel: Description still says per second!

lib/realtime_web/open_api_schemas.ex
```diff
@@ -214,6 +224,16 @@ defmodule RealtimeWeb.OpenApiSchemas do
             }
           }
         },
+        max_client_presence_events_per_window: %Schema{
+          type: :number,
+          description: "Maximum client presence events per second (overrides environment default when set)",
```
edgurgel: Same here

edgurgel (approved):

PR #1714, opened 2026-02-12
fix: implement migration reconciliation

filipecabaco:

In certain scenarios we end up with a difference between our metrics and the users metrics when it comes to migrations so this will self correct by checking the real count the user has in their database and we update our records

## Release Notes

* **New Features**
  * Tenant database connections now track and report the count of executed migrations.
  * Implemented automatic reconciliation that identifies and corrects discrepancies between cached migration counts and actual database state during tenant initialization.

* **Tests**
  * Added comprehensive test coverage for migration count tracking and reconciliation scenarios.

### DIFF
--- a/lib/realtime/database.ex
+++ b/lib/realtime/database.ex
@@ -107,11 +107,9 @@ defmodule Realtime.Database do
         if requirement < available_connections do
           {:ok, conn, migrations_ran}
         else
-          log_error(
-            "DatabaseLackOfConnections",
-            "Only #{available_connections} available connections. At least #{requirement} connections are required."
-          )
-
+          msg = "Only #{available_connections} available connections. At least #{requirement} connections are required."
+          log_error("DatabaseLackOfConnections", msg)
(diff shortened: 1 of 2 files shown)


lib/realtime/database.ex
```diff
@@ -100,34 +100,56 @@ defmodule Realtime.Database do
...
-              log_error(
-                "DatabaseLackOfConnections",
-                "Only #{available_connections} available connections. At least #{requirement} connections are required."
-              )
-
-              {:error, :tenant_db_too_many_connections}
-            end
-
-          {:error, e} ->
-            Process.exit(conn, :kill)
```
mentels, line 123: Any reason to remove this line?

mentels (approved):
LGTM, just one q

lib/realtime/database.ex
filipecabaco: good catch will add back

mentels (approved):

Issue #486, opened 2023-02-16
RFC: Allow filter chaining


selbyk:
# Summary
[summary]: #summary

Allow filter chaining

# Rationale
[rationale]: #rationale

Subscriptions to compound keys or multiple criteria would be useful in some scenarios

# Design
[design]: #design

An dense explanation in sufficient detail that someone familiar with the
project could implement the feature. Specifics and corner cases should be covered.

# Examples
[examples]: #examples

```
filters: [`store_id=eq.${store.id}`,`product_id=eq.${product.id}`]
filter: `store_id=eq.${store.id},product_id=eq.${product.id}`
filter: `store_id=eq.${store.id},.or(product_id=eq.${product.id},product_id=eq.${productTwo.id})`
```

# Drawbacks
[drawbacks]: #drawbacks

What are the negative trade-offs?

# Alternatives
[alternatives]: #alternatives

What other solutions have been considered?

# Unresolved Questions
[unresolved]: #unresolved-questions

Not sure what work would be involved
Filter syntax probably needs work

sathishkannan162:
This feature would prove to be very useful. Looking forward to see this feature implemented.

olirice:
> any plans for this or is supabase looking for community contributions for this one?
@filipecabaco could you please weigh in on this one ^

fwiw, walrus does support [arrays of filters](https://github.com/supabase/walrus/blob/ef516d68133c4e716bea5c82c8376880ba382c64/test/sql/test_integration_filters.sql#L19) but I understand that isn't he only consideration

amerryma:
Hi there, just keeping this thread fresh. I'm curious though, does supabase/realtime use walrus under the hood and it's just a matter of hook-up?

amerryma:
Some projects using Refine.dev with Supabase are being affected by this bug, and they're needing to add special one-off logic for Supabase to warn users if they attempt to do something like this using their generic wrapper over realtime logic. I'd love it if they didn't have to build this in at all. Does anyone have any updates on this? 

cc: @olirice @filipecabaco

amerryma:
Any additional thoughts on this or direction for how to work through this?

PR #1718, opened 2026-02-13
Support Refine.dev filter chaining and isnull/notnull filter

M4R7iNP:
* Support filter chaining
* * NOTE: only as AND. OR is not handled
* Support Refine's `is.null` (`operator: "null"`) and `not.is.null` (`operator: "nnull"`)  filters using postgres' `isnull` and `notnull` operators

https://github.com/supabase/realtime/issues/486
https://github.com/refinedev/refine/issues/6360


3ddyBoi:
I need this

edgurgel:
PG Ghanges filters have been updated to support `AND` plus other features: https://supabase.com/blog/postgres-changes-filters-and-column-selection

PR #1726, opened 2026-02-26
fix: support legacy metrics controller

filipecabaco:

support legacy metrics controller

### DIFF
--- a/lib/realtime_web/controllers/legacy_metrics_controller.ex
+++ b/lib/realtime_web/controllers/legacy_metrics_controller.ex
@@ -1,0 +1,73 @@
+defmodule RealtimeWeb.LegacyMetricsController do
+  use RealtimeWeb, :controller
+  require Logger
+  alias Realtime.PromEx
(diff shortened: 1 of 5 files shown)


lib/realtime_web/controllers/legacy_metrics_controller.ex
```diff
@@ -0,0 +1,65 @@
...
+    serve_metrics(conn, [Node.self() | Node.list()], "combined cluster")
+  end
+
+  def region(conn, %{"region" => region}) do
+    serve_metrics(conn, Realtime.Nodes.region_nodes(region), "combined region=#{region}")
+  end
+
+  def get_combined_metrics do
+    bump_max_heap_size()
+    IO.iodata_to_binary(PromEx.get_global_metrics()) <> IO.iodata_to_binary(TenantPromEx.get_metrics())
```
v0idpwn: Probably there's a small win by just using iodata, no?

PR #1785, opened 2026-04-02
fix: rate limit logging based on tenant id and error op code

filipecabaco:

Adding new logging option that rate limits logs based on tenant and error opcode so we can prevent noisy logs at the tenant level and at the error level as a more broad approach to reduce log volume

### DIFF
--- a/lib/realtime_web/channels/realtime_channel/logging.ex
+++ b/lib/realtime_web/channels/realtime_channel/logging.ex
@@ -6,6 +6,10 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
   alias Realtime.Telemetry
   require Logger
 
+  @log_rate_table :log_rate_limiter
+  @log_rate_default_max 10
+  @log_rate_default_window_ms 300_000
+
   defmacro __using__(_opts) do
     quote do
       require Logger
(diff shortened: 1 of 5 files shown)


lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -18,60 +22,86 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
...
+  defp rate_limited?(tenant, code, opts) do
+    max = Keyword.get(opts, :max, @log_rate_default_max)
+    window_ms = Keyword.get(opts, :window_ms, @log_rate_default_window_ms)
+    key = {tenant, code}
+    now = :erlang.monotonic_time(:millisecond)
+    count = :ets.update_counter(@log_rate_table, key, {2, 1}, {key, 0, now})
+
+    case :ets.lookup(@log_rate_table, key) do
+      [{^key, ^count, window_start}] when now - window_start >= window_ms ->
+        :ets.insert(@log_rate_table, {key, 1, now})
```
chasers, line 94: This creates a row per tenant, code, 300,000 milliseconds? With N tenants and N codes how many rows with this have after a day? Are we cleaning it anywhere?

chasers (changes requested):

filipecabaco:
https://github.com/supabase/realtime/pull/1793

PR #1795, opened 2026-04-06
fix: rate limit common errors

filipecabaco:

to reduce log noise, we will start to rate limit ChannelRateLimitReached, RealtimeDisabledForTenant and ConnectionRateLimitReached

### DIFF
--- a/lib/realtime_web/channels/realtime_channel/logging.ex
+++ b/lib/realtime_web/channels/realtime_channel/logging.ex
@@ -16,9 +16,9 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
   @doc """
   Logs an error message
   """
-  @spec log_error(socket :: Phoenix.Socket.t(), code :: binary(), msg :: any(), opts :: keyword()) ::
+  @spec log_error(socket :: Phoenix.Socket.t(), code :: binary(), msg :: any(), opts :: map()) ::
           {:error, %{reason: binary}}
-  def log_error(socket, code, msg, opts \\ []) do
+  def log_error(socket, code, msg, opts \\ %{}) do
     built_msg = build_msg(code, msg)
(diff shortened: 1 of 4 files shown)


lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -93,18 +93,21 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
   defp do_log(%{assigns: %{tenant: tenant}} = socket, level, code, msg, throttle: {max_count, window_ms}) do
```
edgurgel: Not sure that matching on a keyword is a good idea here because if we introduce a new option later this might not match due to how lists match. We either use a map or inside the function call Keyword.get

edgurgel (changes requested):

lib/realtime_web/channels/realtime_channel/logging.ex
```diff
@@ -93,18 +93,21 @@ defmodule RealtimeWeb.RealtimeChannel.Logging do
...
+            1
+
+          n when is_integer(n) ->
+            Cachex.incr!(cache, key)
+        end
+      end)
+
+    if count <= max_count,
+      do: log(socket, level, code, msg),
+      else: emit_telemetry(level, code, tenant)
```
edgurgel: I thought we always wanted to emit telemetry though 🤔 

Should we always have the metric and just sometimes log?

filipecabaco: log itself has the telemetry call, not the best abstraction

lib/realtime_web/channels/realtime_channel/logging.ex
filipecabaco: changing to a map for now

leandrocp (approved):

PR #1809, opened 2026-04-11
fix: add REALTIME_IP_VERSION and auto-detect for ipv4/ipv6

aantti:

Fix

On IPv4-only hosts (e.g. when IPv6 is disabled in grub config), Realtime won't start.

## What is the new behavior?

Add `REALTIME_IP_VERSION` back and also add auto-detect for ipv4/ipv6 stack.

## Additional context

Closes #1750


edgurgel (approved):
LGTM

PR #1810, opened 2026-04-14
fix: livedashboard recon & node info

filipecabaco:

Adds Recon and Node Info views to LiveDashboard
Screenshot 2026-04-14 at 15 55 29

Screenshot 2026-04-14 at 15 55 36


test/e2e/check-local.ts
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env bun
```
edgurgel: Do we want this 🤔 ?

lib/realtime_web/dashboard/node_info.ex
```diff
@@ -0,0 +1,110 @@
...
+            <div class="card-body text-danger small"><%= node_data.error %></div>
+          <% else %>
+            <div class="card-body p-0">
+              <table class="table table-sm table-hover mb-0">
+                <thead><tr><th colspan="2" class="text-muted small ps-3">Realtime</th></tr></thead>
+                <tbody>
+                  <tr><td class="ps-3">Region</td><td><%= node_data.region || "not set" %></td></tr>
+                  <tr><td class="ps-3">Master Region</td><td><%= node_data.master_region || "not set" %></td></tr>
+                  <tr><td class="ps-3">Is Master</td><td><%= node_data.is_master %></td></tr>
+                  <tr><td class="ps-3">Read Replica</td><td><%= node_data.read_replica %></td></tr>
```
edgurgel, line 62: I don't think this module will be different from `Realtime.Repo` and `Realtime.Repo.Replica` because we are not using those specific modules for regions anymore

I think we would need to have the replica host to know which host it's being used

lib/realtime_web/dashboard/recon_trace.ex
```diff
@@ -0,0 +1,827 @@
+defmodule RealtimeWeb.Dashboard.ReconTrace do
```
edgurgel, line 1: Is it using :recon_trace or :erlang.trace? :recon_trace has a rate limit in place so we should definitely try to use that.

edgurgel (changes requested):

lib/realtime_web/dashboard/recon_trace.ex
filipecabaco: fixed

lib/realtime_web/dashboard/node_info.ex
filipecabaco: fixed

test/e2e/check-local.ts
filipecabaco: removed

lib/realtime_web/dashboard/recon_trace.ex
```diff
@@ -0,0 +1,808 @@
...
+    mod = Map.get(params, "mod", "")
+    fun = Map.get(params, "fun", "_")
+    max_calls = Map.get(params, "max_calls", "100")
+    scope = Map.get(params, "scope", "local")
+
+    with {:ok, mod_atom} <- parse_module(mod),
+         {:ok, max_val} <- parse_max_calls(max_calls) do
+      fun_atom = parse_fun(fun)
+      scope_atom = if scope == "global", do: :global, else: :local
+      io_server = spawn(fn -> io_discard_loop() end)
```
edgurgel: Should we use spawn_link here so we don't have a process that might never go away if this process crashes before it can reach line 632

edgurgel (approved):

edgurgel:
nitpick: this is not a fix but a `feat`

filipecabaco:
why? this doesn't break anything and it's just a dashboard that does not impact the main functionalities 🤔

edgurgel:
Well this is not fixing a bug 🤷 But yeah just a nitpick!

PR #1811, opened 2026-04-14
fix: validate track payload

filipecabaco:

Properly handle track payload


lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -119,7 +119,8 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
     %{assigns: %{presence_key: presence_key, tenant_topic: tenant_topic}} = socket
     payload = Map.get(payload, "payload", %{})
 
-    with :ok <- check_track_payload(socket.assigns, payload),
+    with true <- is_map(payload) || {:error, :invalid_payload},
+         :ok <- check_track_payload(socket.assigns, payload),
```
edgurgel: nit: would checking for a map make more sense inside the existing `check_trace_payload/2` ?

lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -146,6 +147,9 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
             {:error, :unable_to_track_presence}
         end
 
+      {:error, :invalid_payload} ->
+        {:error, :invalid_payload}
```
edgurgel: What will this result from the websocket perspective? Will it return an "Unknown error " thingy?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -428,6 +428,10 @@ defmodule RealtimeWeb.RealtimeChannel do
       {:error, :payload_size_exceeded} ->
         shutdown_response(socket, "Track message size exceeded")
 
+      {:error, :invalid_payload} ->
+        log_error(socket, "InvalidPresencePayload", :invalid_payload)
```
edgurgel, line 432: Oh we need to include this error in the README: https://github.com/supabase/realtime#error-operational-codes

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -428,6 +428,10 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        {:reply, {:error, %{reason: "Presence track payload must be a map"}}, socket}
```
edgurgel, line 433: Any idea what our realtime-js library will see when this error happens? I don't remember how we surface such error on `track`

edgurgel (approved):
Added a question and a small doc change but other than that LGTM

lib/realtime_web/channels/realtime_channel.ex
filipecabaco: not sure either will try to verify but overall errors in the clients lib ain't the best from what I recall

PR #1824, opened 2026-04-20
feat: squash realtime migrations

filipecabaco:

Our migrations are currently doing too many operations so we are squashing them and moving the older migrations to be no-op to reduce time to run them.


lib/realtime/tenants/repo/migrations/20260420000000_squash_migrations.ex
```diff
@@ -0,0 +1,824 @@
+defmodule Realtime.Tenants.Migrations.SquashMigrations do
```
leandrocp, line 1: I'm wondering if we should keep the no-op migrations 🤔 

Existing databases would have a `schema_migrations` containing migrations that no longer exist (unless this migration also updates it manually but that's too agressive) but that shouldn't break migrations and the upstream file would just have `def change, do: nil` anyway. Rollback as far as I can tell is not supported already (I see migrations that would fail a rollback).

Unless there's something else I'm not aware of. Wdyt?

leandrocp (commented):
Hey @filipecabaco looks good but I'm wondering if we can clean up a bit more.

lib/realtime/tenants/repo/migrations/20260420000000_squash_migrations.ex
filipecabaco: this is due to the fact that we count the number of migrations to see if things are up to date and since we do this check and keep it in all tenants we would need some special and more complex logic to update all schema_migrations tables and the respective migrations_ran values in our db which would be quite more complex and can potentially be done in the future but for now a more naive approach will yield the desired improvements and even allow us to have the ground works for it (aka we have the squashed migrations done)

leandrocp: Gotcha! Does this check exist to make sure all objects were created?

filipecabaco: and ensure we only run migrations when strictly required as they are expensive

lib/realtime/tenants/repo/migrations/20260420000000_squash_migrations.ex
```diff
@@ -0,0 +1,824 @@
...
+                'ERROR'
+            );
+        END IF;
+    END$$;
+    """)
+
+    execute("""
+    DO $$
+    BEGIN
+        IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'equality_op' AND typnamespace = 'realtime'::regnamespace) THEN
```
edgurgel, line 54: Can we check if the type is the expected type instead of checking if the type exists? I'm asking because if the project stopped at this migration here:

https://github.com/supabase/realtime/blob/ee76679b4904d32f33f42d6eb4da9e12d29baca3/lib/realtime/tenants/repo/migrations/20211116024918_create_realtime_subscription_table.ex#L6-L15

It will have the type BUT it will not have the `in` that was added later.

lib/realtime/tenants/repo/migrations/20260420000000_squash_migrations.ex
```diff
@@ -0,0 +1,824 @@
...
+            );
+        END IF;
+    END$$;
+    """)
+
+    execute("""
+    DO $$
+    BEGIN
+        IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'wal_column' AND typnamespace = 'realtime'::regnamespace) THEN
+            CREATE TYPE realtime.wal_column AS (
```
edgurgel, line 85: Same thing here. We create wal_column and change later so we can't just check if it exists.

https://github.com/supabase/realtime/blob/ee76679b4904d32f33f42d6eb4da9e12d29baca3/lib/realtime/tenants/repo/migrations/20220603231003_add_quoted_regtypes_support.ex#L12-L21



https://github.com/supabase/realtime/blob/ee76679b4904d32f33f42d6eb4da9e12d29baca3/lib/realtime/tenants/repo/migrations/20211116212300_create_realtime_build_prepared_statement_sql_function.ex#L25-L31

Notice that this migration actually does the check I'm asking about: https://github.com/supabase/realtime/blob/ee76679b4904d32f33f42d6eb4da9e12d29baca3/lib/realtime/tenants/repo/migrations/20211116212300_create_realtime_build_prepared_statement_sql_function.ex#L12-L33

filipecabaco:
we will take another approach

PR #1828, opened 2026-04-21
fix: add granularity to unable to set policies errors

filipecabaco:

add granularity to unable to set policies errors

### DIFF
--- a/lib/realtime/tenants/authorization.ex
+++ b/lib/realtime/tenants/authorization.ex
@@ -152,9 +152,8 @@ defmodule Realtime.Tenants.Authorization do
     end
   end
 
-  def get_write_authorizations(db_conn, authorization_context) do
-    get_write_authorizations(%Policies{}, db_conn, authorization_context)
-  end
+  def get_write_authorizations(db_conn, authorization_context),
+    do: get_write_authorizations(%Policies{}, db_conn, authorization_context)
 
   defp handle_policies_result(result, rate_counter) do
     case result do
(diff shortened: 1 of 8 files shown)


lib/realtime/tenants/authorization.ex
```diff
@@ -164,6 +164,15 @@ defmodule Realtime.Tenants.Authorization do
...
         {:error, :rls_policy_error, error}
 
+      {:error, %Postgrex.Error{postgres: %{code: :invalid_parameter_value}} = error} ->
+        {:error, :rls_policy_error, error}
+
+      {:error, %MatchError{term: {:error, %Postgrex.Error{postgres: %{code: :query_canceled}} = error}}} ->
+        {:error, :query_canceled, error}
+
+      {:error, %MatchError{term: {:error, %Postgrex.Error{postgres: %{code: :check_violation, table: "messages"}}}}} ->
+        {:error, :missing_partition}
```
edgurgel: Can we instead of relying on `MatchError` fix the place where this is not matching and send a proper error?

edgurgel (changes requested):

lib/realtime/tenants/authorization.ex
```diff
@@ -109,7 +109,7 @@ defmodule Realtime.Tenants.Authorization do
   Automatically uses RPC if the database connection is not in the same node
   """
   @spec get_write_authorizations(Policies.t(), pid(), __MODULE__.t(), keyword()) ::
```
leandrocp:
```suggestion
  @spec get_write_authorizations(Policies.t(), pid(), t(), keyword()) ::
```

lib/realtime/tenants/authorization.ex
```diff
@@ -109,7 +109,7 @@ defmodule Realtime.Tenants.Authorization do
...
-          {:ok, Policies.t()} | {:error, any()} | {:error, :rls_policy_error, any()}
+          {:ok, Policies.t()} | {:error, any()} | {:error, :rls_policy_error | :query_canceled, any()}
```
leandrocp: `{:error, any()}` here is swallowing any kind of error and we lose the opportunity to make it more explicit, ideally that would be a strict set of errors like:

```elixir
          {:ok, Policies.t()} |
          {:error, :rls_policy_error, any()} |
          :error, :query_canceled, any()} |
          {:error, :missing_partition} |
          {:error, :increase_connection_pool} |
          {:error, :tenant_database_unavailable} |
          # ... errors from `get_read_policies_for_connection` and `get_write_policies_for_connection` including `Database.transaction`
```

But to properly fix it we need to revisit upstream modules like `Database` that is also returning `{:error, any()}` which is kinda of out scope for this PR so I can work on specs in a following PR.

leandrocp (changes requested):
My understanding of this module is still limited but changes looks good. Only one cosmetic suggestion to remove that unnecessary `__MODULE__` and https://github.com/supabase/realtime/pull/1828/changes#r3154332355 ~deserves another PR~ or can be done in this one :D

leandrocp (approved):

edgurgel (approved):

PR #1829, opened 2026-04-21
chore: clean up unused code

leandrocp:
Used Elixir 1.20 type system to identify violations.

### DIFF
--- a/lib/realtime/repo_replica.ex
+++ b/lib/realtime/repo_replica.ex
@@ -2,8 +2,6 @@ defmodule Realtime.Repo.Replica do
   @moduledoc """
   Generates a read-only replica repo for the region specified in config/runtime.exs.
   """
-  require Logger
-
   use Ecto.Repo,
     otp_app: :realtime,
     adapter: Ecto.Adapters.Postgres,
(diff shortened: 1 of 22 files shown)


lib/realtime/repo_replica.ex
```diff
@@ -30,12 +28,14 @@ defmodule Realtime.Repo.Replica do
     "us-west-1" => Realtime.Repo.Replica.SanJose
   }
 
-  @ast (quote do
-          use Ecto.Repo,
-            otp_app: :realtime,
-            adapter: Ecto.Adapters.Postgres,
-            read_only: true
-        end)
+  for replica_module <- Enum.uniq(Map.values(@replicas_fly) ++ Map.values(@replicas_aws)) do
```
leandrocp, line 31: I didn't find a reason to justify creating these modules at runtime, otherwise we can simplify it and save the overhead of calling `Code` and `Module` (even if just once per module).

lib/realtime_web/controllers/fallback_controller.ex
```diff
@@ -40,18 +40,6 @@ defmodule RealtimeWeb.FallbackController do
     |> render("error.json", message: message)
   end
 
-  def call(conn, {:error, %Ecto.Changeset{valid?: false} = changeset}) do
```
leandrocp, line 43: Dup of of `call/2` at line 22

lib/realtime_web/router.ex
```diff
@@ -172,7 +171,7 @@ defmodule RealtimeWeb.Router do
     # Must have been set by BaggageRequestId
     # We can't set the span attribute there because the phoenix span only starts after it reaches the Router
     if request_id = Logger.metadata()[:request_id] do
-      Tracer.set_attribute(:request_id, request_id)
+      OpenTelemetry.Tracer.set_attribute(:request_id, request_id)
```
leandrocp, line 174: `set_attribute/2` is not a macro.

filipecabaco (approved):

PR #1832, opened 2026-04-24
feat: add sql inspector to live dashboard

filipecabaco:

SQL inspector screen to run queries in read-only mode to avoid accessing the database via other less secure means. Sensitive data is properly hidden from the users.
Screenshot 2026-04-24 at 18 21 23


leandrocp (approved):
🚢

PR #1833, opened 2026-04-27
feat: improve observability on tenant migrations

leandrocp:
Related to REAL-765

**feat: record elapsed time to run tenant migrations**

We need this simple metric to track tenant migrations total execution time. It won't give individual migration timing per tenant but it's enough to signal which tenant need further observation.

**feat: log migration error pg_code and routine**

To facilitate querying and not crash the log server otherwise we need to parse the message at runtime.

### DIFF
--- a/lib/realtime/tenants/migrations.ex
+++ b/lib/realtime/tenants/migrations.ex
@@ -245,7 +245,8 @@ defmodule Realtime.Tenants.Migrations do
 
       try do
         opts = [all: true, prefix: "realtime", dynamic_repo: repo]
-        Ecto.Migrator.run(Repo, @migrations, :up, opts)
+        {time, _} = :timer.tc(fn -> Ecto.Migrator.run(Repo, @migrations, :up, opts) end)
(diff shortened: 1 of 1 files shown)


lib/realtime/tenants/migrations.ex
```diff
@@ -245,17 +245,27 @@ defmodule Realtime.Tenants.Migrations do
...
       rescue
         error ->
-          log_error("MigrationsFailedToRun", error)
+          log_error("MigrationsFailedToRun", error, migration_error_metadata(error))
           {:error, error}
       end
     end)
   end
 
+  defp migration_error_metadata(%Postgrex.Error{postgres: postgres}) when is_map(postgres) do
```
leandrocp, line 260: Store and normalize `pg_code` and `routine` from:

```
%Postgrex.Error{ message: nil, postgres: %{ code: :undefined_column, line: "3665", message: "column \"extension\" does not exist", position: "124", file: "parse_relation.c", unknown: "ERROR", severity: "ERROR", pg_code: "42703", routine: "errorMissingColumn"
```

filipecabaco (approved):

PR #1837, opened 2026-04-28
feat: remove awscli and jq as a dependency; remove erl crash dump

filipecabaco:

remove awscli and jq as a dependency; remove erl crash dump


Dockerfile
```diff
@@ -75,7 +75,7 @@ ENV SLOT_NAME_SUFFIX="${SLOT_NAME_SUFFIX}" \
     ERL_AFLAGS="-proto_dist inet6_tcp"
 
 RUN apt-get update -y && \
-    apt-get install -y libstdc++6 openssl libncurses5 locales iptables sudo tini curl awscli jq && \
+    apt-get install -y libstdc++6 openssl libncurses5 locales iptables sudo tini curl && \
```
leandrocp: The only small suggestion would be removing `sudo` because the container is already running as `root` but it saves only a few kb so it doesn't matter much

leandrocp (approved):

edgurgel (changes requested):
Have we tested that these curl calls actually work?

PR #1839, opened 2026-05-01
fix: unblock tenants and make migrations safer

leandrocp:
Step 1 of a bigger effort to make migrations more resilient.

close REAL-778
close REAL-779


compose.dbs.yml
```diff
@@ -18,7 +17,6 @@ services:
 
   tenant_db:
     image: supabase/postgres:17.6.1.074
-    container_name: tenant-db
```
edgurgel, line 21: why change this? 🤔

priv/repo/migrations/20260501120000_add_tenants_migrations_ran_index.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Realtime.Repo.Migrations.AddTenantsMigrationsRanIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change do
+    create index(:tenants, [:migrations_ran], concurrently: true)
```
edgurgel: Do we want to index migrations_ran?

The number of possible values for this column is not massive (low cardinality) so many many rows will land on very few places so this index will not be super fast to maintain or read

leandrocp: I added it mostly because queries in sql inspector are timing out when filtering by `migrations_ran` but for my use case adding metrics makes more sense actually so this won't be needed.

compose.dbs.yml
leandrocp: The only reason is to avoid conflicts when working with workktrees. Are you using that name to exec commands on the container?

filipecabaco (approved):

compose.dbs.yml
edgurgel: Oh no it was just easy to see which db was what but if that's easier with worktres let's remove it 👍

edgurgel (approved):

PR #1852, opened 2026-05-07
feat: setup `supabase_realtime_admin`

leandrocp:
Part of project **Migrations v2**.

The main change is connecting tenants using `supabase_realtime_admin` instead of `supabase_admin`. Related to https://github.com/orgs/supabase/discussions/34270

TLDR:
- PG 14 is mostly supported running as superuser, see README.md
- PG 15 up to 15.14.1.015 requires superuser due to a missing supautil grant on `realtime.subscription`
- PG >= supabase/postgres:15.14.1.129 can connect and work as `supabase_realtime_admin` with `REPLICATION` and `supautils` [managing policies](https://github.com/supabase/postgres/blob/438c36be6614a851e43a24399f92f92fb9a88613/ansible/files/postgresql_config/supautils.conf.j2).

The complexity of this PR is because we need to handle existing tenants running with `supabase_admin` and new projects that should start with `supabase_realtime_admin`, running from PG14 to 17. This is done mostly by executing the migration `setup_supabase_realtime_admin.ex` that is gated behind the feature flag `use_supabase_realtime_admin` (disabled by default).

### Review
Start on `## Postgres compatibility` in `README.md` as it sets the expectations and limitations, then `schema_test.exs` along with changes in `tests.yml` and `integration_tests.yml` that shows what is expected to work on each PG version.

### Tasks
Closes REAL-25
Closes REAL-377
Closes REAL-527
Closes REAL-630
Closes REAL-778
Closes REAL-781
Closes REAL-816
Closes SEC-562


lib/realtime/tenants/repo/migrations/20260606110000_subscription_check_filters_use_pg_attribute.ex
leandrocp, line 1: Previous function required `grant usage` because of `::regclass` casting, see test `"subscription works when role lacks usage permission"`.

priv/repo/dev_seeds.exs
```diff
@@ -69,5 +67,3 @@ case Tenants.Migrations.run_migrations(tenant) do
   :noop -> :ok
   _ -> raise "Running Migrations failed"
 end
-
-Tenants.Migrations.run_migrations(tenant)
```
leandrocp, line 70: This seems redundant so I removed it but not sure about this change.

test/realtime/extensions/cdc_rls/cdc_rls_test.exs
```diff
@@ -100,7 +100,8 @@ defmodule Realtime.Extensions.CdcRlsTest do
       %{oids: oids2} = :sys.get_state(subscriber_manager_pid)
       assert !Map.equal?(oids, oids2)
 
-      Postgrex.query!(conn, "create publication supabase_realtime_test for all tables", [])
```
leandrocp, line 103: `for all tables` requires `superuser` - is this a blocker? Otherwise we can grant it on `supabase/supautils` or create a function like `create_publication_for_all_tables(name text)`.

test/realtime/tenants/schema_test.exs
leandrocp, line 1: New test module to verify permissions, ie: restrict what must be restricted without breaking the system.

test/support/containers.ex
```diff
@@ -212,11 +211,27 @@ defmodule Containers do
...
+    Postgrex.query!(admin_conn, "GRANT USAGE ON SCHEMA realtime TO postgres, anon, authenticated, service_role", [])
+    Postgrex.query!(admin_conn, "GRANT ALL ON SCHEMA realtime TO supabase_realtime_admin WITH GRANT OPTION", [])
+  end
+
   def stop_containers() do
     {list, 0} = System.cmd("docker", ["ps", "-a", "--format", "{{.Names}}", "--filter", "name=realtime-test-*"])
-    names = list |> String.trim() |> String.split("\n")
 
-    for name <- names do
+    for name <- String.split(list, "\n", trim: true) do
```
leandrocp, line 239: Fixes a small bug on CI trying to stop empty `""` containers.

test/realtime/extensions/cdc_rls/cdc_rls_test.exs
edgurgel: I don't think this is a blocker because we don't control this publication ourselves, right? 🤔 This is just for the test

edgurgel (approved):
Great tests on schema_test.exs

test/realtime/extensions/cdc_rls/cdc_rls_test.exs
leandrocp: Now it's more clear that's not a blocker. Realtime doesn't depend on such statements and users are not supposed to have superuser to have that grant either.

lib/realtime/tenants/migrations.ex
```diff
@@ -297,14 +297,18 @@ defmodule Realtime.Tenants.Migrations do
...
+        create = """
         CREATE TABLE IF NOT EXISTS realtime.#{partition_name}
         PARTITION OF realtime.messages
-        FOR VALUES FROM ('#{start_timestamp}') TO ('#{end_timestamp}');
+        FOR VALUES FROM ('#{start_timestamp}') TO ('#{end_timestamp}')
         """
 
-        case Postgrex.query(conn, query, []) do
-          {:ok, _} -> Logger.debug("Partition #{partition_name} created")
+        alter_owner = "ALTER TABLE realtime.#{partition_name} OWNER TO supabase_realtime_admin"
```
edgurgel: oh so we always need to change this?

leandrocp: Yes and also new objects created through migrations (sorry I forgot to mention this). The reason is because we can't switch to `supabase_realtime_admin` on all projects, really only supabase/postgres >= 15.14.1.018 will work.

PR #1864, opened 2026-05-13
feat: single broadcast api payload

edgurgel:

Introduce a new endpoint: `POST /api/broadcast/:topic/events/:event_name` where the content-type defines the type of payload (`application/json` vs `application/octet-stream`)

if the channel is private a query string can be passed `?private=true`

### DIFF
--- a/lib/realtime_web/plugs/validate_broadcast_content_type.ex
+++ b/lib/realtime_web/plugs/validate_broadcast_content_type.ex
@@ -1,0 +1,39 @@
+defmodule RealtimeWeb.Plugs.ValidateBroadcastContentType do
+  @moduledoc """
+  Validates the request `Content-Type` for the broadcast-single endpoint.
+
+  Allowed: `application/json` and `application/octet-stream` (optionally with
+  parameters such as `; charset=utf-8`). A missing header is also allowed for
+  backward compatibility with callers that historically POSTed JSON without
+  setting the header.
+
+  Any other media type is rejected with a 415 response carrying a JSON body.
+  """
+  import Plug.Conn
+
+  @allowed ["application/json", "application/octet-stream"]
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    case get_req_header(conn, "content-type") do
+      [] ->
+        conn
+
+      [content_type | _] ->
+        if Enum.any?(@allowed, &String.starts_with?(content_type, &1)) do
+          conn
+        else
+          conn
+          |> put_resp_content_type("application/json")
+          |> send_resp(
+            415,
+            Jason.encode!(%{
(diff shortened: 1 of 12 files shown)


lib/realtime_web/plugs/validate_broadcast_content_type.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    case get_req_header(conn, "content-type") do
+      [] ->
+        conn
+
+      [content_type | _] ->
+        if Enum.any?(@allowed, &String.starts_with?(content_type, &1)) do
```
leandrocp: I think `String.starts_with?` is a bit naive, it matches `application/octet-streamfoo` and fails `Application/json` because it's case-sensitive. It's better to use [Plug.Conn.Utils.content_type/1](https://hexdocs.pm/plug/Plug.Conn.Utils.html#content_type/1)

test/realtime_web/controllers/broadcast_single_controller_test.exs
```diff
@@ -0,0 +1,639 @@
...
+                 "payload" => payload,
+                 "event" => event,
+                 "type" => "broadcast"
+               },
+               "ref" => nil,
+               "topic" => sub_topic
+             }
+
+      # Assert binary message received with V2Serializer format
+      assert_receive {:socket_push, :binary, data}
```
leandrocp, line 81: Asserting binary inside "returns 202 when JSON message is broadcasted"? 🤔

leandrocp (changes requested):

lib/realtime_web/plugs/validate_broadcast_content_type.ex
edgurgel: Oh nice TIL `Utils.content_type/1`

test/realtime_web/controllers/broadcast_single_controller_test.exs
edgurgel: Oh the `:binary` means that the WebSocket format is binary (That's what we broadcast on v2 for binary or json messages)

I will update the comment to avoid the confusion

leandrocp: Gotcha 👍🏻

edgurgel: Done!

lib/realtime_web/plugs/validate_broadcast_content_type.ex
edgurgel: Done!

leandrocp (approved):

PR #1866, opened 2026-05-13
feat(dashboard): tenant migrations page

leandrocp:
New page to help debug and fix tenants crashing due to bad DB state, usually caused by missing or altered objects. It won't execute any statement automatically but it shows a plan generated by [pg-delta](https://github.com/supabase/pg-toolbelt/tree/main/packages/pg-delta) and allows to apply the changes in the tenant DB.

### Features
- list all applied migrations
- output plan to fix bad state using pg-delta
- show potential risk (calculated by pg-delta)
- apply changes (fix issues)

### Extra
- added `postgres_version` in tenant info page (relevant to migrations)

#### No drift
no drift

#### Safe fix (create missing table)
add missing table safe

#### Risk of data loss
rename data loss

#### Apply
apply


Dockerfile
```diff
@@ -78,6 +98,8 @@ RUN apt-get update -y && \
     apt-get install -y libstdc++6 openssl libncurses5 locales iptables sudo tini curl awscli jq && \
     apt-get clean && rm -f /var/lib/apt/lists/*_*
 
+COPY --from=pgdelta-builder /tmp/pgdelta /usr/local/bin/pgdelta
```
leandrocp: Adds about ~95MB to runner image

lib/realtime_web/dashboard/tenant_migrations.ex
```diff
@@ -0,0 +1,302 @@
...
+       error: nil
+     )}
+  end
+
+  @impl true
+  def handle_params(%{"external_id" => ref}, _uri, socket) when ref != "" do
+    ref = String.trim(ref)
+
+    with %Tenant{} = tenant <- Api.get_tenant_by_external_id(ref),
+         {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id),
```
edgurgel: I don't think we should use `Connect` here but `Database.connect_db` or something as Connect will also start `ReplicationConnection` and we don't necessarily want that

lib/mix/tasks/realtime.export_tenant_db_baseline.ex
```diff
@@ -0,0 +1,74 @@
...
+  The target DB is read from `DB_HOST` / `DB_PORT` / `DB_NAME` / `DB_USER` /
+  `DB_PASSWORD` env vars.
+
+  The target DB is expected to already have all tenant migrations applied.
+
+  Requires the `pgdelta` binary on `$PATH`.
+  """
+  use Mix.Task
+
+  @baseline_path "priv/repo/tenant_db_baseline.json"
```
edgurgel, line 21: I'm wondering if we should have a CI task that breaks if baseline has changed but it has not been committed. Case: new migration is added but baseline has not been updated. Thoughts? 🤔

lib/realtime_web/dashboard/tenant_migrations.ex
leandrocp: Good to know, I will test `connect_db`

lib/mix/tasks/realtime.export_tenant_db_baseline.ex
leandrocp: Initially I had it in `ecto.migrate` alias but we don't want to call `pgdelta` on every migration but I agree we need some kind of check. I guess CI is our best option indeed.

avallete: Also worth to know that this might need to be regenerated when `pg-delta` version change.

Catalog extraction while it's alpha (even if we avoid it) might have changes in it's content, so new version will need a regeneration of the baseline.

Dockerfile
avallete: Just so I'm sure, this will get shipped within the actual dev container we ship on the cli ?

If so, that looks like a huge price to pay for something we might use only sometimes 🤔

leandrocp: > Just so I'm sure, this will get shipped within the actual dev container we ship on the cli ?

Correct.

> If so, that looks like a huge price to pay for something we might use only sometimes 🤔

Let me give more context to make it more clear. We have a small set of tenants crashing due to missing or altered objects (undefined_table, invalid_function_definition, etc) that we need to fix and this PR is adding the tool to help us do it (this is the short-term fix). The long-term solution is [locking down the `realtime` schema](https://github.com/supabase/realtime/pull/1852) which should prevent such errors and we could remove pgdelta and this page from realtime, so it's supposed to be temporary. If we decide to keep this page permanentelly then probably should move it to admin studio but requires way more work to implement.

avallete: Thanks for the context.

Just a naive question, since we already have pg-delta on the mgmt api (running withing workers).

If we want to produce a diff between a static catalog and the user database. Maybe we could expose an endpoint that does just that ? 

That would keep pgdelta out of the image while allowing a command within the realtime image to call the api asking for "diff with the running db" 

BTW I don't speak the Elixir and I've only flew over the PR so feel free to dismiss my comment if that doesn't make sense 😅

leandrocp: Removing pgdelta out of the image would be a big win!

> BTW I don't speak the Elixir and I've only flew over the PR so feel free to dismiss my comment if that doesn't make sense 😅

No worries! And I'm still learning what the platform does and how it operates (joined recently) 😄 

> If we want to produce a diff between a static catalog and the user database. Maybe we could expose an endpoint that does just that ?

Sounds interesting, I found the mgmt api and I'm exploring this idea. Thanks!

leandrocp: Hey @avallete I did some tests with mgmt api but hit a few complications, so I just added a compressed binary and was able to shave off a few MB from the image as well, taking it from  129MB (current main) down to 114MB even with the pre-built pgdelta.

I think that's pretty good to evaluate pgdelta and help us unblock the tenants and then we can re-evaluate moving it to somewhere else.

test/realtime_web/dashboard/tenant_info_test.exs
```diff
@@ -12,7 +11,7 @@ defmodule RealtimeWeb.Dashboard.TenantInfoTest do
       Application.delete_env(:realtime, :dashboard_credentials)
     end)
 
-    tenant = tenant_fixture()
+    tenant = Containers.checkout_tenant(run_migrations: true)
```
leandrocp, line 14: Changed this one because I added a query to fetch `pg_version` in tenant info page. We have this info in admin studio but it's worth having it at hand on realtime admin.

filipecabaco (approved):

PR #1869, opened 2026-05-15
feat: add support to realtime.messages binary payloads

edgurgel:

* Add `binary_payload` column to realtime.messages with a constraint that only apply to new rows that checks that we can't have `binary_payload` AND `payload` NOT NULL at the same time. They can both be NULL though because we already allowed `payload` to be NULL so I didn't want to break anythiung. We could be more strict and ensure that one of them exclusively must be NULL.
* Add `realtime.send` accepting `bytea` overloading the existing function.


lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex
```diff
@@ -0,0 +1,39 @@
...
+      BEGIN
+        generated_id := gen_random_uuid();
+
+        EXECUTE format('SET LOCAL realtime.topic TO %L', topic);
+
+        INSERT INTO realtime.messages (id, binary_payload, event, topic, private, extension)
+        VALUES (generated_id, payload, event, topic, private, 'broadcast');
+      EXCEPTION
+        WHEN OTHERS THEN
+          RAISE WARNING 'ErrorSendingBroadcastMessage: %', SQLERRM;
```
leandrocp, line 33: Maybe add this error into ERROR_CODES.md?

lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Realtime.Tenants.Migrations.AddBinaryPayloadToMessages do
+  @moduledoc false
+  use Ecto.Migration
+
+  def change do
+    execute("ALTER TABLE realtime.messages ADD COLUMN IF NOT EXISTS binary_payload bytea")
```
leandrocp, line 6: How do we handle big binaries? I can simulate a OOM crash with this payload:

```elixir
payload = :binary.copy(<<0>>, 900 * 1024 * 1024)

Postgrex.query!(
  db,
  "SELECT realtime.send($1::bytea, 'oom', 'adios', true)",
  [payload]
)
```

_Running in a container with ~1.2GB free mem_

edgurgel: The process would reach the max heap size and crash before OOMing (The container must have a bit more memory than the max heap size we set)

https://github.com/supabase/realtime/blob/ce7bd507eefdd27855cad37e15722ca5caa3b756/rel/vm.args.eex#L14

lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex
edgurgel: Oh yup!

lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex
leandrocp: Yep I was just concerned about exposing a possible scenario where one could try to send a really large bin for whatever reason.

leandrocp (approved):

lib/realtime/tenants/repo/migrations/20260514120000_add_binary_payload_to_messages.ex
edgurgel: Oh yeah makes sense! We do have this protection to not let any single process kill the system. The assumption is that we have enough free memory to not hit the limit quickly 👍

PR #1904, opened 2026-05-22
chore: test pg14

leandrocp:
I found tenants running on versions as old as 14.1.0.3

I used 14.1.0.85 in tests because that's the oldest amd64 image available, and the only incompatibility I found was in the function `realtime.broadcast_changes` that fails on PG <=14.5 with:

```
Failed to process the row: type of parameter 7 (<table_name>) does not
match that when preparing the plan (record)
```

I believe it's linked to https://github.com/postgres/postgres/commit/56d45fdab7b12cd0c767a8c1e7ab6c04390a32d8 but I found no error on logs, so I just skipped the 2 tests that hit that scenario.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -38,12 +38,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        postgres: [pg14, pg17]
+        postgres: [pg14, pg15, pg17]
         include:
           - postgres: pg14
             postgres_image: supabase/postgres:14.1.0.85
+          - postgres: pg15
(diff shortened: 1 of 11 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -38,10 +38,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        postgres: [pg15, pg17]
+        postgres: [pg14, pg17]
```
filipecabaco: we could have pg14,15,17 here IMO

filipecabaco (approved):

PR #1912, opened 2026-05-26
feat: AND filters in Postgres Changes

filipecabaco:

Implement AND filter support for Postgres Changes with comma separated parameters

### DIFF
--- a/lib/extensions/postgres_cdc_rls/subscriptions.ex
+++ b/lib/extensions/postgres_cdc_rls/subscriptions.ex
@@ -8,8 +8,11 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
 
   @type conn() :: Postgrex.conn()
   @type filter :: {binary, binary, binary}
-  @type subscription_params :: {action_filter :: binary, schema :: binary, table :: binary, [filter]}
-  @type subscription_list :: [%{id: binary, claims: map, subscription_params: subscription_params}]
+  @type subscription_params ::
+          {action_filter :: binary, schema :: binary, table :: binary, [filter]}
+  @type subscription_list :: [
+          %{id: binary, claims: map, subscription_params: subscription_params}
+        ]
 
   @filter_types ["eq", "neq", "lt", "lte", "gt", "gte", "in"]
 
@@ -274,11 +277,13 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
         {:ok, {action_filter, schema, table, []}}
 
       %{"schema" => schema}
-      when is_binary(schema) and not is_map_key(params, "table") and not is_map_key(params, "filter") ->
+      when is_binary(schema) and not is_map_key(params, "table") and
+             not is_map_key(params, "filter") ->
         {:ok, {action_filter, schema, "*", []}}
 
       %{"table" => table}
-      when is_binary(table) and not is_map_key(params, "schema") and not is_map_key(params, "filter") ->
+      when is_binary(table) and not is_map_key(params, "schema") and
+             not is_map_key(params, "filter") ->
         {:ok, {action_filter, "public", table, []}}
 
       map when is_map_key(map, "user_token") or is_map_key(map, "auth_token") ->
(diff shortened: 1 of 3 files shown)


test/integration/rt_channel/postgres_changes_test.exs
```diff
@@ -388,6 +388,73 @@ defmodule Realtime.Integration.RtChannel.PostgresChangesTest do
     end
   end
 
+  describe "AND filter composition" do
+    test "receives row matching all filters, ignores row matching only one filter",
```
leandrocp: Not a big deal but we could split this long test into smaller ones with a single block setup:

```
test "delivers row matching all filters" ...
test "ignores row matching only one filter" ...
```

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -209,6 +212,19 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
 
+  AND composition — multiple filters separated by commas:
+
+      iex> parse_subscription_params(%{"schema" => "public", "table" => "messages", "filter" => "id=gt.0,id=lt.100"})
+      {:ok, {"*", "public", "messages", [{"id", "gt", "0"}, {"id", "lt", "100"}]}}
+
+  empty or whitespace-only filter string is treated as no filter:
+
+      iex> parse_subscription_params(%{"schema" => "public", "table" => "messages", "filter" => ""})
+      {:ok, {"*", "public", "messages", []}}
```
leandrocp, line 226: Empty filters fail on main:

```
iex> Extensions.PostgresCdcRls.Subscriptions.parse_subscription_params(%{"schema" => "public", "table" => "messages", "filter" => ""})
{:error, "Error parsing `filter` params: [\"\"]"}
```

Should this change be considered a soft-breaking change or is this fine?

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -296,6 +304,60 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
 
   defp action_filter(_), do: "*"
 
+  defp parse_filters(filter) do
```
leandrocp: I did some tests parsing with binary matching and I found some performance gains, about 3-5x faster and 5-10x memory reduction for small filters (increases as more filters are used) but the code might not look the prettiest:

```elixir
  defp parse_filters(filter) when is_binary(filter) do
    case String.trim(filter) do
      "" -> {:ok, []}
      filter -> scan(filter, filter, 0, 0, 0, [])
    end
  end

  defp scan(<<>>, orig, start, len, _depth, acc) do
    case parse_segment(binary_part(orig, start, len)) do
      {:ok, parsed} -> {:ok, Enum.reverse([parsed | acc])}
      {:error, _} = e -> e
    end
  end

  defp scan(<<"(", rest::binary>>, orig, start, len, depth, acc) do
    scan(rest, orig, start, len + 1, depth + 1, acc)
  end

  defp scan(<<")", rest::binary>>, orig, start, len, depth, acc) do
    new_depth = if depth > 0, do: depth - 1, else: 0
    scan(rest, orig, start, len + 1, new_depth, acc)
  end

  defp scan(<<",", rest::binary>>, orig, start, len, 0, acc) do
    case parse_segment(binary_part(orig, start, len)) do
      {:ok, parsed} -> scan(rest, orig, start + len + 1, 0, 0, [parsed | acc])
      {:error, _} = e -> e
    end
  end

  defp scan(<<_::8, rest::binary>>, orig, start, len, depth, acc) do
    scan(rest, orig, start, len + 1, depth, acc)
  end

  defp parse_segment(segment) do
    case String.trim(segment) do
      "" ->
        {:error, "filter must not contain empty segments (check for extra commas)"}

      segment ->
        case :binary.split(segment, "=") do
          [col, "eq." <> v] -> {:ok, {col, "eq", v}}
          [col, "neq." <> v] -> {:ok, {col, "neq", v}}
          [col, "lte." <> v] -> {:ok, {col, "lte", v}}
          [col, "lt." <> v] -> {:ok, {col, "lt", v}}
          [col, "gte." <> v] -> {:ok, {col, "gte", v}}
          [col, "gt." <> v] -> {:ok, {col, "gt", v}}
          [col, "in." <> v] -> format_in(col, v)
          [_col, rest] -> {:error, inspect(:binary.split(rest, "."))}
          [bad] -> {:error, inspect([bad])}
        end
    end
  end

  defp format_in(col, value) do
    size = byte_size(value)

    if size >= 2 and binary_part(value, 0, 1) == "(" and binary_part(value, size - 1, 1) == ")" do
      {:ok, {col, "in", "{" <> binary_part(value, 1, size - 2) <> "}"}}
    else
      {:error, "`in` filter value must be wrapped by parentheses"}
    end
  end
```

leandrocp (approved):
Non-blocking comments, just a question about empty filters.

lib/extensions/postgres_cdc_rls/subscriptions.ex
filipecabaco: uuuhhh nice!!! I think we can use the better version as this is a very used feature in our stack so any reduction is a good reduction

lib/extensions/postgres_cdc_rls/subscriptions.ex
filipecabaco: not sure honestly... IMO we could see this as a bug we had and fix it or we can see it as properly informing the user of this error AS an error. wdyt?

test/integration/rt_channel/postgres_changes_test.exs
filipecabaco: will do

lib/extensions/postgres_cdc_rls/subscriptions.ex
leandrocp: IMO an empty filter means... no filter so this change looks correct and since it used to return error, the impact of that change should be minimal.

But my question was also to understand if such changes should be listed on https://supabase.com/changelog or any other comm.

filipecabaco: the and filter will also need to be shared as this will potentially reduce data sent so it has that benefit for us also

leandrocp: To answer my own question, I think we should update realtime docs (new AND operator and this empty filter behavior change) and also announce on /changelog

filipecabaco: will get the docs PR ready so we can merge almost at the same time

filipecabaco: https://github.com/supabase/supabase/pull/46404

leandrocp (approved):

PR #1919, opened 2026-05-28
fix: ReplicationPoller consumes slot if publication is not empty

edgurgel:

* Change ReplicationPoller to only start a replication slot if the publication is not empty. 
* Also track when publication changes from empty to non-empty and vice-versa and react accordingly

The goal here is only to establish the replication slot when absolute necessary

Why do we even keep the ReplicationPoller up if there is no publication? If we don't whenever a postgres_changes join happens it will start it again. While this approach will leave this process up and running until it notices that there are no subscriptions (SubscriptionManager does this) and stop everything.

I've also fixed the case when fetching the publication tables could fail but we would assume that the publication was empty


filipecabaco (approved):

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -169,6 +216,35 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
+        {:noreply, prepare_replication(%{state | oids: new_oids})}
+
+      {n, 0} when n > 0 ->
+        cancel_timer(state.poll_ref)
+
+        case Replications.drop_replication_slot(conn, state.slot_name) do
+          {:error, reason} when reason != :slot_not_found ->
+            # The slot is a temporary logical replication slot tied to this connection,
+            # so stopping the process releases it without leaking WAL.
+            log_error("DropReplicationSlotFailed", reason)
```
leandrocp, line 269: Missing entry in ERROR_CODES.md :)

edgurgel: :tableflip:

edgurgel: Fixed!

filipecabaco (approved):

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -99,6 +142,11 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
   end
 
   @impl true
+  def handle_info(:poll, %{oids: oids, poll_ref: poll_ref} = state) when map_size(oids) == 0 do
+    cancel_timer(poll_ref)
```
leandrocp, line 145: I see that no caller use the return info from `cancel_timer` so that helper could be rewritten to:

```elixir
  @spec cancel_timer(reference() | nil) :: :ok
  def cancel_timer(ref) when is_reference(ref), do: Process.cancel_timer(ref, info: false)
  def cancel_timer(_), do: :ok
```

lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -1,14 +1,54 @@
...
+
+  ## Poll loop
+
+  Each `:poll` calls `Replications.list_changes/5`, which drains the slot and
+  fans changes out to subscriber nodes. Reschedule cadence depends on activity:
+
+    * rows processed → poll again immediately,
+    * raw slot changes present but nothing for subscribers → poll after
+      `poll_interval_ms` (+ jitter),
+    * fully idle → back off to `poll_interval_ms * @idle_multiplier`.
```
leandrocp:
```suggestion
    * rows processed → poll again immediately
    * raw slot changes present but nothing for subscribers → poll after `poll_interval_ms` (+ jitter)
    * fully idle → back off to `poll_interval_ms * @idle_multiplier`.
```

leandrocp (approved):
Just 2 small suggestions, non-blocking.

lib/extensions/postgres_cdc_rls/replication_poller.ex
edgurgel: Oh yeah that's a great point

PR #1920, opened 2026-05-28
feat: add filters to pg_changes

filipecabaco:

add filters to pg_changes:

| Operator | SQL | Description |
|----------|-----|-------------|
| `like` | `LIKE` | Column matches pattern (case-sensitive). |
| `ilike` | `ILIKE` | Column matches pattern (case-insensitive). |
| `not_like` | `NOT LIKE` | Column does not match pattern (case-sensitive). |
| `not_ilike` | `NOT ILIKE` | Column does not match pattern (case-insensitive). |
| `is` | `IS` | `null`, `true`, `false`, `unknown` |
| `not_is` | `IS NOT` | `null`, `true`, `false`, `unknown` |


filipecabaco:
https://github.com/supabase/realtime/pull/1922 

we had bad branch name

PR #1921, opened 2026-05-28
feat: allow select on pg_changes

filipecabaco:

Allows to select which columns to return on pg changes payload


lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -296,6 +299,22 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
 
+  defp parse_select(%{"select" => cols}) when is_list(cols) do
+    case Enum.filter(cols, &is_binary/1) do
+      [] -> nil
+      valid -> valid
+    end
+  end
+
+  defp parse_select(%{"select" => str}) when is_binary(str) do
+    case str |> String.trim() |> String.split(",", trim: true) |> Enum.map(&String.trim/1) do
```
leandrocp: What about special cases?

```elixir
%{"select" => "foo "} # parse as ["foo"]
%{"select" => "foo,bar"} # parse as ["foo", "bar"] but table name could be "foo,bar"
%{"select" => "."} # not sure if this would break something
# ... and potentialy more 😄
```

test/e2e/supabase/.temp/cli-latest
leandrocp, line 1: It this .temp dir supposed to be staged?

mise.toml
```diff
@@ -60,3 +60,17 @@ run = "docker compose -f compose.tests.yml stop"
 [tasks.test-rm]
 description = "Remove test services"
 run = "docker compose -f compose.tests.yml rm -sf"
+
+[tasks.e2e]
```
leandrocp, line 64: This is so useful!

leandrocp (approved):

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -74,29 +74,31 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
         $6,
         $5,
-        $7
+        $7,
+        $8
       from
         sub_tables
         on conflict
-        (subscription_id, entity, filters, action_filter)
+        (subscription_id, entity, filters, action_filter, coalesce(selected_columns, '{}'))
```
edgurgel, line 90: Do we want to coalesce here to `{}`? I thought this would mean no selected columns? Or `NULL` and `{}` mean all columns?

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -296,6 +299,22 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
+      valid -> valid
+    end
+  end
+
+  defp parse_select(%{"select" => str}) when is_binary(str) do
+    case str |> String.split(",") |> Enum.reject(&(&1 == "")) do
+      [] -> nil
+      cols -> cols
+    end
+  end
```
edgurgel, line 314: Do we even want to accept this as a single string? Can we just enforce it to be a list of strings to avoid issues on parsing?

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -263,28 +265,29 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
...
+          {:ok, filters} -> {:ok, {action_filter, schema, table, filters, selected_columns}}
           {:error, reason} -> {:error, "Error parsing `filter` params: #{reason}"}
         end
 
       %{"schema" => schema, "table" => table}
       when is_binary(schema) and is_binary(table) and not is_map_key(params, "filter") ->
-        {:ok, {action_filter, schema, table, []}}
+        {:ok, {action_filter, schema, table, [], selected_columns}}
 
       %{"schema" => schema}
```
edgurgel, line 282: As discussed offline we might want to block select on this type of pg changes subscription as tables have different columns etc?

PR #1922, opened 2026-05-28
feat: add filters to pg_changes

filipecabaco:

add filters to pg_changes:

| Operator | SQL | Description |
|----------|-----|-------------|
| `like` | `LIKE` | Column matches pattern (case-sensitive). |
| `ilike` | `ILIKE` | Column matches pattern (case-insensitive). |
| `not_like` | `NOT LIKE` | Column does not match pattern (case-sensitive). |
| `not_ilike` | `NOT ILIKE` | Column does not match pattern (case-insensitive). |
| `is` | `IS` | `null`, `true`, `false`, `unknown` |
| `not_is` | `IS NOT` | `null`, `true`, `false`, `unknown` |

### DIFF
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -105,11 +105,15 @@ Multiple filters are separated by commas. All conditions must be true for a chan
 col1=operator.value,col2=operator.value
 ```
 
-Values that contain a comma or whitespace must be wrapped in double quotes:
+#### Reserved characters
+
+Values that contain reserved characters must be wrapped in double quotes:
 
 ```
-col="some value, with comma"
+col=eq."some value, with comma"
 ```
+
+Reserved characters: comma (`,`), and whitespace.
 
 #### Comparison operators
 
--- a/FEATURES.md
+++ b/FEATURES.md
@@ -1,0 +1,185 @@
+# Realtime Features
+
+Realtime exposes three features over WebSockets. All three share the same connection and channel model: a client opens a single WebSocket, then joins one or more named topics (channels). Each feature is opted into per channel via the `config` payload at join time.
+
+---
+
+## Broadcast
+
+Send ephemeral messages between clients subscribed to the same topic with low latency.
+
+### How it works
+
+1. A client sends a `broadcast` event to a channel topic.
+2. The server validates the payload size against the tenant's `max_payload_size_in_kb` limit.
+3. On private channels, RLS write policies are checked before the message is forwarded.
+4. The message is dispatched to all other subscribers on the topic. By default, the sender is excluded.
+
+### Configuration
+
+Set at channel join time under `config.broadcast`:
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `self` | boolean | When `true`, the sender also receives its own broadcast messages. Default `false`. |
+| `ack` | boolean | When `true`, the server replies with `:ok` after dispatching. Default `false` (fire-and-forget). |
+
+### Constraints
+
+- Payload size is capped at the tenant's `max_payload_size_in_kb` (with a ±500 byte margin).
+- Rate is capped at the tenant's configured events-per-second limit. Exceeding it disconnects the client.
+- On private channels, if the write policy denies access, the message is silently dropped.
+
+---
+
+## Presence
+
+Track and synchronize which clients are connected, along with arbitrary metadata.
+
+### How it works
+
+1. A client sends a `presence` event with `track` or `untrack` action and an optional payload map.
+2. `track` creates or updates the client's presence entry in a distributed ETS-backed store (Phoenix.Presence).
+3. All subscribers on the topic receive a `presence_diff` message with the join/leave delta.
+4. New joiners receive a `presence_state` message with the full current state for that topic.
+5. `untrack` removes the entry immediately.
+
+Presence state is keyed per client (UUID by default, or a caller-provided string via `config.presence.key`). Duplicate `track` calls with the same payload are no-ops.
+
+### Configuration
+
+Set at channel join time under `config.presence`:
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `key` | string | Custom key to identify this client's presence entry. Defaults to a generated UUID. |
+
+### Constraints
+
+- Payload must be a JSON object (map).
+- Server-side rate limit: `max_presence_events_per_second` per tenant.
+- Client-side rate limit: 5 events per 30-second window by default.
+- Payload size is subject to the same `max_payload_size_in_kb` limit as Broadcast.
+- On private channels, RLS read and write policies are enforced.
+
+---
+
+## Postgres Changes
+
+Listen to INSERT, UPDATE, and DELETE events on Postgres tables and receive them in real time.
+
+### How it works
+
+1. At channel join, the client declares one or more subscriptions in `config.postgres_changes`. Each subscription specifies an event type, a table, and optional filters.
+2. The server creates a `realtime.subscription` row in the tenant database for each subscription, storing the filters alongside the subscriber identity.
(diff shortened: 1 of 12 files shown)


lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -153,6 +155,15 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
...
         {:noreply, %{state | backoff: backoff, retry_ref: retry_ref, retry_count: retry_count + 1}}
 
+      {:error, %Postgrex.Error{postgres: %{code: :raise_exception, message: msg}}} ->
+        log_error("SubscriptionFatalError", msg)
+
+        for {pid, _id, _ref, _node} <- :ets.tab2list(subscribers_pids_table) do
+          send(pid, {:subscription_fatal_error, msg})
+        end
+
+        {:noreply, state}
```
edgurgel, line 165: Could you explain why we need this? Is it related to filters?

FEATURES.md
leandrocp, line 1: It would be interesting to add a topic in README linking to this doc, useful for humans and helps AI to discover it too.

FEATURES.md
```diff
@@ -0,0 +1,185 @@
...
+
+Presence state is keyed per client (UUID by default, or a caller-provided string via `config.presence.key`). Duplicate `track` calls with the same payload are no-ops.
+
+### Configuration
+
+Set at channel join time under `config.presence`:
+
+| Key | Type | Description |
+|-----|------|-------------|
+| `key` | string | Custom key to identify this client's presence entry. Defaults to a generated UUID. |
```
leandrocp, line 55: Since v4 is more popular it might be worth noting it generates v1 

```suggestion
| `key` | string | Custom key to identify this client's presence entry. Defaults to a generated UUID v1. |
```

FEATURES.md
```diff
@@ -0,0 +1,185 @@
...
+
+1. At channel join, the client declares one or more subscriptions in `config.postgres_changes`. Each subscription specifies an event type, a table, and optional filters.
+2. The server creates a `realtime.subscription` row in the tenant database for each subscription, storing the filters alongside the subscriber identity.
+3. A replication poller reads changes from a logical replication slot (WAL) on a configurable interval.
+4. For each WAL record, the tenant database runs `realtime.apply_rls()`, which:
+   - Checks RLS policies for each subscribing role.
+   - Runs `realtime.is_visible_through_filters()` to match the row against each subscription's filters.
+5. Matching subscription IDs are returned alongside the change and dispatched to the appropriate clients.
+
+### Subscription parameters
```
leandrocp, line 81: Should add `select` too?

FEATURES.md
```diff
@@ -0,0 +1,185 @@
...
+
+```
+col1=operator.value,col2=operator.value
+```
+
+Values that contain a comma or whitespace must be wrapped in double quotes:
+
+```
+col="some value, with comma"
+```
```
leandrocp: We could introduce the idea of [reserved characters](https://postgrest.org/en/stable/references/api/url_grammar.html#reserved-chars) already, something like:

````suggestion
#### Reserved characters

Values that contain reserved characters must be wrapped in double quotes:

```
col=eq."some value, with comma"
```

Reserved characters: comma (`,`), and whitespace.
````

Then we can expand the list as we add more operators.

File: FEATURES.md [hunk 5]
```diff
@@ -0,0 +1,185 @@
...
+#### Pattern matching operators
+
+Standard SQL pattern syntax applies: `%` matches any sequence of characters, `_` matches any single character.
+
+| Operator | SQL | Description |
+|----------|-----|-------------|
+| `like` | `LIKE` | Column matches pattern (case-sensitive). |
+| `ilike` | `ILIKE` | Column matches pattern (case-insensitive). |
+| `not_like` | `NOT LIKE` | Column does not match pattern (case-sensitive). |
+| `not_ilike` | `NOT ILIKE` | Column does not match pattern (case-insensitive). |
```
leandrocp, line 151: Is this documenting `realtime.equality_op` or the public filter operators? I was expecting to see `not.{...}` in this doc 🤔

lib/realtime/tenants/repo/migrations/20260528120000_add_like_ilike_is_not_ops_to_filters.ex
```diff
@@ -0,0 +1,143 @@
...
+  use Ecto.Migration
+
+  def change do
+    execute("alter type realtime.equality_op add value 'like';")
+    execute("alter type realtime.equality_op add value 'ilike';")
+    execute("alter type realtime.equality_op add value 'is';")
+    execute("alter type realtime.equality_op add value 'not_in';")
+    execute("alter type realtime.equality_op add value 'not_like';")
+    execute("alter type realtime.equality_op add value 'not_ilike';")
+    execute("alter type realtime.equality_op add value 'not_is';")
```
leandrocp:
```suggestion
    execute("alter type realtime.equality_op add value if not exists 'like'")
    execute("alter type realtime.equality_op add value if not exists 'ilike'")
    execute("alter type realtime.equality_op add value if not exists 'is'")
    execute("alter type realtime.equality_op add value if not exists 'not_in'")
    execute("alter type realtime.equality_op add value if not exists 'not_like'")
    execute("alter type realtime.equality_op add value if not exists 'not_ilike'")
    execute("alter type realtime.equality_op add value if not exists 'not_is'")
```

lib/realtime/tenants/repo/migrations/20260528120000_add_like_ilike_is_not_ops_to_filters.ex
```diff
@@ -0,0 +1,143 @@
...
+          immutable
+          language plpgsql
+      as $$
+      declare
+          op_symbol text;
+          res boolean;
+      begin
+          -- IS / IS NOT require keyword RHS, not a typed literal
+          if op = 'is' or op = 'not_is' then
+              if val_2 not in ('null', 'true', 'false', 'unknown') then
```
leandrocp, line 32: Two comments here:

1. PostgREST supports `not_null` as valid value for `IN` operator as well, I think we could easily support that one too.

2. Value in PostgREST is case-insensitive (null, NULL, etc). Should we adhere to the same behaviour?

leandrocp (changes requested):
Just minor things, but interesting features!

To resolve the conflict on `tenant_db_baseline.json` just commit whatever and CI will regenerate it.

lib/extensions/postgres_cdc_rls/replication_poller.ex
filipecabaco: handling unknown filters to prevent the user from connecting

FEATURES.md
filipecabaco: will create a sort of index in README

FEATURES.md
filipecabaco: wasn't there yet but will do

lib/realtime/tenants/repo/migrations/20260528120000_add_like_ilike_is_not_ops_to_filters.ex
filipecabaco: yes we should will take that into account

FEATURES.md
filipecabaco: not is a weird one as can only "not" on certain ops...

lib/extensions/postgres_cdc_rls/replication_poller.ex
leandrocp: So `SubscriptionFatalError` is missing in ERROR_CODES.md :)

PR #1925, opened 2026-05-29
fix: prevent usage of select with wildcard; handle edge cases

filipecabaco:

* Prevents using the new `select` with `*`
* Coalesces `NULL` to `{}`
* Force usage of list of strings to prevent errors

### DIFF
--- a/lib/extensions/postgres_cdc_rls/subscriptions.ex
+++ b/lib/extensions/postgres_cdc_rls/subscriptions.ex
@@ -87,6 +87,7 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
       from
         sub_tables
         on conflict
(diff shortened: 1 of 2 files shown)


leandrocp (approved):

lib/extensions/postgres_cdc_rls/subscriptions.ex
```diff
@@ -87,6 +87,7 @@ defmodule Extensions.PostgresCdcRls.Subscriptions do
       from
         sub_tables
         on conflict
+        -- coalesce needed: NULL != NULL in unique constraints; NULL selected_columns means all columns
         (subscription_id, entity, filters, action_filter, coalesce(selected_columns, '{}'))
```
edgurgel, line 91: Again...shouldn't this default to NULL if selected_columns is not defined? It's the default behaviour when no selected_columns are defined AKA before this change. Or am I missing something?

PR #1928, opened 2026-06-01
feat(observability): per-tenant subscription metrics

leandrocp:
Add per-tenant subscriptions metrics to have more visiblity on its operation including:

- subscriptions
- poller dispatches
- poller failures


lib/extensions/postgres_cdc_rls/replication_poller.ex
```diff
@@ -133,6 +149,12 @@ defmodule Extensions.PostgresCdcRls.ReplicationPoller do
         [_, db_pid] = Regex.run(~r/PID\s(\d*)$/, msg)
         db_pid = String.to_integer(db_pid)
 
+        Realtime.Telemetry.execute([:realtime, :replication, :poller, :query, :exception], %{}, %{
+          tenant: tenant_id,
+          reason: :object_in_use,
+          db_pid: db_pid
```
edgurgel: Just to confirm: we don't have intentions to use db_pid as a metric tag right?

edgurgel (approved):

lib/extensions/postgres_cdc_rls/replication_poller.ex
leandrocp: Definitely not. I was hoping to use a telemetry handler to emit logs (like migrations) which is a very common pattern but need to discuss about it first so this is pretty much just a left over.

PR #1937, opened 2026-06-03
fix: handle connection closed errors

filipecabaco:

handle connection closed errors on authorization endpoint which was failing with improperly matched value


lib/realtime/tenants/authorization.ex
```diff
@@ -276,10 +282,13 @@ defmodule Realtime.Tenants.Authorization do
       conn,
       fn transaction_conn ->
         set_conn_config(transaction_conn, authorization_context)
-        policies = check_write_policies(transaction_conn, authorization_context, extensions, policies)
 
-        Postgrex.query!(transaction_conn, "ROLLBACK AND CHAIN", [])
-        policies
+        with {:ok, policies} <- check_write_policies(transaction_conn, authorization_context, extensions, policies) do
```
leandrocp, line 284: What about `check_read_policies`? All conn errors are captured here as well?

```elixir
      |> get_read_policies_for_connection(authorization_context, policies, opts)
      |> handle_policies_result(rate_counter)
```

filipecabaco: added

leandrocp (approved):

PR #1940, opened 2026-06-04
feat: default messages partition

leandrocp:
We still have edge cases where realtime.send would raise ErrorSendingBroadcastMessage because of missing partitions, even with health check creating them (#1939) for example when the function is called before a WS conn or before health check.

This is a proposal to close that gap by using a default partition which works as a catch-all in case a partition doesn't exist yet, again rare cases but still possible.

Fixes REAL-319
Fixes REAL-500
Fixes REAL-848


leandrocp:
After some discussion let's move to a warning approach instead.

PR #1946, opened 2026-06-08
feat: run migration on health check without blocking it - REAL-852

leandrocp:
Blocking Migrations executions are tried whenever an idle tenant is found (no connections) which is not rare and cause timeouts when:

- migrations are failing due to missing or invalid objects
- far away tenant db conn

Either way health checks should respond as fast as possible and not mutate data/db in-place, but still have a chance to trigger a proccess to fix missing migrations.


lib/realtime/tenants/migrations.ex
```diff
@@ -15,158 +15,81 @@ defmodule Realtime.Tenants.Migrations do
...
-    {20_251_120_212_548, AddActionToSubscriptions},
-    {20_251_120_215_549, FilterActionPostgresChanges},
-    {20_260_218_120_000, FixByteaDoubleEncodingInCast},
-    {20_260_326_120_000, ListChangesWithSlotCount},
-    {20_260_514_120_000, AddBinaryPayloadToMessages},
-    {20_260_527_120_000, AddSelectColumnsToSubscriptions},
-    {20_260_528_120_000, Wal2jsonEscapeSpecialChars},
-    {20_260_603_120_000, AddSendBinaryFunction},
-    {20_260_605_120_000, RenameBroadcastSendWarning}
+    {20_211_116_024_918, __MODULE__.CreateRealtimeSubscriptionTable},
```
leandrocp: That's a cosmetic change to avoid duplicating module names in alias + this list but I can revert it if doesn't look good.

lib/realtime/tenants.ex
```diff
@@ -95,11 +95,12 @@ defmodule Realtime.Tenants do
 
       connected_cluster when is_integer(connected_cluster) ->
         tenant = Cache.get_tenant_by_external_id(external_id)
-        migrations = Migrations.run_migrations(tenant)
+        healthy = not Migrations.run_migrations?(tenant)
+        healthy || Migrations.run_migrations_async(tenant)
```
edgurgel: nitpick: I prefer a direct 
`if not healthy, do: Migrations.run_migrations_async(tenant)`

as it's more readable

But it's a preference so not too worried about this

edgurgel (approved):

lib/realtime/tenants.ex
leandrocp: No personal pref and I agree it reads better :)

PR #1954, opened 2026-06-12
feat: introduce least-privilege realtime user

leandrocp:
Realtime needs superuser to run migrations to bootstrap the project, including roles and permissions setup.

Runtime on the other hand should connect as least-privilege, more controlled role for security reasons, which is the DB_USER_REALTIME added in this PR.

That role is optional to avoid disruption of existing projects and/or cases where superuser is still needed.

Follow-up of #1852 and part of Migrations v2 project.

Closes REAL-816
Closes REAL-781


edgurgel (approved):
LGTM

PR #1958, opened 2026-06-15
fix: reject broadcast API from suspended tenants

edgurgel:

Reject broadcast API from suspended tenants

### DIFF
--- a/lib/realtime/tenants/batch_broadcast.ex
+++ b/lib/realtime/tenants/batch_broadcast.ex
@@ -48,6 +48,10 @@ defmodule Realtime.Tenants.BatchBroadcast do
     broadcast(auth_params, %Tenant{} = tenant, messages, super_user)
   end
 
+  def broadcast(_auth_params, %Tenant{suspend: true}, _messages, _super_user) do
+    {:error, :forbidden, "Tenant is suspended"}
+  end
+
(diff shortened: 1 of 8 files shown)


filipecabaco (approved):

lib/realtime/tenants/batch_broadcast.ex
```diff
@@ -48,6 +48,10 @@ defmodule Realtime.Tenants.BatchBroadcast do
     broadcast(auth_params, %Tenant{} = tenant, messages, super_user)
   end
 
+  def broadcast(_auth_params, %Tenant{suspend: true}, _messages, _super_user) do
+    {:error, :forbidden, "Tenant is suspended"}
```
leandrocp, line 52: The fallback controller will return status 403 and also log error `UnprocessableEntity` which is a 422, so two questions here:

1. Status code differs from the actual log error, is that expected?
2. Tenants might flood the log with errors while they're suspended, is that expected?

edgurgel: Oh yeah definitely a bug. We don't want that will fix it

edgurgel: https://github.com/supabase/realtime/pull/1958/commits/17be90bb713c016b961817e5e39072f07aba3c4a

PR #1962, opened 2026-06-15
fix: add system message to signal replication ready

filipecabaco:

Sends a new message when replication connection starts. We only send when we receive the keep alive or a write message from the replication slot.


edgurgel (changes requested):
So this will only publish once right? If you join after ReplicationConnection has already started streaming it would be nice to also find out that it's replicating. Do we want to include this into the `join` payload that we return as well?

filipecabaco:
oh shit yes you are right... 🤦‍♂️

lib/realtime/tenants/replication_connection.ex
```diff
@@ -309,17 +312,29 @@ defmodule Realtime.Tenants.ReplicationConnection do
 
   @impl true
   def handle_data(data, state) when is_keep_alive(data) do
-    %KeepAlive{reply: reply, wal_end: wal_end} = parse(data)
-    wal_end = wal_end + 1
+    case maybe_broadcast_replication_started(state) do
```
edgurgel: BTW I don't think we even need this. 

`Connect` will trigger a "ready" message when the replication connection pid is ready. 

We just need to include it:

https://github.com/supabase/realtime/blob/6b4916bdd6d147a918b8f80027179ea6b31ab44e/lib/realtime/syn_handler.ex#L17

```elixir
Endpoint.local_broadcast(Connect.syn_topic(tenant_id), "ready", %{pid: pid, conn: conn, replication_conn: replication_conn}) 
```

We can then just subscribe to this syn topic ^ and wait for a "ready" with the replication_conn not `nil`

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -104,6 +104,8 @@ defmodule RealtimeWeb.RealtimeChannel do
 
       RealtimeWeb.Endpoint.subscribe(tenant_topic, metadata: metadata)
       RealtimeWeb.Endpoint.subscribe("realtime:operations:" <> tenant_id, metadata: metadata)
+      RealtimeWeb.Endpoint.subscribe(Connect.syn_topic(tenant_id))
+      send(self(), :notify_replication_ready)
```
edgurgel: This will add a new message to every channel that connects. Could we just include this information in the `phx_join` payload instead of always sending a message?

 And only send a message if replication status was offline?

We would only need to subscribe when the replication status is not offline.

We can also unsubscribe once the system message was sent.

WDYT?

filipecabaco:
I changed the approach and followed something similar to what we do for postgres changes where the channel will be the one checking.

the usage of syn and pubsub can actually lead to a race condition that would mean losing this message so it would be quite bad to handle and just increases complexity to be sure things work as expected

cc @edgurgel

edgurgel (approved):
Approving but I think this new option should be under `broadcast` options and not on the top level. ReplicationConnection only serves broadcast messages from `realtime.messages`

PR #1966, opened 2026-06-17
feat: add new pg filters

filipecabaco:

Adds `like`, `ilike`, `is`, `match`, `imatch`, `isdistinct` operators and `not.` negation to postgres_changes filters

https://linear.app/supabase/issue/REAL-832/improve-filter-parity-based-on-postgrest


leandrocp (approved):

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -254,7 +254,7 @@ defmodule RealtimeWeb.RealtimeChannelTest do
       assert_push "system",
                   %{
                     message:
-                      "Unable to subscribe to changes with given parameters. An exception happened so please check your connect parameters: [event: *, schema: public, table: test, filters: [{\"notacolumn\", \"eq\", \"123\"}], select: nil]. Exception: ERROR P0001 (raise_exception) invalid column for filter notacolumn",
+                      "Unable to subscribe to changes with given parameters. An exception happened so please check your connect parameters: [event: *, schema: public, table: test, filters: [{\"notacolumn\", \"eq\", \"123\", false}], select: nil]. Exception: ERROR P0001 (raise_exception) invalid column for filter notacolumn",
```
edgurgel, line 257: nitpick and not a blocker:

I wonder if we should change this to not simply inspect and "pretty print" the filters? 🤔

filipecabaco: yup we need something better than this. we have a task to have opcodes for this type of error so we might use that one to also improve on this error message overall including something better rather than a blank inspect

lib/realtime/tenants/repo/migrations/20260616120000_add_postgrest_filter_ops.ex
```diff
@@ -0,0 +1,673 @@
...
+        end loop;
+
+        perform set_config('role', null, true);
+    end;
+    $$;
+    """)
+
+    # The 5-arg check_equality_op is a new signature created after SetupSupabaseRealtimeAdmin ran,
+    # so its ownership reassignment is not covered there. Align its owner with the existing 4-arg
+    # overload so the realtime schema stays single-owner under the least-privilege setup.
```
edgurgel, line 650: @leandrocp I assume you reviewed this bit?

edgurgel (approved):

PR #1968, opened 2026-06-17
fix: validate join payload - REAL-866

leandrocp:


lib/realtime_web/channels/payloads/presence.ex
```diff
@@ -8,8 +8,8 @@ defmodule RealtimeWeb.Channels.Payloads.Presence do
   alias RealtimeWeb.Channels.Payloads.FlexibleBoolean
 
   embedded_schema do
-    field :enabled, FlexibleBoolean, default: true
-    field :key, :any, default: UUID.uuid1(), virtual: true
+    field :enabled, FlexibleBoolean, default: false
```
leandrocp, line 11: Changed to `false` because the old code was actually defaulting to `false` on invalid or missing params:

```elixir
    presence_enabled? =
      case get_in(params, ["config", "presence", "enabled"]) do
        enabled when is_boolean(enabled) -> enabled
        _ -> false
      end
```

If the intention is to enable by default I can change to `true` and adjust the rest.

lib/realtime_web/channels/payloads/join.ex
```diff
@@ -33,9 +33,9 @@ defmodule RealtimeWeb.Channels.Payloads.Join do
   end
 
   def presence_enabled?(%__MODULE__{config: %Config{presence: %Presence{enabled: enabled}}}), do: enabled
-  def presence_enabled?(_), do: true
+  def presence_enabled?(_), do: false
```
leandrocp, line 36: Keep the same behavior as https://github.com/supabase/realtime/pull/1968/changes#r3430411589

edgurgel (approved):
LGTM. Nice one

PR #1969, opened 2026-06-17
fix: ensure presence.read permission is respected

edgurgel:
* Fix insert to not use `returning` to avoid checking select policy
* Ensure `presence.read` is properly accounted for on RealtimeChannel and MessageDispatcher

### DIFF
--- a/lib/realtime/tenants/repo.ex
+++ b/lib/realtime/tenants/repo.ex
@@ -36,7 +36,7 @@ defmodule Realtime.Tenants.Repo do
   @doc """
   Inserts a given changeset into the database and converts the result into a given struct.
 
-  Mirrors `Ecto.Repo.insert/2`'s `:returning` option. Pass `returning: false` to omit
+  Partially mirrors `Ecto.Repo.insert/2`'s `:returning` option. Pass `returning: false` to omit
   `RETURNING *` and return `{:ok, struct}` built from the fields that were sent, instead
   of the row loaded from the database. This matters for authorization probes: with
   `RETURNING *`, Postgres enforces the SELECT (read) policy on the returned row, which
--- a/lib/realtime/tenants/repo.ex
+++ b/lib/realtime/tenants/repo.ex
@@ -34,7 +34,14 @@ defmodule Realtime.Tenants.Repo do
   end
 
   @doc """
-  Inserts a given changeset into the database and converts the result into a given struct
+  Inserts a given changeset into the database and converts the result into a given struct.
+
+  Mirrors `Ecto.Repo.insert/2`'s `:returning` option. Pass `returning: false` to omit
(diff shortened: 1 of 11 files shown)


filipecabaco (approved):

lib/realtime/tenants/repo.ex
```diff
@@ -34,7 +34,14 @@ defmodule Realtime.Tenants.Repo do
   end
 
   @doc """
-  Inserts a given changeset into the database and converts the result into a given struct
+  Inserts a given changeset into the database and converts the result into a given struct.
+
+  Mirrors `Ecto.Repo.insert/2`'s `:returning` option. Pass `returning: false` to omit
```
leandrocp: Nitpick but I think "mirror" is not exactly correct because Ecto takes a list of fields and here it only accepts true/false.

lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
```diff
@@ -7,8 +7,16 @@ defmodule RealtimeWeb.RealtimeChannel.MessageDispatcher do
...
-  def fastlane_metadata(fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids \\ MapSet.new()) do
-    {:rc_fastlane, fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids}
+  def fastlane_metadata(
+        fastlane_pid,
+        serializer,
+        topic,
+        log_level,
+        tenant_id,
+        replayed_message_ids \\ MapSet.new(),
+        presence_read? \\ true
```
leandrocp, line 17: My understanding is that `presence_read?` becomes a cached value so I tried to understand where this should be invalidated/updated, especially if it changes from `true` to `false`. I can't tell for sure because it's my first time reading this workflow but it would be nice to double check. That would be my only concern with this PR.

leandrocp (approved):

lib/realtime/tenants/repo.ex
edgurgel: true. "Partially mirrors" 😆

lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
edgurgel: It's "cached" only if we know upfront the value. We leave it as `nil` if we haven't checked for it.

But saying that now one case I didn't cover was the case when we a receive new access token (say after a refresh). 

https://github.com/supabase/realtime/blob/a2ddcc52f79aca5fb17aeaefc351734559fa3810/lib/realtime_web/channels/realtime_channel.ex#L493

But I think I will follow-up with a separate PR if that's ok.

leandrocp: Yep that's fine

PR #1970, opened 2026-06-18
feat(seeds): allow overriding tenant ssl_enforced via DB_SSL_ENFORCED env

Kampe:
## Summary

`priv/repo/seeds.exs` hardcodes `\"ssl_enforced\" => false` on the bundled self-host tenant's `postgres_cdc_rls` extension settings. Make it configurable via a `DB_SSL_ENFORCED` env var, defaulting to `false` so existing self-hosters see no behavior change.

## Motivation

The seed runs on every container start in self-hosted setups (typically `/app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'` from the entrypoint or chart args). It deletes the existing tenant and recreates it with `ssl_enforced=false` every time — overwriting any value an operator set via the management API or directly in `_realtime.extensions.settings`.

For managed Postgres backends that require TLS (e.g. AWS RDS with `rds.force_ssl=1`, GCP Cloud SQL with **Require SSL/TLS connections**), the non-TLS tenant CDC connection is rejected with:

```
FATAL 28000 (invalid_authorization_specification) no pg_hba.conf entry for host \"…\", user \"…\", database \"…\", no encryption
```

…and the tenant migration step then logs `Failed to set-up tenant: {:error, :shutdown}`. The Realtime app continues running but every per-tenant CDC subscription fails the same way until an operator manually updates `_realtime.extensions.settings.ssl_enforced` — which the next pod restart promptly reverts.

I reproduced this on real prod (AWS RDS Postgres 17 with `rds.force_ssl=1`) and worked around it with a pg_cron job that flips `ssl_enforced=true` every minute, but that's a workaround for a missing knob, not a real fix.

## The change

One file (`priv/repo/seeds.exs`):

```diff
 tenant_name = System.get_env(\"SELF_HOST_TENANT_NAME\", \"realtime-dev\")
 default_db_host = \"host.docker.internal\"
+
+db_ssl_enforced =
+  System.get_env(\"DB_SSL_ENFORCED\", \"false\") |> String.downcase() |> String.trim() in [\"true\", \"1\"]
```

```diff
             \"poll_max_record_bytes\" => 1_048_576,
-            \"ssl_enforced\" => false
+            \"ssl_enforced\" => db_ssl_enforced
```

Plus a row in `ENVS.md` next to the existing `DB_SSL` / `DB_SSL_CA_CERT` rows.

## Naming

The new env is **`DB_SSL_ENFORCED`** — deliberately distinct from the existing `DB_SSL`:

- `DB_SSL` controls Realtime's own connection to the metadata DB.
- `DB_SSL_ENFORCED` controls per-tenant CDC connections by populating the `ssl_enforced` field on the tenant extension settings.

These are separate code paths that already exist as separate concerns in the model — this just lets seeders set both from env.

## Out of scope (worth a separate follow-up)

`seeds.exs` deletes and recreates the tenant on every container start, which clobbers any tenant fields edited via the management API — `ssl_enforced` being the example that motivated this PR, but also `jwt_secret`, `max_concurrent_users`, etc. An idempotent seed (skip if tenant exists, or upsert specific fields) would be a meaningful improvement, but is a behavior change that deserves its own PR.

## Risks

- **No schema migration.** Pure script-level env read; no schema changes.
- **Default is `\"false\"` (string).** Behavior identical for existing self-hosters who don't set the env.
- **No new tests required.** This is a single env read at seed time; the assertion that `ssl_enforced` flows through to the tenant connection is already covered by existing tenant-config tests under `test/realtime/tenants_test.exs` and friends.


priv/repo/seeds.exs
```diff
@@ -9,6 +9,18 @@ alias Realtime.Tenants
...
 default_db_host = "host.docker.internal"
 
+# Tenant per-CDC ssl_enforced flag. Distinct from DB_SSL (which controls
+# Realtime's connection to its own metadata DB) — this flips whether
+# tenant CDC connections use TLS. Defaults to false to preserve existing
+# behavior; set to "true" or "1" when seeding against a managed Postgres
+# that requires TLS (e.g. AWS RDS with rds.force_ssl=1, GCP Cloud SQL
+# with "require SSL/TLS connections" on).
+db_ssl_enforced =
+  System.get_env("DB_SSL_ENFORCED", "false")
```
edgurgel, line 19: Could we call this SELF_HOST_DB_SSL_ENFORCED or something like this so it's clear that this is just for this seed?

leandrocp: What if we reuse `DB_SSL`? In that env both DBs will most likely enforce SSL right? If so the whole PR is just one line change:

```elixir
"ssl_enforced" => Realtime.Env.get_boolean("DB_SSL", false)
```

@Kampe @edgurgel wdyt?

leandrocp (approved):

leandrocp:
Hi @Kampe I've changed it to reuse `DB_SSL` so both the Realtime DB and Tenant DB share the same config. Please let us know if that's not enough or if you need to configure each one separately. Thanks!

PR #1973, opened 2026-06-22
fix: improve parity with Postgrest filters

filipecabaco:

We now handle special chars closer to what PostgRest does so we have a clear direction on how we want to support filters going forward


edgurgel (approved):
LGTM

PR #1977, opened 2026-06-24
feat: Forum.Muster

edgurgel:

Introduces `Forum.Muster` which should help reducing nodes targeted for a broadcast message.

Phoenix PubSub works by keeping track of subscription locally on every node and a broadcast always goes to every node. For Realtime this is not ideal because we host many many tenants which might be in a small subset of the cluster. Whenever a new node is added more work is added to the cluster as there is one extra node to be communicated with each broadcast. Imagine you have 40 nodes but only 2,3 websockets connected for a specific tenant. Broadcasting to 40 nodes for every message adds a lot of network messages being exchanged and wasted CPU. While we could simply send the first message to the "router" node and from there broadcast to the 2,3 websockets.

Realtime wants to only "try" to broadcast to a node if there is at least 1 websocket connected to the tenant that the broadcast originated from. We will pay the price in advance of updating the router node with the payoff of avoiding broadcast messages flooding the cluster.

### Goal

`Forum.Muster` is there to provide a way for a single node (in Realtime's case per region) to be responsible to answer the question: "which nodes have at least one process belonging to a certain group". This enables Realtime to send messages to less nodes for each broadcast.

Once `Muster.join/3` returns `:ok` it should receive a broadcast message that was sent right after the join finished.

Muster is fine with false positives where a router node thinks that a group is occupied but it's not anymore (we just drop the message like today). It should **never** have false negatives where a router node thinks a group is vacant but it's not.

We rely on [consistent hashing](https://github.com/discord/ex_hash_ring) to reduce the amount of groups moving around and keeping rebalances as light as possible. 

### High level execution

When a local member joins the group on node A it sends a **synchronous** call to the router node when the group becomes occupied. Subsequent process from node A joining will not need to notify the router so only the first join pays the price of contacting the router node and we keep track of what we have already communicated.

When the last member of the group leaves (after a cooldown) the router node is notified that the group is vacant.

When a broadcast happens the caller asks `Muster.router/2` for the group's router node and routes the message there (in our case probably adapting the PubSub adapter). The router then fans out to the member nodes. While the cluster view is in flux, `router/2` returns `{:rebalancing, members}` and the caller fans out to every member instead.

From the router node we then call `Muster.targets/3` which can return `{:ok, nodes}` or `{:error, :flood}`. It returns `flood` when there is a rebalance happening and the router can't be trusted to answer the question.

When a node joins or leaves the cluster a rebalance takes place which causes routing to not be trustworthy until all nodes agree on the new membership. `Muster.join/3` and `Muster.leave/3` keep working regardless if there is a rebalance taking place or not.

For more details on each of the mentioned behaviours check `forum/README.md`.

### Testing

We are heavily using [Snabbkaffe](https://github.com/kafka4beam/snabbkaffe) to force some specific timings and crashes to exercise most failure scenarios.


forum/lib/forum/muster.ex
leandrocp, line 1: @edgurgel I did the following test:

1. start muster
2. join
3. kill the ring

What's expected in this scenario? In my test the ring didn't recover but I was expecting `Muster` to restart it.

```elixir
test "ring recovery" do
  scope = :"muster_#{System.unique_integer([:positive])}"
  ring_name = :"#{scope}_muster_ring"
  self = node()

  muster = start_supervised!(%{
    id: {Muster, scope},
    start: {Muster, :start_link, [scope, []]},
    type: :supervisor
  })

  pid = spawn(fn -> Process.sleep(:infinity) end)

  assert :ok = Muster.join(scope, "room:1", pid)

  # single node, route to itself
  assert {:ok, ^self} = Muster.router(scope, "room:1")

  # kill the ring
  ring = Process.whereis(ring_name)
  ref = Process.monitor(ring)
  assert Process.exit(ring, :kill)
  assert_receive {:DOWN, ^ref, :process, ^ring, :killed}

  Process.sleep(5_000)

  assert is_pid(muster) #-> pass
  assert {:ok, [^self]} = ExHashRing.Ring.get_nodes(ring_name) #-> fails
  assert {:ok, ^self} = Muster.router(scope, "room:1") #-> fails
end
```

leandrocp (commented):
The idea of fallbacking to fan-out to everyone during rebalancing is genius IMHO.

forum/lib/forum/muster.ex
edgurgel: Good catch. I've made the ring a rest_for_one and it will crash the Scope and Shards if the ring crashes 👍

josevalim:
Hi @edgurgel! This is very exciting. I have some thoughts while doing a broad reviewing of the code and documents, so I thought I'd write some of them down. Please do not act on these changes immediately, I just thought I would externalize possible trade-offs and alternatives based on what I am reading. If these are mostly noise, I apologize.

* Should you use consistent hashing vs [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)? They are quite similar but rendezvous is simpler to implement but compute time is linear based on the number of nodes N. It may be worth if the number of nodes is low (let's say 10).

* Should you merge census and muster? Muster needs to know which nodes are members of a topic which is a subset of what census needs to know: which is how many subscribers there are in a given topic. However, IIRC census is not immediate, it is a broadcast in an interval. You could have a mixed design: census immediately notifies new topics around the cluster but updates to each topic count are still based on a timer. This design is chattier but it means you don't need router nodes (every node has everything to broadcast, which means fewer hops during dispatch time)

* Flooding the network can be expensive, especially if it happens during network partitions. The cost of recomputing and migrating the tables can be expensive too. So the question of when to rebalance matters too:
  * Instead of flooding the network, fallback to sending to everyone based on census. But this requires you to evaluate the trade-offs mentioned in the bullet item above. In a way, this is less precise, but flood is also vulnerable to network partitions too (the node flooding may only have a partial view of the network)
  * When do you trigger rebalancing? If every time a node joins or leaves triggers rebalancing, I am worried that temporary network partitions will cause further instability. One option is to not immediately rebalance when a node is down and only after a timeout, but what happens when the router itself is down? You can fix this by assigning two router nodes per group, then you can route to the second one in such cases. This doubles the registration cost though

* How do the nodes agree they are seeing the same list? My understanding is that you compute a hash over the node list and decide they disagree on the hash but that assumes there is something else making sure all the nodes connect (such as libcluster or similar). If that's the case, it may be worth documenting it! Especially because if the network view is fragmented, it seems muster does not attempt to fix it by itself (please correct me if I'm wrong)

Another possible design is to separate membership from routing and have a leader/quorum publish explicit list of nodes. This means all nodes can agree on what nodes you should route too. Another benefit is that you can tie each list of nodes to epochs, so when you are rebalancing with epoch 40, you can still say "this messages uses epoch 39 for routing" and avoid floods. On the downside, it introduces a stronger coordination dependency and you need to decide what happens if a node cannot reach the leader/quorum, such as falling back to flooding.

I will prepare a cup of tea and be back for a more in-depth code review. :)

josevalim:
Oh, one additional remark for now, I believe putting census and muster together would be similar to the design Chris took with [group](https://github.com/phoenixframework/group).

edgurgel:
Hi @josevalim! Thanks for the initial comments! Happy to have you reviewing this 😊 

> Should you use consistent hashing vs [rendezvous hashing](https://en.wikipedia.org/wiki/Rendezvous_hashing)? They are quite similar but rendezvous is simpler to implement but compute time is linear based on the number of nodes N. It may be worth if the number of nodes is low (let's say 10).

Will __definitely__ check this out. But our number of nodes can vary wildly per region (Muster will run a Scope per region) so it can be 4 nodes or 40.

> Should you merge census and muster? Muster needs to know which nodes are members of a topic which is a subset of what census needs to know: which is how many subscribers there are in a given topic. However, IIRC census is not immediate, it is a broadcast in an interval. You could have a mixed design: census immediately notifies new topics around the cluster but updates to each topic count are still based on a timer. This design is chattier but it means you don't need router nodes (every node has everything to broadcast, which means fewer hops during dispatch time).

I'm glad you said that because I think that long term Muster can absorb Census but I decided to keep them separate for now so that we can focus on the routing/rebalance/etc.

What we want from Muster is that if the `join` return `:ok` we can be 100% sure that a broadcast message will arrive to that topic without any arbitrary wait hence the sync call to the router and having a single place to route from. If we just broadcast that the channel is occupied we won't know fore sure if the broadcast will arrive (at least not from which moment for a certain node). The message that the channel is occupied could take some time to be processed so maybe 20 nodes got that info in 100ms and another 10 got it just after 500ms. A message that was broadcasted before the 100ms will not show up anywhere, after 100ms will definitely show up on websockets for at least 20 nodes and so on. 

> Flooding the network can be expensive, especially if it happens during network partitions. 

Yes and that's what we do today for every message so we know that it's expensive but if it's just momentarily we hope that this will suffice.

> Instead of flooding the network, fallback to sending to everyone based on census. But this requires you to evaluate the trade-offs mentioned in the bullet item above. In a way, this is less precise, but flood is also vulnerable to network partitions too (the node flooding may only have a partial view of the network).

Yes this is definitely an option that we can take! That's why Muster doesn't handle broadcasting messages. It leaves to the user to decide what to do once it knows that it can't be trusted.

> When do you trigger rebalancing? If every time a node joins or leaves triggers rebalancing, I am worried that temporary network partitions will cause further instability. One option is to not immediately rebalance when a node is down and only after a timeout, but what happens when the router itself is down? You can fix this by assigning two router nodes per group, then you can route to the second one in such cases. This doubles the registration cost though

Yeah this is a great question. I think we can decide to delay the "nodedown"/DOWN hoping that the node reconnects etc. But when I started working on this delay I noticed how much more complicated rebalancing became so I decided to re-evaluate this later.

And yes it should be possible to have the concept of N "replicas" but as you said the registration cost increases so not sure if it's worth it given that we have a fallback (fan-out) option.

> How do the nodes agree they are seeing the same list? My understanding is that you compute a hash over the node list and decide they disagree on the hash but that assumes there is something else making sure all the nodes connect (such as libcluster or similar). If that's the case, it may be worth documenting it! Especially because if the network view is fragmented, it seems muster does not attempt to fix it by itself (please correct me if I'm wrong)

Yeah. Realtime will use libcluster but Muster doesn't enforce a convergence there. It just trusts that the sub-cluster running the Muster scope is fully connected.

And yes we use that hash to ensure that whoever is sending the message to the router has the same view (account for a rebalance after the message left the sender on the way to the router) and after rebalance we only consider that the rebalance is done and the nodes are "ready" once they all sent the same view hash. 

I'm not sure what Muster could do? Maybe after X seconds/minutes if it doesn't converge crash? 

> Another possible design is to separate membership from routing and have a leader/quorum publish explicit list of nodes. This means all nodes can agree on what nodes you should route too. Another benefit is that you can tie each list of nodes to epochs, so when you are rebalancing with epoch 40, you can still say "this messages uses epoch 39 for routing" and avoid floods. On the downside, it introduces a stronger coordination dependency and you need to decide what happens if a node cannot reach the leader/quorum, such as falling back to flooding.

Right. I did look at using something like Raft to have a subset of the nodes to become routers but I went with a more decentralized way so that we don't have nodes receiving way more messages than others and potentially overloading these routers. 


> Oh, one additional remark for now, I believe putting census and muster together would be similar to the design Chris took with [group](https://github.com/phoenixframework/group).

Oh I will definitely have a look again at `group` but from memory it would not work like Muster where after a `join` it would 100% deliver a broadcast to that group due to the async nature of spreading that information but I could be mistaken cause I spent just ~15 minutes looking at the code maybe a month ago.

josevalim:
> Yes and that's what we do today for every message so we know that it's expensive but if it's just momentarily we hope that this will suffice.

This is a very good point and it changes a lot of what I said above. This should be strictly an improvement, so it makes sense to ship it with strict defaults (like view hashes should match) and then measure its impact and failure points.

> And yes we use that hash to ensure that whoever is sending the message to the router has the same view (account for a rebalance after the message left the sender on the way to the router) and after rebalance we only consider that the rebalance is done and the nodes are "ready" once they all sent the same view hash. I'm not sure what Muster could do? Maybe after X seconds/minutes if it doesn't converge crash?

For now logging telemetry event for flood is probably enough. Also make sure to distinguish between the different sorts of floods (the client is rebalancing, the server is rebalancing, etc).

> Right. I did look at using something like Raft to have a subset of the nodes to become routers but I went with a more decentralized way so that we don't have nodes receiving way more messages than others and potentially overloading these routers.

Yeah, I had the same thought and concluded some nodes would then become hotspots. So my second thought, the one above, is to use leader election to have everyone agree on the topology (but only if it turns out you are flooding too much).

> where after a join it would 100% deliver a broadcast to that group due to the async nature of spreading that information

I believe you are correct!

forum/lib/forum/muster/scope.ex
```diff
@@ -0,0 +1,1217 @@
...
+    state = reannounce_local_groups_at_init(state)
+    # Above the seqs of the rows just re-announced, so receivers may judge them
+    # once this watermark is announced.
+    state = %{state | view_seq: next_seq()}
+
+    {:ok, state, {:continue, :discover}}
+  end
+
+  @impl true
+  def handle_continue(:discover, state) do
```
josevalim, line 482: I think there is a subtle behaviour/race here: you are racing the discovery with the shards startup. This means if another node sends a discovery to you, that may trigger a rebalance before all shards are ready (which will either crash because you can't talk to the shard or leave them in inconsistent state). I am not exactly sure what is the fix here... perhaps have a process that does the discovery broadcast that runs after the shards start? I am assuming that this node can only be reached to after this initial discovery broadcast (which may be a wrong assumption).

forum/lib/forum/muster/scope.ex
```diff
@@ -0,0 +1,1217 @@
...
+
+  ## Peer/membership
+
+  defp register_peer(state, peer) when is_pid(peer) do
+    cond do
+      node(peer) == node() ->
+        # Ignore self-discovery (loopback from our own broadcast).
+        state
+
+      Map.has_key?(state.peers, peer) ->
```
josevalim, line 1317: This is keyed by process and not node name, so I believe you are assuming that a PID can uniquely identify a node. That's actually going to be true for nodedown/nodeup ([source here](https://www.erlang.org/docs/26/man/net_kernel.html#monitor_nodes-2)) but that's not going to be true for crashes. If PID crashes due to a bug, then you may receive the broadcast of a new pid for node A before you receive its DOWN message.

It would go like this:

* scope crashes
* its supervisor receives a DOWN message and starts a new scope
* a new scope is started
* the DOWN message is delivered to the remote node just now

It is very unlikely but, in a nutshell, the delivery of DOWN messages is not guaranteed to be atomic. I have been bitten by this bug once or twice in my wife. :D

This can then cause issues in a few places where you are doing node(pid), because you may temporarily have two pids mapping to the same node.

josevalim:
I have pushed some comments and I will do another review tomorrow to see if I spot anything I missed!

forum/lib/forum/supervisor.ex
```diff
@@ -31,31 +60,146 @@ defmodule Forum.Supervisor do
...
+  #
+  # The ring stores its node set in a process-owned ETS table, and the coordinator
+  # is the sole writer of that set (Ring.set_nodes, at init and on rebalance). So a
+  # ring crash brings the set back EMPTY with nothing to re-seed it until the next
+  # membership change. :rest_for_one fixes that: the ring is the dependency root, so
+  # its crash restarts the inner unit (coordinator + shards) after it, and the
+  # coordinator's init re-seeds the ring. A coordinator crash stays contained in the
+  # inner :one_for_one (the ring, sitting before the inner child, is untouched — so
+  # a coordinator restart still does not take the ring down under the shards that
+  # read it directly); a single shard crash restarts only that shard.
```
josevalim: I am a bit worried about coordinator being one_for_one. The ring is shared state, which means that, if the scope/coordinator crashes, it will reset the ring, and it means the ring state may change at any time within the shard execution. Is there a reason to not be a bit more conservative and have all shards crash? effectively:

```
- ring
- coordinator
- supervisor
```

The fact they share state (ring, persistent term, etc) should advocate for stronger restart strategies.

josevalim:
Hi @edgurgel! No additional comments from me. I have two final considerations:

1. Indeed it may be worth looking at leader election to force all nodes to agree on a topology, then hashing happens on this shared view. If the node you are supposed to hash to or the router mismatches, you flood as today. Nothing to do for now, but in the future in case flooding is too frequent. RabbitMQ has a good library for Raft although leader election may impose restrictions during autoscaling and minimum number of nodes. There would also be a benefit of using a well established library/protocol.

2. Since you have an extensive test suite structure, it may be worth getting Codex on extra high thinking to find bugs and write failed test cases for reproduction. Codex is really good for adversarial testing. In case you are in the list of companies that can use Fable/Sol, I have heard they are even better!

forum/lib/forum/muster/scope.ex
edgurgel: Oh wow it makes total sense as the senders are different so they can overlap in any way.

I will address this. I think Census has the same issue.

forum/lib/forum/muster/scope.ex
edgurgel: Oh yup definitely have to fix this.

forum/lib/forum/supervisor.ex
edgurgel: Yeah good point.  

Maybe the right way is to simple have them on `one_for_all` and be done with it? The ETS tables survive and the state can be reconstructed etc.

edgurgel:
>     1. Indeed it may be worth looking at leader election to force all nodes to agree on a topology, then hashing happens on this shared view. If the node you are supposed to hash to or the router mismatches, you flood as today. Nothing to do for now, but in the future in case flooding is too frequent. RabbitMQ has a good library for Raft although leader election may impose restrictions during autoscaling and minimum number of nodes. There would also be a benefit of using a well established library/protocol.

Yeah I will definite have a look and see how it much it might simplify Muster. I actually started with `ra` and ditched early on for reasons that I can't remember!


>     2. Since you have an extensive test suite structure, it may be worth getting Codex on extra high thinking to find bugs and write failed test cases for reproduction. Codex is really good for adversarial testing. In case you are in the list of companies that can use Fable/Sol, I have heard they are even better!

Thanks will do that! Claude Code + Snabbkaffe have been great to uncover weird issues that only forcing some order or injecting crashes at specific moments would show up

forum/lib/forum/supervisor.ex
edgurgel: BTW reviewing it again do you think that the current tree doesn't cover the ring reset?

So we have it like this: 

```
  #   outer (:rest_for_one)
  #   ├── ring                       # node set lives in the ring's own ETS table
  #   └── inner (:one_for_one)
  #       ├── coordinator (scope)    # the ONLY writer of the ring's node set
  #       └── shard_0 .. shard_n     # read the ring directly

```

When the ring crashes everything else restarts.

The ring reset on coordinator `init/1` (`Ring.set_nodes(ring, [node()])`) and it's paired with the `:status/:view_hash` handshake. While the ring is in that state `can_decide?/2` returns `false` for any sender whose view doesn't match, so routers fall back to flooding instead of trusting a stale occupancy table. Rebalances will naturally occur once the nodes are re-discovered, the ring gets updated, the shards get updated and everything should continue fine.

Thoughts? could you elaborate more on what issue you might be seeing?

josevalim: Good question. I should definitely expand on this because it is a theoretical concern. There are three overlapping concerns here:

1. Global mutable state: for example, using ets and persistent term means that if you call `can_decide?` twice in the same function, it may return different result

2. There is state that is stored in the supervisor which means that, if any of the crash, you need to reevaluate which of this state is still valid

3. When there is a crash, you do a partial restart of the supervision tree

The combination of the three means, when there is a crash (3), it is hard to reevaluate which state is still valid (2) and any attempt to reset or prune state can cause concurrency bugs that are very subtle to detect (1). So when reviewing the shard code, for each GenServer callback, there is a new failure mode I need to consider and explore: what will happen if the coordinator crashes at any point between this function. This is somewhat easy to do but answer the question: what happens if the coordinator crashes between changes to occupied_pending/vacant_flushing/occupied/etc are very hard!

Therefore, if you remove 3, then it both reduces changes to global state (1) and makes reevaluating which state is valid (and the overall failure modes) on crashes much simpler (2). You could even enforce more of this decision in the design: have only the data that should survive restarts in the supervisor but everything else belongs to shards.

I hope this makes sense. :)

edgurgel: Yup it does! Yeah I think I will err to the conservative approach and make so that if the coordinator goes down, all shards are restarted.

forum/lib/forum/muster/scope.ex
edgurgel: Ok I ended up using a process that starts after all shards and messages Scope to say that all processes are up and running:

https://github.com/supabase/realtime/pull/1977/changes/55698e94ba389277912251f48366c7e783efed58

I think this solves the race?

forum/lib/forum/supervisor.ex
edgurgel: Ok now we have a rest_for_one with an inner supervisor holding the shards:

https://github.com/supabase/realtime/pull/1977/changes/406049467c529131761cea9cec9fbb4d80eff4df

```
:rest_for_one
├── ring                       # node set lives in the ring process's own ETS table
├── coordinator (scope)        # the ONLY writer of the ring's node set
├── shards_supervisor (:one_for_one)
│   ├── shard_0                # read the ring directly; durable ETS claim state
│   ├── shard_1
│   └── shard_n
└── shards_ready_sentinel      # :ignore's every start; no pid
```

forum/lib/forum/muster/scope.ex
edgurgel: I'm still working through this one as I also noticed that there is a couple of possible bugs depending on when DOWN arrives after a scope restarts. Thanks for pointing this one out.

edgurgel:
> Since you have an extensive test suite structure, it may be worth getting Codex on extra high thinking to find bugs and write failed test cases for reproduction. Codex is really good for adversarial testing. In case you are in the list of companies that can use Fable/Sol, I have heard they are even better!

BTW I asked Fable to build a TLA+ model based on Muster's docs, code and tests and it has found 2-3 bugs that I would've not found otherwise. One of the bugs was actually related to the late `DOWN` you mentioned. 

Working on those ATM

forum/lib/forum/muster/scope.ex
edgurgel: @josevalim finally got it working: https://github.com/supabase/realtime/commit/5755b9446c817bc4e544bd2607e78114c6bf6209#diff-b5d395e96cfb38be4bddaf790e85f0afe88f53c1b3d08aba4c46f04b740ada68

Essentially keeping track of which `pid` wrote which rows. We don't drop/tombstone rows just based on nodes now. Also got a test showing the problem by forcing a specific order of events.

So we only get rid of data (occupancy, member views etc) if the pid had a DOWN message.

Otherwise we don't delete. Any new information will have a new pid and it won't be deleted when the DOWN finally arrives if the new information has a new pid already.

josevalim (commented):
The updates look good to me but from this point on you should probably trust the TLA+ model much more than you trust me. :)

forum/mix.exs
```diff
@@ -27,6 +27,11 @@ defmodule Forum.MixProject do
   defp deps do
     [
       {:telemetry, "~> 1.3"},
+      {:ex_hash_ring, "~> 6.0"},
+      # Trace-based testing. The Elixir `Snabbkaffe` interface (lib/snabbkaffe.ex)
+      # wraps these Erlang macros so trace points are discarded in non-test builds.
+      {:snabbkaffe, "~> 1.0"},
```
leandrocp, line 33: Add in root's `mix.exs` to fix CI, otherwise release will fail.

```elixir
{:snabbkaffe, git: "https://github.com/kafka4beam/snabbkaffe", tag: "1.0.10", override: true},
```

https://github.com/emqx/gen_rpc/blob/891f90d713e83e3fca049345fb641afd9a1def28/rebar.config#L32

forum/test/forum/muster/shard_test.exs
```diff
@@ -0,0 +1,333 @@
...
+    # Membership and claim state are reconciled back from the durable tables.
+    assert Shard.member_count(scope, :group1) == 1
+    assert Shard.member_count(scope, :group2) == 1
+    assert Shard.member?(scope, :group1, pid1)
+    assert Shard.member?(scope, :group2, pid2)
+
+    states = group_states(new_shard)
+    assert states[:group1] == :occupied
+    assert states[:group2] == :occupied
+  end
```
leandrocp, line 332:
```suggestion
  end
  
  test "recover from a restart when the entries table is poisoned", %{
    scope: scope,
    shard: shard,
    opts: opts
  } do
    pid1 = spawn(fn -> Process.sleep(:infinity) end)
    assert :ok = join(shard, :group1, pid1)
    entries_table = Supervisor.partition_entries_table(Supervisor.partition_name(scope, 0))

    # force invalid data
    :ets.insert(entries_table, {:poison})

    ref = Process.monitor(shard)
    Process.exit(shard, :kill)
    assert_receive {:DOWN, ^ref, :process, ^shard, :killed}

    spec = %{
      id: Supervisor.shard_name(scope, 0),
      start: {Shard, :start_link, [scope, 0, opts]},
      type: :worker,
      restart: :temporary
    }

    new_shard = start_supervised!(spec)

    assert Process.alive?(new_shard)

    monitored = get_in(Process.info(new_shard, [:monitors]), [:monitors, :process])
    assert pid1 == monitored
    assert %{{:group1, ^pid1} => _} = :sys.get_state(new_shard).monitors

    assert Shard.member_count(scope, :group1) == 1
    assert Shard.member?(scope, :group1, pid1)
    assert group_states(new_shard)[:group1] == :occupied
    assert Shard.groups(scope) == [:group1]

    pid2 = spawn(fn -> Process.sleep(:infinity) end)
    assert :ok = join(new_shard, :group2, pid2)
    assert Shard.member?(scope, :group2, pid2)
    assert %{group1: :occupied, group2: :occupied} = group_states(new_shard)
  end
```

forum/lib/forum/muster/shard.ex
```diff
@@ -0,0 +1,985 @@
...
+  # surviving entry. A pid that died while we were down is monitored here and its
+  # immediate DOWN (queued after init) drives the normal removal + vacancy path.
+  defp rebuild_membership(state) do
+    monitors =
+      state.entries_table
+      |> :ets.tab2list()
+      |> Enum.reduce(%{}, fn {{group, pid}}, acc ->
+        ref = Process.monitor(pid, tag: {:DOWN, group})
+        Map.put(acc, {group, pid}, ref)
+      end)
```
leandrocp, line 230: This would crash on the very unlikely scenario of invalid rows in the entries table. This might be too defensive so it's up to you analyzing if that makes sense. The test https://github.com/supabase/realtime/pull/1977/changes#r3560427031 shows it can recover ignoring the invalid rows.

```suggestion
      |> Enum.reduce(%{}, fn
        {{group, pid}}, acc when is_pid(pid) ->
          ref = Process.monitor(pid, tag: {:DOWN, group})
          Map.put(acc, {group, pid}, ref)

        _row, acc ->
          acc
      end)
```

leandrocp (changes requested):

leandrocp (approved):

PR #1986, opened 2026-06-29
chore: add orioledb to testing matrix

filipecabaco:

Add oriole to CI testing matrix


leandrocp (approved):
Even tho it's failing I think we should merge it then fix the migration(s) to make it pass so we can unblock deployments.

filipecabaco:
I agree if anything this is a good way for us to fix this issue and flag it to oriole also

PR #1993, opened 2026-06-30
fix: restrict the realtime schema

leandrocp:
- Apply schema restrictions when supautils grants are available to avoid unintentional and unexpected changes to realtime schema that's supposed to be protected.
- Always connect as `db_user` and removes `db_user_realtime`. The restriction is controlled by migrations only.
- Remove feature flag `use_supabase_realtime_admin`

Ref REAL-25, REAL-527, REAL-630, REAL-631, REAL-773, REAL-778
Task SEC-562 is still open.


dev/postgres/zb-supabase-schema.sql
```diff
@@ -3,7 +3,7 @@
 do $$
 begin
   if not exists (select from pg_roles where rolname = 'supabase_realtime_admin') then
-    create user supabase_realtime_admin noinherit createrole login replication password 'postgres';
+    create role supabase_realtime_admin with noinherit nologin noreplication;
```
leandrocp, line 6: Same as migrations.

lib/realtime/feature_flags/cache.ex
```diff
@@ -26,13 +26,13 @@ defmodule Realtime.FeatureFlags.Cache do
...
   @spec get_flag(String.t()) :: FeatureFlag.t() | nil
   def get_flag(name) do
-    with {_, value} <-
-           Cachex.fetch(__MODULE__, cache_key(name), fn _key ->
-             with %FeatureFlag{} = flag <- Api.get_feature_flag(name),
-                  do: {:commit, flag},
-                  else: (_ -> {:ignore, nil})
-           end) do
-      value
+    case Cachex.fetch(__MODULE__, cache_key(name), fn _key ->
```
leandrocp, line 29: This change is unrelated but I found it while testing and it was small enough to include in this PR. See the tests.

lib/realtime/tenants/repo/migrations/20260630120000_restrict_realtime_schema.ex
```diff
@@ -47,33 +36,29 @@ defmodule Realtime.Tenants.Migrations.SetupSupabaseRealtimeAdmin do
...
     END $$;
     """)
 
-    execute("REVOKE CREATE ON SCHEMA realtime FROM postgres")
-    execute("REVOKE ALL ON realtime.schema_migrations FROM anon, authenticated, service_role, postgres")
-    execute("GRANT USAGE ON SCHEMA realtime TO postgres WITH GRANT OPTION")
+    execute("REVOKE ALL ON realtime.schema_migrations FROM postgres, dashboard_user, anon, authenticated, service_role")
+
+    execute(
+      "REVOKE INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER ON realtime.schema_migrations FROM supabase_realtime_admin"
```
leandrocp: `SELECT` is still required but it's safe.

test/realtime/tenants/schema_test.exs
leandrocp, line 1: These are the most important assertions. Need to double check if those changes are not too restrictive.

lib/realtime/tenants/repo/migrations/20260630120000_restrict_realtime_schema.ex
edgurgel: why `supabase_realtime_admin` here matters?

edgurgel (approved):

lib/realtime/tenants/repo/migrations/20260630120000_restrict_realtime_schema.ex
leandrocp: Checked again and it doesn't. I thought it was impacting tests but nope so I added `supabase_realtime_admin`to the `REVOKE ALL ON realtime.schema_migrations` because only `supabase_admin` is supposed to use that table.

PR #1998, opened 2026-07-02
feat: load tenant database dump

leandrocp:
Shortcut to load the entire Tenant DB schema at once instead of running migrations sequentially that adds overhead to tenant creation workflow. Applies only to new tenants when `migrations_ran = 0 && empty schema_migrations`.

Log will show:

```
13:21:07.868 [info] Finished loading the dump of 81 migrations for tenant realtime-dev in 18ms file=lib/realtime/telemetry/logger.ex pid=<0.1375.0> project=realtime-dev external_id=realtime-dev
```

### Test

Run:
```sh
PGPASSWORD=postgres psql -h 127.0.0.1 -p 5433 -U supabase_admin -d postgres -c "DROP SCHEMA IF EXISTS realtime CASCADE;"
PGPASSWORD=postgres psql -h 127.0.0.1 -p 5433 -U supabase_admin -d postgres -f priv/repo/tenant_db_dump_17.sql
mix run priv/repo/dev_seeds.exs
mise run dev
```

Check: http://127.0.0.1:4000/admin/dashboard/tenant_migrations?external_id=realtime-dev

Ref REAL-916


lib/realtime/tenants/migrations.ex
```diff
@@ -249,6 +258,51 @@ defmodule Realtime.Tenants.Migrations do
...
+    end
+  end
+
+  defp fetch_pg_major(conn) do
+    query = """
+    SELECT current_setting('server_version_num'), EXISTS(SELECT 1 FROM pg_extension WHERE extname = 'orioledb')
+    """
+
+    case Postgrex.query(conn, query, [], timeout: @dump_timeout) do
+      {:ok, %{rows: [[_version_num, true]]}} -> {:error, :orioledb_not_supported}
```
leandrocp, line 317: Skip OrioleDB because it differs `default_table_access_method` and it's not worth adding it yet.

lib/realtime/tenants/migrations.ex
```diff
@@ -245,6 +259,74 @@ defmodule Realtime.Tenants.Migrations do
   defp error_code(%DBConnection.ConnectionError{}), do: :connection_error
   defp error_code(_), do: :other
 
+  @dump_timeout 30_000
+
+  defp load_db_dump(%Database{} = settings) do
+    with {:ok, conn} <- Database.connect_db(%{settings | pool_size: 1}),
```
edgurgel: So down on `schema_migrations_empty?` you execute `Repo.query` with `repo`. Why can't we do the same here and not need to use `Database.connect_db` ? 🤔

test/integration/tenant_db_dump_test.exs
```diff
@@ -0,0 +1,25 @@
...
+        send(pid, {:migrations_executed, metadata.migrations_executed})
+      end,
+      %{pid: self()}
+    )
+
+    on_exit(fn -> :telemetry.detach("tenant-db-dump-test") end)
+
+    assert Migrations.run_migrations(tenant) == :ok
+    assert_receive {:migrations_executed, executed}
+    assert executed == Enum.count(Migrations.migrations())
```
edgurgel: How can we be certain that this test is actually loading the dump and not running all migrations ? 🤔

leandrocp: Included the source on assert match to verify it was done by `:dump`

lib/realtime/tenants/migrations.ex
leandrocp: Good point, fixed it.

lib/realtime/telemetry/logger.ex
```diff
@@ -49,10 +49,16 @@ defmodule Realtime.Telemetry.Logger do
...
+    message =
+      case metadata.source do
+        :dump ->
+          "Finished loading the dump of #{metadata.migrations_executed} migrations for tenant #{metadata.external_id} in #{duration_ms}ms"
+
+        :migrator ->
+          "Finished applying #{metadata.migrations_executed} migrations for tenant #{metadata.external_id} in #{duration_ms}ms"
+      end
+
+    Logger.info(message, project: metadata.external_id)
```
edgurgel: For consistency we should also tag with `external_id`. I think we do this everywhere

leandrocp: Done

test/integration/tenant_db_dump_test.exs
edgurgel: Nice one

test/integration/tenant_db_dump_test.exs
```diff
@@ -0,0 +1,27 @@
+defmodule Realtime.Integration.TenantDbDumpTest do
```
edgurgel: Not a blocker but unsure why this is in a new file and not in the MigrationTest existing file. 

And if we need it do we need it to be async: false ?

edgurgel (approved):
I would like just to point out that from now on our tests will solely use "load dump" given that Containers use Migrations module

https://github.com/supabase/realtime/blob/9e657a2bdd79c8306d27e18214bc3d3510195eeb/test/support/containers.ex#L116

Not sure if we want to keep exercising running all migration steps as well?

LGTM

test/integration/tenant_db_dump_test.exs
leandrocp: Not really required tbh.

lib/realtime/tenants/migrations.ex
```diff
@@ -245,6 +259,72 @@ defmodule Realtime.Tenants.Migrations do
...
+  defp schema_migrations_empty?(repo) do
+    case Repo.query("SELECT count(*)::int FROM realtime.schema_migrations", [], dynamic_repo: repo) do
+      {:ok, %{rows: [[0]]}} ->
+        true
+
+      {:ok, %{rows: [[_count]]}} ->
+        false
+
+      {:error, %Postgrex.Error{postgres: %{code: :undefined_table}}} ->
+        true
```
edgurgel, line 295: Do we have a test for this specific error?

leandrocp: Added one now 👍🏻

test/integration/tenant_db_dump_test.exs
leandrocp: Moved into the existing MigrationTests 👍🏻

leandrocp:
> I would like just to point out that from now on our tests will solely use "load dump"

Actually only PG15+ load the tenant schema using this new DB dump, while PG14 and OrioledDB still use the sequential migrations even for new tenants (not changed). But I added a test to exercise the sequential migrations on all variants because they are still required when `migrations_ran>0` so we need to make sure they work whenever we add new migrations.

PR #2004, opened 2026-07-06
feat: use :syn process groups for inter-node broadcast fan-out

Olyno:
## Summary

Replace the current inter-node broadcast fan-out path with a `:syn`-based broker.

The goal is to reduce the cost of broadcasting the same tenant message across the cluster by publishing once to a `:syn` process group, then dispatching locally on  each node.

## Changes

- Add `Realtime.Broker.Syn`
- Route `TenantBroadcaster` broadcasts through the broker
- Batch messages per topic before local dispatch
- Pre-encode V2 broadcast payloads once before fan-out
- Add targeted broker/dispatcher tests
- Add fan-out benchmarks

## Benchmarks

Local benchmark results show the new `:syn` path improves fan-out throughput:

- `20 nodes / 500 messages`: `24.48ms` → `10.29ms` (`2.38x` faster)
- Fire-and-forget multi-node throughput: `12.43K ips` → `129.09K ips` (`10.4x` faster)
- Round-trip multi-node latency: `80.44µs` → `53.92µs` (`1.49x` faster)

These are local loopback benchmarks, so they should be read as relative comparisons rather than production numbers.

### DIFF
--- a/bench/fanout_throughput.exs
+++ b/bench/fanout_throughput.exs
@@ -1,0 +1,225 @@
+# Run with:
+#   MIX_ENV=test mix run --no-start bench/fanout_throughput.exs
+#
+# Local loopback scenario benchmark comparing per-node RPC fan-out with a
+# single :syn process-group publish. This intentionally uses a custom timer
+# instead of Benchee because each scenario needs distributed setup, delivery
+# verification, and peer cleanup.
+
+Application.ensure_all_started(:syn)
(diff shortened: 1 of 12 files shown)


edgurgel:
Hi thanks for the contribution but we use GenRpc for inter-node broadcast (now with Muster). gen_rpc already compresses and batches messages.

Issue #1953, opened 2026-06-12
Presence track() write-authorization probe also probes 'broadcast', logging noisy-but-benign RLS errors on Broadcast-from-Database setups


JuanCamiloDorado:
### Bug description (operational noise — functionality is correct)

On a **private channel**, the first presence `track()` of each subscription runs the
write-authorization probe, and the probe test-INSERTs one row into `realtime.messages`
**per extension** — `broadcast` and `presence` — not just the extension being exercised:

- [`presence_handler.ex#L85-L92`](https://github.com/supabase/realtime/blob/dae4aa31fd4facfdda24e639712aa3e8e5d9387d/lib/realtime_web/channels/realtime_channel/presence_handler.ex#L85-L92) — the first `track()` with `presence.write == nil` calls `get_write_authorizations/3` with no scoping opts
- [`authorization.ex#L296-L302`](https://github.com/supabase/realtime/blob/dae4aa31fd4facfdda24e639712aa3e8e5d9387d/lib/realtime/tenants/authorization.ex#L296-L302) — `extensions_to_check/1` returns `[:broadcast, :presence]` unless `presence_enabled?: false`
- [`authorization.ex#L323-L342`](https://github.com/supabase/realtime/blob/dae4aa31fd4facfdda24e639712aa3e8e5d9387d/lib/realtime/tenants/authorization.ex#L323-L342) — `check_write_policies/4` inserts one probe row per extension, each in its own savepoint; a denied insert (`42501`) is caught and correctly mapped to `write: false`

The savepoint handling is silent on the Realtime side, but **Postgres has already emitted an
`ERROR`-level log line** before the savepoint rolls back:

```
ERROR: new row violates row-level security policy for table "messages"
```

For projects on the documented [Broadcast from Database](https://supabase.com/docs/guides/realtime/broadcast#broadcast-from-the-database)
pattern, asymmetric write policies are the *intended* setup: clients write `presence`
(who's-online/cursors) but must **not** be able to write `broadcast` — server events are
emitted by a trigger via `realtime.send()` (which bypasses RLS), and an INSERT policy
permissive enough to let the probe through would let any authenticated client forge
server-emitted events. The result is a guaranteed, recurring RLS `ERROR` in the project's
Postgres logs on **every** first `track()` per subscription (editor open, reconnect, token
refresh), polluting the logs and masking real errors.

### To reproduce

RLS on `realtime.messages` for a private topic, allowing INSERT only for `presence`:

```sql
[...]

PR #2022, opened 2026-07-14
feat: broadcast persistence

leandrocp:

Introduce Broadcast Persistence allowing users to opt-in to `:persistence` to store broadcasted messages that can be replayed or manipulated in `realtime.messages`.

Closes REAL-956

### Notes
- Always try to broadcast first
- Ack `true` await persistence and return message id on response
- Ack `false` do not wait persistence, just fire and forget
- Feature flag `broadcast_persistence`

Ref REAL-956

### 1. Enable feature flag

```elixir
{:ok, flag} = Api.upsert_feature_flag(%{name: "broadcast_persistence", enabled: true})
FeatureFlags.Cache.update_cache(flag)
```

### 2. Create policies

```sql
-- policies
create policy demo_read on realtime.messages for select to authenticated
  using (realtime.topic() like 'persisted:%');

create policy demo_send on realtime.messages for insert to authenticated
  with check (realtime.messages.extension = 'broadcast' and realtime.topic() like 'persisted:%');

create policy demo_persist on realtime.messages for insert to authenticated
  with check (realtime.messages.extension = 'persistence' and realtime.topic() like 'persisted:%');
```

### 3. Set token

```shell
claims = %{"role" => "authenticated", "sub" => "demo", "exp" => System.system_time(:second) + 100000000000}
token = Joken.generate_and_sign!(%{}, claims, Joken.Signer.create("HS256", "dev"))
```

### 4. Broadcast from WS

```ws
# broadcast
{"topic":"realtime:persisted:demo","event":"phx_join","payload":{"config":{"private":true,"broadcast":{"self":true,"ack":true}}},"ref":"1"}
{"topic":"realtime:persisted:demo","event":"broadcast","payload":{"type":"broadcast","event":"hello","payload":{"msg":1}},"ref":"2"}
```

### 5. Broadcast from API

```shell
TOKEN="..."

curl -i -X POST "http://realtime-dev.localhost:4000/api/broadcast/persisted:demo/events/hello?private=true" \
  -H "authorization: Bearer $TOKEN" \
  -H "content-type: application/json" \
  -d '{"msg":2}'
```

### 6. Broadcast from DB

```sql
set time zone 'UTC';
select realtime.send('{"msg":3}'::jsonb, 'hello', 'persisted:demo', true);
```

### 7. Replay all messages

```ws
# replay
{"topic":"realtime:persisted:demo","event":"phx_join","payload":{"config":{"private":true,"broadcast":{"replay":{"since":0,"limit":100}}}},"ref":"1"}
```


lib/realtime/messages.ex
```diff
@@ -4,12 +4,49 @@ defmodule Realtime.Messages do
...
+  Automatically uses RPC if the database connection is not on the same node.
+  """
+  @spec persist(DBConnection.conn(), String.t(), String.t(), term()) ::
+          {:ok, binary()} | {:error, any()} | {:error, :rpc_error, term}
+  def persist(conn, topic, event, payload) when node(conn) == node() do
+    insert(conn, topic, event, payload)
+  end
+
+  def persist(conn, topic, event, payload) do
+    Realtime.GenRpc.call(node(conn), __MODULE__, :persist, [conn, topic, event, payload], key: topic)
```
edgurgel: We should also pass `tenant_id:` so that we get the `ErrorOnRpcCall` properly tagged

lib/realtime/messages.ex
```diff
@@ -53,7 +90,7 @@ defmodule Realtime.Messages do
         where:
           m.topic == ^topic and
             m.private == true and
-            m.extension == :broadcast and
+            m.extension in [:broadcast, :persistence] and
```
edgurgel: I was wondering if `:persistence` should be used just to validate the policy but the actual message would be saved with `:broadcast` 🤔

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -14,54 +15,66 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
 
   @event_type "broadcast"
-  @spec handle(payload, Socket.t()) :: {:reply, :ok, Socket.t()} | {:noreply, Socket.t()}
-  def handle(payload, %{assigns: %{private?: false}} = socket), do: handle(payload, nil, socket)
 
-  @spec handle(payload, pid() | nil, Socket.t()) :: {:reply, :ok, Socket.t()} | {:noreply, Socket.t()}
+  @doc """
+  Handles an outgoing broadcast for a channel.
+
+  `db_conn` is `nil` for public channels which don't write authorization nor persist messages.  is the tenant database connection, used to run write authorization and to persist the
```
edgurgel: Missing `It ` ?

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -14,54 +15,66 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
 
-    case run_authorization_check(policies || %Policies{}, db_conn, authorization_context) do
-      {:ok, %Policies{broadcast: %BroadcastPolicies{write: true}} = policies} ->
-        socket =
-          socket
-          |> assign(:policies, policies)
-          |> increment_rate_counter()
+    with {:ok, %Policies{broadcast: %BroadcastPolicies{write: true}} = policies} <-
+           run_authorization_check(policies || %Policies{}, db_conn, authorization_context),
+         socket = socket |> assign(:policies, policies) |> increment_rate_counter(),
```
edgurgel: I have to say that I'm not a fan of this pipe inside this with clause 😢

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -99,21 +112,15 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
+        send_message(tenant_id, self_broadcast, tenant_topic, payload)
+        if ack_broadcast, do: {:reply, :ok, socket}, else: {:noreply, socket}
 
-      true ->
-        {:noreply, socket}
+      {:error, :payload_size_exceeded} ->
+        if ack_broadcast,
+          do: {:reply, {:error, :payload_size_exceeded}, socket},
+          else: {:noreply, socket}
     end
```
edgurgel: did this change result in any behaviour change or just refactoring?

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -158,6 +165,26 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
+
+  defp maybe_persist(_policies, _db_conn, _topic, _payload), do: :skip
+
+  defp convert_to_persistable_fields(%{"event" => event, "payload" => payload}), do: {:ok, event, payload}
+
+  defp convert_to_persistable_fields({event, :json, user_payload, _metadata}) do
+    case Jason.decode(user_payload) do
+      {:ok, payload} -> {:ok, event, payload}
+      {:error, _} -> :error
+    end
```
edgurgel: Do we **need** to `Jason.decode/1` here? Cause ideally we would avoid decoding something that is already encoded and simply send it to Postgres to store as json and leave it up to Postgres do deal with it. Would this be possible?

lib/realtime/tenants/repo.ex
```diff
@@ -106,14 +106,6 @@ defmodule Realtime.Tenants.Repo do
     end
   end
 
-  defp result_to_single_struct(
```
leandrocp, line 109: Dead code.

lib/realtime/messages.ex
leandrocp: Done!

lib/realtime/messages.ex
leandrocp: That's a good point, I changed to check the persistence policy but store as broadcast.

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
leandrocp: Fair 😄

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
leandrocp: Reverted changes to have less diff possible.

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
leandrocp: Yep changed to `Jason.Fragment`

lib/realtime/api/message.ex
```diff
@@ -12,10 +12,11 @@ defmodule Realtime.Api.Message do
   @timestamps_opts [type: :naive_datetime_usec]
   schema "messages" do
     field(:topic, :string)
-    field(:extension, Ecto.Enum, values: [:broadcast, :presence])
+    field(:extension, Ecto.Enum, values: [:broadcast, :presence, :persistence])
     field(:payload, :map)
     field(:event, :string)
     field(:private, :boolean)
+    field(:skip_broadcast, :boolean)
```
leandrocp, line 20: I'm not super happy with this name but its intention is to flag if a message should be broadcasted or skipped (a filter). In a broadcast from WS/API it flags `skip_broadcast=true` so it doesn't broadcast again when the row is streamed.
Using https://www.postgresql.org/docs/current/logical-replication-row-filter.html could potentially simplify the whole thing but it's PG15+ only.

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -44,14 +53,22 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
 
         res =
           case Tenants.validate_payload_size(tenant_id, payload) do
-            :ok -> send_message(tenant_id, self_broadcast, tenant_topic, payload)
-            {:error, error} -> {:error, error}
+            # Broadcast first to prioritize throughput.
```
leandrocp, line 56: I believe sending first makes more sense because there's no delivery guarantee, so assuming there's no persistence guarantee either. In that case prioritize sending first and persist if possible.

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -44,14 +53,22 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
+            {:error, error} ->
+              {:error, error}
           end
 
         cond do
           ack_broadcast && match?({:error, :payload_size_exceeded}, res) ->
             {:reply, {:error, :payload_size_exceeded}, socket}
 
+          ack_broadcast && match?({:ok, _}, res) ->
+            {:reply, res, socket}
```
leandrocp, line 70: Returns message id when persisted, for eg:

```json
{
  "ref": "2",
  "event": "phx_reply",
  "topic": "realtime:persisted:demo",
  "payload": {
    "status": "ok",
    "response": {
      "id": "f00152b4-b3ad-4f2f-9941-e03c62515822"
    }
  }
}
```

lib/realtime/api/message.ex
leandrocp: Adopted row filter for PG15+

lib/realtime/tenants/authorization/policies.ex
```diff
@@ -2,20 +2,24 @@ defmodule Realtime.Tenants.Authorization.Policies do
...
   """
 
   alias Realtime.Tenants.Authorization.Policies.BroadcastPolicies
   alias Realtime.Tenants.Authorization.Policies.PresencePolicies
+  alias Realtime.Tenants.Authorization.Policies.PersistencePolicies
 
   defstruct broadcast: %BroadcastPolicies{},
-            presence: %PresencePolicies{}
+            presence: %PresencePolicies{},
+            persistence: %PersistencePolicies{}
```
edgurgel, line 15: What if we instead had `:persist` inside the BroadcastPolicies instead? Given that PersistencePolicies here is only applied to Broadcast?

lib/realtime/tenants/authorization.ex
```diff
@@ -329,7 +330,12 @@ defmodule Realtime.Tenants.Authorization do
   end
 
   defp check_write_policies(conn, authorization_context, extensions, policies) do
-    Enum.reduce_while(@all_extensions, {:ok, policies}, fn extension, {:ok, acc} ->
+    extensions =
+      if FeatureFlags.broadcast_persistence_enabled?(authorization_context.tenant_id),
+        do: extensions ++ [:persistence],
+        else: extensions
+
```
edgurgel, line 334: If long term we always check for persistence we might hit the same issue we are seeing here:

https://github.com/supabase/realtime/issues/1953

Logging pretty much on every private channel out there an error even though the join didn't request for persistence.

If we did based on the broadcast event (as metadata of the broadcast) then we could check lazily when the first persisted broadcast came through

lib/realtime/tenants/batch_broadcast.ex
```diff
@@ -150,6 +165,23 @@ defmodule Realtime.Tenants.BatchBroadcast do
     )
   end
 
+  defp maybe_persist(%PersistencePolicies{write: true}, db_conn, tenant, events) do
+    if FeatureFlags.broadcast_persistence_enabled?(tenant.external_id) do
+      Task.Supervisor.start_child(Realtime.TaskSupervisor, fn ->
```
edgurgel: Not for now but long term we might want to use a `PartitionSupervisor` dedicated for this?

lib/realtime/tenants/replication_connection.ex
```diff
@@ -224,20 +231,25 @@ defmodule Realtime.Tenants.ReplicationConnection do
...
+    else
+      check_publication_exists(state)
+    end
+  end
 
-    {:query, query, [timeout: state.query_timeout], %{state | step: :create_publication}}
+  def handle_result([%Postgrex.Result{rows: [[pg_version]]}], %__MODULE__{step: :check_pg_version} = state) do
+    state = %{state | pg_version: String.to_integer(pg_version)}
+    check_publication_exists(state)
   end
```
edgurgel, line 245: Given that this is a generic information we gather from the server should se just do it regardless of the feature flag?

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -201,10 +205,26 @@ defmodule Realtime.Tenants.SingleBroadcast do
...
+    :ok
+  end
+
+  defp maybe_persist(_persistence, _db_conn, _tenant, _topic, _event, _payload), do: :ok
+
   defp permissions_for_message(tenant, auth_params, topic) do
-    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id) do
-      auth_params = %{auth_params | topic: topic}
-      Authorization.get_write_authorizations(db_conn, auth_params)
+    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id),
```
edgurgel: should we just move this Connect clause out so you don't need to return db_conn here and keep this function just checking for the write authorizations?

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -58,7 +61,7 @@ defmodule Realtime.Tenants.SingleBroadcast do
           String.t(),
           String.t(),
           boolean(),
-          any(),
+          map() | binary(),
           content_type()
         ) :: :ok | {:error, term()} | {:error, atom(), String.t()}
   def broadcast(_auth_params, %Tenant{suspend: true}, _topic, _event, _private, _payload, _content_type) do
```
edgurgel: So if I understand well this means that every API request will try to persist?

I think this should be an option you pass from the API request. And if that's the case let's do this just on the new API which uses SingleBroadcast and leave BatchBroadcast alone?

lib/realtime/messages.ex
```diff
@@ -4,12 +4,65 @@ defmodule Realtime.Messages do
...
   alias Realtime.Api.Message
+  alias Realtime.Tenants.Repo
 
   import Ecto.Query, only: [from: 2]
 
   @hard_limit 25
   @default_timeout 5_000
 
+  @doc """
+  Persists a broadcast sent over WebSocket for `topic`.
```
edgurgel: This is also used by the API right?

lib/realtime/tenants/authorization/policies.ex
leandrocp: I tried to reuse the `:write/:read` pattern and didn't realize we could just add a new field, but I changed to `:persist` now and it reads much better!

lib/realtime/tenants/batch_broadcast.ex
leandrocp: That's the code that will need to change in a following PR to implement hard limits and/or buffering so I just used a simple task for now to get it done.

lib/realtime/tenants/replication_connection.ex
leandrocp: No preference really, It's generic and cheap but I used a feature flag so there's no modification in the current workflow. I'll remove it once the feature is done and the flag has to be removed anyway.

lib/realtime/tenants/authorization.ex
leandrocp: Since we already have #1953 it seems better to tackle that one first. Wdyt?

lib/realtime/tenants/single_broadcast.ex
leandrocp: Good point!

> I think this should be an option you pass from the API request

Added a query param `persist` (similar to `private`) to enable persistence.

lib/realtime/tenants/single_broadcast.ex
leandrocp: I'm not sure if you meant out of the `with` clauses or out of `handle_private_message` but on `main` `handle_private_message` already calls `Connect.lookup_or_start_connection/1` indirectly and now it just moved to to be more explicit and reused by `permissions_for_message` and `maybe_persist` so I'm not sure there's something else we can change here. 🤔

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -201,11 +229,24 @@ defmodule Realtime.Tenants.SingleBroadcast do
     end
   end
 
-  defp permissions_for_message(tenant, auth_params, topic) do
-    with {:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id) do
-      auth_params = %{auth_params | topic: topic}
-      Authorization.get_write_authorizations(db_conn, auth_params)
+  defp maybe_persist(%BroadcastPolicies{persist: true}, db_conn, tenant, topic, event, payload) do
+    if FeatureFlags.broadcast_persistence_enabled?(tenant.external_id) do
+      Task.Supervisor.start_child(Realtime.TaskSupervisor, fn ->
```
leandrocp: Same behavior as `ack=false`.

It might useful to add a `ack` query param but not on this PR.

lib/realtime_web/controllers/broadcast_single_controller.ex
```diff
@@ -52,6 +52,16 @@ defmodule RealtimeWeb.BroadcastSingleController do
         example: false,
         description: "Whether this is a private broadcast (requires RLS authorization). Defaults to false."
       ]
+      # TODO: uncomment when broadcast persistence is public and ready
```
leandrocp, line 55: To not expose it yet.

lib/realtime/tenants/authorization.ex
edgurgel: Yup! Work has started here: https://github.com/supabase/realtime/pull/2089

lib/realtime/tenants/batch_broadcast.ex
edgurgel: Makes sense!

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -152,11 +155,12 @@ defmodule Realtime.Tenants.SingleBroadcast do
 
   defp handle_private_message(tenant, auth_params, topic, event, payload, content_type, rate_counter) do
     case permissions_for_message(tenant, auth_params, topic) do
```
edgurgel: Calling `{:ok, db_conn} <- Connect.lookup_or_start_connection(tenant.external_id)` here, and then you don't need to get db_conn from "permissions_for_message" now that the connection is being used for two different things it makes less sense to have Connect being called from `permissions_for_message`

lib/realtime/tenants/single_broadcast.ex
edgurgel: I've added a comment above to be more clear

lib/realtime/tenants/single_broadcast.ex
edgurgel: Hmm yeah we might want to consider something like `persist=ack` or something like this? As ack here is only related to the case when something is persisted?

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
```diff
@@ -158,6 +176,55 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandler do
...
+        :ok
+    end
+  end
+
+  @spec convert_to_persistable_fields(payload) ::
+          {:ok, String.t(), map() | binary()} | {:error, :unsupported_payload}
+  defp convert_to_persistable_fields(%{"event" => event, "payload" => payload}), do: {:ok, event, payload}
+
+  defp convert_to_persistable_fields({event, :json, user_payload, _metadata}),
+    do: {:ok, event, Jason.Fragment.new(user_payload)}
```
edgurgel, line 223: Oh so Jason.Fragment will be properly stored as JSON without decoding/re-encoding by Ecto?

test/integration/rt_channel/broadcast_test.exs
```diff
@@ -256,6 +258,176 @@ defmodule Realtime.Integration.RtChannel.BroadcastTest do
...
 
+  describe "broadcast persistence" do
+    setup [:rls_context]
+
+    setup do
+      enable_broadcast_persistence_flag!()
+      :ok
+    end
+
+    test "public broadcast is delivered but not stored", %{
```
edgurgel, line 269: Should we make one of these tests send a binary payload?

test/realtime/tenants/batch_broadcast_test.exs
```diff
@@ -211,8 +216,8 @@ defmodule Realtime.Tenants.BatchBroadcastTest do
       Authorization
       |> expect(:build_authorization_params, 2, fn params -> params end)
       |> expect(:get_write_authorizations, 2, fn
-        _, %{topic: ^topic} -> %Policies{broadcast: %BroadcastPolicies{write: true}}
-        _, _ -> %Policies{broadcast: %BroadcastPolicies{write: false}}
+        _, %{topic: ^topic} -> {:ok, %Policies{broadcast: %BroadcastPolicies{write: true}}}
```
edgurgel: huh how was this test not breaking?

test/realtime/tenants/batch_broadcast_test.exs
```diff
@@ -490,6 +495,54 @@ defmodule Realtime.Tenants.BatchBroadcastTest do
     end
   end
 
+  describe "message persistence" do
```
edgurgel, line 500: There are no more changes on batch_broadcast 🤔 How are these tests passing?

test/realtime/tenants/single_broadcast_test.exs
```diff
@@ -392,6 +397,156 @@ defmodule Realtime.Tenants.SingleBroadcastTest do
...
+      expect(Connect, :lookup_or_start_connection, fn _ -> {:ok, db_conn} end)
+      expect(TenantBroadcaster, :pubsub_broadcast, fn _, _, _, _, _ -> :ok end)
+
+      expect(Authorization, :get_write_authorizations, fn _, _ ->
+        {:ok, %Policies{broadcast: %BroadcastPolicies{write: true}}}
+      end)
+
+      assert :ok = SingleBroadcast.broadcast(auth_params, tenant, topic, "event", true, %{"a" => "b"}, :json, true)
+
+      assert {:ok, []} = Repo.all(db_conn, messages_for(topic), Message)
```
edgurgel: This test could be passing if the task supervisor has not inserted right?  That's why eventually is used above, correct?

I wonder if we should check that there are no `children` under the task supervisor? 🤔 

Or us a refute eventually with retries like down below?

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -51,26 +54,40 @@ defmodule Realtime.Tenants.SingleBroadcast do
...
           String.t(),
           boolean(),
-          any(),
-          content_type()
+          map() | binary(),
+          content_type(),
+          boolean()
         ) :: :ok | {:error, term()} | {:error, atom(), String.t()}
-  def broadcast(_auth_params, %Tenant{suspend: true}, _topic, _event, _private, _payload, _content_type) do
+  def broadcast(auth_params, tenant, topic, event, private, payload, content_type, persist \\ false)
```
edgurgel: I wonder if we should use `opts` here for private and persist? It gets harder to read boolean arguments like this. WDYT?

lib/realtime/tenants/single_broadcast.ex
```diff
@@ -81,19 +98,28 @@ defmodule Realtime.Tenants.SingleBroadcast do
...
     |> validate_required([:topic, :event, :content_type])
     |> validate_payload_present(content_type, payload)
     |> validate_inclusion(:content_type, ["json", "binary"])
     |> validate_payload_size(tenant, content_type)
+    |> validate_persist_is_private()
+  end
+
+  defp validate_persist_is_private(changeset) do
+    case {get_field(changeset, :persist), get_field(changeset, :private)} do
+      {true, false} -> add_error(changeset, :persist, "can only be used on private broadcasts")
```
edgurgel: private channels*

test/realtime_web/channels/realtime_channel/broadcast_handler_test.exs
```diff
@@ -87,6 +91,8 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandlerTest do
     @tag policies: [:authenticated_read_matching_user_sub, :authenticated_write_matching_user_sub], sub: UUID.generate()
     test "with valid sub, is able to send message",
          %{topic: topic, tenant: tenant, db_conn: db_conn, sub: sub, serializer: serializer} do
+      stub(FeatureFlags, :broadcast_persistence_enabled?, fn _tenant_id -> true end)
```
edgurgel: hmm why the flag is true here when the describe block of persistence is down below? 🤔

test/realtime_web/channels/realtime_channel/broadcast_handler_test.exs
```diff
@@ -457,6 +465,246 @@ defmodule RealtimeWeb.RealtimeChannel.BroadcastHandlerTest do
...
+      v2_payload = {"event123", :json, "not json at all", %{}}
+
+      log =
+        capture_log(fn ->
+          assert {:reply, :ok, _socket} = BroadcastHandler.handle(v2_payload, db_conn, socket)
+        end)
+
+      assert_receive {:socket_push, _encoding, _data}
+      assert log =~ "UnableToPersistMessage"
+      assert {:ok, []} = Repo.all(db_conn, messages_for(topic), Message)
```
edgurgel, line 590: given the task stuff is it enough to look once for messages?

test/realtime_web/controllers/broadcast_single_controller_test.exs
```diff
@@ -660,6 +668,127 @@ defmodule RealtimeWeb.BroadcastSingleControllerTest do
...
+
+      {:ok, conn: conn, db_conn: db_conn}
+    end
+
+    test "persist=true stores the message and does not broadcast it again from the database", %{
+      conn: conn,
+      db_conn: db_conn,
+      tenant: tenant
+    } do
+      stub(GenCounter, :add, fn _ -> :ok end)
```
edgurgel, line 700: Why the stub on gencounter? 🤔

test/realtime/tenants/batch_broadcast_test.exs
leandrocp: Because the only test now is `test "the batch API never stores messages, even when authorized to persist"` to make it explicit that batch broadcast doesn't persist messages.

lib/realtime/tenants/single_broadcast.ex
leandrocp: Yep I prefer `opts` as well, changed it. 👍🏻

lib/realtime_web/channels/realtime_channel/broadcast_handler.ex
leandrocp: Yes although it's a bit tricky to test that exact path, but still tested indirectly by all other tests.

test/integration/rt_channel/broadcast_test.exs
leandrocp: Good point, added it!

test/realtime/tenants/batch_broadcast_test.exs
leandrocp: Because the code is:

```elixir
      case Authorization.get_write_authorizations(db_conn, auth_params) do
        {:ok, policies} -> policies
        {:error, :not_found} -> nil
        error -> error
      end
```

So it just falls into the `error` clause returning the struct which is the exact same return when it pass `{:ok, policies}` (the first clause).

Since `Authorization.get_write_authorizations/2` returns only `{:ok, ...} | {:error, ...}` (per spec) we should probably remove `error -> error`. Wdyt?

test/realtime/tenants/single_broadcast_test.exs
leandrocp: For now I'm avoiding adding too much code around tasks because that will be most likely temporary, we need something better to handle hard limits and pressure which is to be done in a following PR so I think it's okay for this version.

test/realtime_web/channels/realtime_channel/broadcast_handler_test.exs
leandrocp: `socket_fixture` in this module does `ack_broadcast = true` by default so the broadcast handler wait for that insert, but I also added `ack_broadcast: true` in the test to make it explicit.

test/realtime_web/controllers/broadcast_single_controller_test.exs
leandrocp: Because it shares the same setup and tenant with others tests that assert `rate_counter.avg` causing flaky tests.

lib/realtime/tenants/single_broadcast.ex
leandrocp: What if we add the `ack` param (defaults to false) into the API as well?

> As ack here is only related to the case when something is persisted?

Currently there's no ack right? So the only possible logic is to try to persist in a best-effort case, but having `ack` we could wait and respond with message id. Not sure is there's much value tho, need to discuss.

test/realtime_web/channels/realtime_channel/broadcast_handler_test.exs
leandrocp: Just a leftover, removed it.

lib/realtime/tenants/single_broadcast.ex
edgurgel: It really depends because we are mixing things here I think?

API is always ACK (single broadcast at least) as it will return error if broadcasting failed for example if the payload size is above the limit. See the status codes the single broadcast API can return

On the websocket if you don't want `ack` it will not reply if it exceeded a payload size.

Now persisting is sort of another layer on top of this not the only thing

WDYT?

test/realtime/tenants/batch_broadcast_test.exs
edgurgel: ooooh sorry I should've read it...

test/realtime_web/controllers/broadcast_single_controller_test.exs
edgurgel: I see... We might want fix these flaky tests then. If you see them again let's see if we can fix it!

lib/realtime/tenants/authorization.ex
edgurgel: Ok this has been merged!

edgurgel (approved):
I've added some comments but other than those LGTM

PR #2026, opened 2026-07-15
fix: log kid on bad signer error

filipecabaco:

To improve the capacity for a user to debug the issues they have with badly created tokens we will log in the error message the kid used for a given token


lib/realtime_web/channels/auth/jwt_verification.ex
edgurgel: Should we update the typespec here ?

filipecabaco: yes we should ... forgot those 🤦

filipecabaco: done 👍

edgurgel (approved):
LGTM

PR #2027, opened 2026-07-16
feat: use Muster.join under a feature flag

edgurgel:

Feature flagged Muster.join. Using a task due to the potential RPC that might take some time


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -690,6 +696,53 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+  # fanned out through Muster would miss this pid) - except in the rare case
+  # the join is genuinely slow, where we cut our losses instead of blocking
+  # the reply on Muster's own much larger worst-case timeout (its internal
+  # :claim_call_timeout ceiling is 60s).
+  @muster_join_await_ms 4_000
+
+  defp await_muster_join(task, socket) do
+    case Task.yield(task, @muster_join_await_ms) || Task.shutdown(task, :brutal_kill) do
+      {:ok, :ok} -> :ok
+      {:ok, {:error, reason}} -> log_error(socket, "MusterJoinError", inspect(reason))
```
leandrocp, line 738: Add `MusterJoinError` to https://github.com/supabase/realtime/blob/main/ERROR_CODES.md ?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -690,6 +696,53 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+  # (DB connect, auth checks, ...) instead of adding pure tail latency.
+  # Plain Task.async/1 (no shared Task.Supervisor) avoids serializing a join
+  # storm through one process; the try/rescue/catch inside guarantees the
+  # task always returns a plain term instead of exiting abnormally, so its
+  # link to this process can never propagate a crash back to the channel.
+  defp maybe_start_muster_join(tenant_id, transport_pid) do
+    if FeatureFlags.enabled?("use_muster_channel_join", tenant_id) do
+      scope = Application.get_env(:realtime, :muster_scope)
+
+      unless Muster.local_member?(scope, tenant_id, transport_pid) do
```
leandrocp, line 710: `Muster.local_member?/3` can raise here:

```elixir
:unknown -> raise "Forum for scope #{inspect(scope)} is not started"
```

I'm wondering if we should make `Muster.local_member?/3` safer, ie: return _only_ true/false.

edgurgel: Yeah I might do that 👍

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Yup will do that 👍 . I was thinking on not adding as I was hoping for this to be temporary but I think it will be useful for a while.

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Well the only way that this will fail is if the Scope `NEVER` started. Which should not be possible. Even if the Scope and shards are restarting this will reply `true/false` 🤔 So I think it's safe as is.

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Fixed!

leandrocp (approved):

PR #2054, opened 2026-07-28
chore: update pg-delta to v1.0.0-alpha.33

leandrocp:
Just maintenance as usual.

Changes: https://github.com/supabase/pg-toolbelt/compare/%40supabase%2Fpg-delta%401.0.0-alpha.30...%40supabase%2Fpg-delta%401.0.0-alpha.33


.github/workflows/update-tenant-db-snapshots.yml
leandrocp, line 1: Renamed because this actions does generate both the pg-delta catalog and the pg dump files.

filipecabaco (approved):

PR #2057, opened 2026-07-29
feat: update Erlang OTP and Elixir

leandrocp:
Ref REAL-926


filipecabaco (approved):

Dockerfile
```diff
@@ -1,6 +1,6 @@
-ARG ELIXIR_VERSION=1.18
-ARG OTP_VERSION=27.3
-ARG DEBIAN_VERSION=bookworm-20250929-slim
+ARG ELIXIR_VERSION=1.19.5
+ARG OTP_VERSION=28.5.0.4
```
leandrocp, line 2: Using exact version (major.minor.patch) everywhere because it's safer, more predictable, and easier to grep/replace.

Dockerfile
```diff
@@ -1,6 +1,6 @@
...
+ARG DEBIAN_VERSION=trixie-20260713-slim
```
leandrocp, line 3: Updating from oldstable to stable because Debian `trixie` is well.. very stable even tho oldstable still gets security patches.

lib/realtime/application.ex
```diff
@@ -27,8 +27,10 @@ defmodule Realtime.Application do
   end
 
   def start(_type, _args) do
-    check_for_local_ipv6_host()
-    opentelemetry_setup()
+    if Application.get_env(:logflare_logger_backend, :url) do
+      LoggerBackends.add(LogflareLogger.HttpBackend)
```
edgurgel: Would you mind adding to the PR description why we need this change

leandrocp: Actually that's not necessary, I'm just used to use it and didn't realize that `LogflareLogger` already handle batching, etc so I'm removing it.

config/runtime.exs
```diff
@@ -301,8 +301,7 @@ if logs_engine == "logflare" do
 
   config :logger,
     sync_threshold: 6_000,
-    discard_threshold: 6_000,
-    backends: [LogflareLogger.HttpBackend]
```
edgurgel, line 305: and do we still need this removed?

leandrocp: Yes it was hard-deprecated in v1.19 https://github.com/elixir-lang/elixir/blob/098c01a7ee3d0f7d6516d649efaec338cccd24ea/lib/logger/lib/logger/app.ex#L14-L29

We don't need `:logger_backends` _yet_ because [Logger.add_backend/2](https://logger.hexdocs.pm/1.19.5/Logger.html#add_backend/2) was only soft-deprecated, no warnings until v1.20

edgurgel (approved):

lib/realtime/tenants/janitor.ex
```diff
@@ -71,8 +71,7 @@ defmodule Realtime.Tenants.Janitor do
         task =
           Task.Supervisor.async_nolink(
             __MODULE__.TaskSupervisor,
-            fn -> perform_maintenance_tasks(chunks) end,
-            ordered: false
```
leandrocp, line 75: Removed `:ordered` because that's not a valid opt.

PR #2058, opened 2026-07-30
chore(ci): check unused deps

leandrocp:
We should avoid locking unused deps. Found this when I was removing an installed dep.


edgurgel (approved):
LGTM

PR #2060, opened 2026-07-30
fix(auth): accept the RFC 8037 EdDSA alg for OKP keys

jclab-joseph:

Bug fix.

This PR fixes JWT verification for EdDSA tokens signed with Ed25519 or Ed448 keys.

JWT verification only recognizes the JOSE algorithm names that identify the curve directly, such as `Ed25519` and `Ed448`.

However, RFC 8037 specifies `alg` as `EdDSA`, with the specific curve provided by the JWK's `crv` field. Conformant libraries and services—including jose for JavaScript, go-jose, python-jose, and the Supabase Storage API—emit Ed25519 tokens using this format.

As a result, a valid Ed25519 token with `alg: "EdDSA"` fails verification with `:error_generating_signer`, even when the matching key is present in the tenant's JWKS.

## What is the new behavior?

JWT verification now recognizes the `EdDSA` algorithm and resolves the appropriate signer using the JWK's `crv` value.

For example:

- `alg: "EdDSA"` with `crv: "Ed25519"` resolves to the Ed25519 signer.
- `alg: "EdDSA"` with `crv: "Ed448"` resolves to the Ed448 signer.
- An OKP key using an unsupported curve continues to fail cleanly instead of selecting an invalid signer.

This allows standards-compliant EdDSA tokens produced by commonly used JOSE libraries and the Supabase Storage API to be verified successfully.

## Additional context

RFC 8037 separates the signing algorithm from the selected Edwards curve:

- The JWT header uses `alg: "EdDSA"`.
- The JWK identifies the curve through its `crv` field.

This change preserves support for the existing curve-specific spellings while adding support for the RFC-compliant EdDSA representation.


test/realtime_web/channels/auth/jwt_verification_test.exs
```diff
@@ -509,6 +509,45 @@ defmodule RealtimeWeb.JwtVerificationTest do
       assert {:ok, _claims} = JwtVerification.verify(token, @jwt_secret, jwks)
     end
 
+    test "using an Ed25519 JWK with the RFC 8037 EdDSA alg" do
```
edgurgel: could we add a test for the case with a wrong signature (but same algo) as well?

edgurgel:
@jclab-joseph it turns out that Joken added support to EdDSA: 

https://github.com/joken-elixir/joken/blob/main/CHANGELOG.md#260---2023-01-29

Can we update joken and avoid building the `Joken.Signer` manually?

jclab-joseph:
@edgurgel Yes, that's better. I've updated it.

edgurgel (approved):
LGTM! Thank you!

PR #2069, opened 2026-08-04
feat: backfill GCM encryption

filipecabaco:

Adds GCM encryption to credentials stored in Realtime and feature flags the migration.


lib/realtime/tenants.ex
```diff
@@ -480,9 +485,104 @@ defmodule Realtime.Tenants do
...
+      end)
+    end
+
+    :ok
+  end
+
+  defp needs_encryption_reconciliation?(tenant) do
+    (is_nil(tenant.jwt_secret_gcm) and not is_nil(tenant.jwt_secret)) or
+      Enum.any?(tenant.extensions, &(is_nil(&1.settings_gcm) and encrypted_settings_keys(&1.type) != []))
+  end
```
edgurgel, line 528: Should it also check if gcm_migrated_at is not null?

filipecabaco: what would be the advantage?

edgurgel: I'm just trying to understand how do we check the difference between a new tenant that might not have the old encrypted field versus one that had and migrated. 🤔

config/runtime.exs
```diff
@@ -19,6 +19,8 @@ dashboard_password = System.get_env("DASHBOARD_PASSWORD", :crypto.strong_rand_by
 dashboard_user = System.get_env("DASHBOARD_USER", :crypto.strong_rand_bytes(12) |> Base.encode16(case: :lower))
 db_after_connect_query = System.get_env("DB_AFTER_CONNECT_QUERY")
 db_enc_key = System.get_env("DB_ENC_KEY")
+db_enc_key_gcm = System.get_env("DB_ENC_KEY_GCM")
+db_enc_write_gcm = System.get_env("DB_ENC_WRITE_GCM") == "true"
```
leandrocp: You can use `Env` here because it does handle some common user errors and returns `boolean()`:

```suggestion
db_enc_write_gcm = Env.get_boolean("DB_ENC_WRITE_GCM")
```

lib/realtime/api/tenant.ex
```diff
@@ -34,6 +34,7 @@ defmodule Realtime.Api.Tenant do
     field(:client_presence_window_ms, :integer)
     field(:presence_enabled, :boolean, default: false)
     field(:feature_flags, :map, default: %{})
+    field(:gcm_migrated_at, :utc_datetime)
```
leandrocp, line 37: Is this field temporary and will be deleted along with `lib/realtime/tenants/encryption_reconciler.ex` when the migration is done?

I'm asking because it seems we could rely on the `g1:` prefix to check if a tenant has migrated unless you need to know exactly when it was migrated.

lib/realtime/tenants/encryption_reconciler.ex
```diff
@@ -0,0 +1,105 @@
+defmodule Realtime.Tenants.EncryptionReconciler do
+  @moduledoc """
+  Moves a tenant's stored secrets from the legacy AES-128-ECB cipher to AES-256-GCM.
+
+  Temporary: once every tenant is on GCM this module and its call site in `Realtime.Tenants` are
+  deleted wholesale, along with the legacy branch in `Realtime.Crypto`.
+  """
+
```
leandrocp, line 8: Nit but it would be nice to have a TODO marker here to help us track pending work, like:


```suggestion
# TODO: delete file when gcm migration is done.
```

leandrocp (approved):
Just non-blocking comments.

lib/realtime/api/tenant.ex
filipecabaco: this is an easier way for us to track it at the DB level IMO vs checking all strings so that way we can keep count but after the migration is done probably yes this will be a removed field

leandrocp (approved):

PR #2071, opened 2026-08-05
feat: presence enabled accepted as a valid to connect

filipecabaco:

presence enabled accepted as a valid to connect

### DIFF
--- a/lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
+++ b/lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
@@ -14,12 +14,15 @@ defmodule RealtimeWeb.RealtimeChannel.MessageDispatcher do
         log_level,
         tenant_id,
         replayed_message_ids \\ MapSet.new(),
-        presence_read? \\ true
+        presence_read? \\ true,
+        broadcast_read? \\ true
       ) do
-    {:rc_fastlane, fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids, presence_read?}
+    {:rc_fastlane, fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids, presence_read?,
+     broadcast_read?}
   end
 
   @presence_diff "presence_diff"
+  @broadcast "broadcast"
 
   @doc """
   This dispatch function caches encoded messages if fastlane is used
@@ -46,9 +49,10 @@ defmodule RealtimeWeb.RealtimeChannel.MessageDispatcher do
         {pid, _}, {cache, count} when pid == from ->
           {cache, count}
 
-        # Subscriber is authorized for broadcast.read but denied presence.read: withhold the
-        # presence_diff. Mirrors the can_read_presence?/1 gate on the presence_state push.
-        {_pid, {:rc_fastlane, _fastlane_pid, _serializer, _join_topic, _log_level, _tenant_id, _replayed, false}},
+        # Subscriber is denied presence.read: withhold the presence_diff. Mirrors the
+        # can_read_presence?/1 gate on the presence_state push.
+        {_pid,
+         {:rc_fastlane, _fastlane_pid, _serializer, _join_topic, _log_level, _tenant_id, _replayed, false, _bcast}},
         {cache, count} ->
           {cache, count}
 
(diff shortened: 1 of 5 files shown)


lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
```diff
@@ -14,12 +14,15 @@ defmodule RealtimeWeb.RealtimeChannel.MessageDispatcher do
         log_level,
         tenant_id,
         replayed_message_ids \\ MapSet.new(),
-        presence_read? \\ true
+        presence_read? \\ true,
+        broadcast_read? \\ true
       ) do
-    {:rc_fastlane, fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids, presence_read?}
+    {:rc_fastlane, fastlane_pid, serializer, topic, log_level, tenant_id, replayed_message_ids, presence_read?,
+     broadcast_read?}
```
PragTob, line 21: maybe the creation/read speed for tuples outweighs this here, but otherwise feels like a struct may be beneficial (would cause less code changes on an update, as we don't need to reiterate all fields all the time)

test/integration/rt_channel/presence_test.exs
```diff
@@ -127,6 +127,77 @@ defmodule Realtime.Integration.RtChannel.PresenceTest do
...
     end
 
+    @tag policies: [:authenticated_read_presence, :authenticated_write_presence]
+    test "private presence with presence only permissions can join, track and never sees broadcasts",
+         %{tenant: tenant, topic: topic, serializer: serializer} do
+      {socket, _} = get_connection(tenant, serializer, role: "authenticated")
+      {service_role_socket, _} = get_connection(tenant, serializer, role: "service_role")
+      topic = "realtime:#{topic}"
+
+      WebsocketClient.join(socket, topic, %{config: %{presence: %{key: "", enabled: true}, private: true}})
```
PragTob: I like that we test this on the websocket client level 👌  💚

test/integration/rt_channel/presence_test.exs
```diff
@@ -127,6 +127,77 @@ defmodule Realtime.Integration.RtChannel.PresenceTest do
...
+      assert_receive %Message{event: "presence_state", payload: %{}, topic: ^topic}, 500
+
+      WebsocketClient.join(service_role_socket, topic, %{config: %{private: true}})
+      assert_receive %Message{event: "phx_reply", payload: %{"status" => "ok"}, topic: ^topic}, 500
+
+      payload = %{name: "realtime_presence_96", t: 1814.7000000029802}
+      WebsocketClient.send_event(socket, topic, "presence", %{type: "presence", event: "TRACK", payload: payload})
+
+      assert_receive %Message{event: "presence_diff", payload: %{"joins" => joins, "leaves" => %{}}, topic: ^topic}, 500
+      join_payload = joins |> Map.values() |> hd() |> get_in(["metas"]) |> hd()
```
PragTob: I don't think this is stable, as I don't believe `Map.values()` has a defined return order so `hd()` might get different values. We should probably pick the value explicitly to avoid glaky tests

test/realtime_web/channels/realtime_channel/message_dispatcher_test.exs
```diff
@@ -109,6 +123,47 @@ defmodule RealtimeWeb.RealtimeChannel.MessageDispatcherTest do
...
+
+      for msg <- [
+            %Broadcast{topic: "some:other:topic", event: "broadcast", payload: %{data: "test"}},
+            %UserBroadcast{topic: "some:other:topic", user_event: "event", user_payload: %{data: "test"}}
+          ] do
+        assert MessageDispatcher.dispatch(subscribers, from_pid, msg) == :ok
+      end
+
+      assert Agent.get(TestSerializer, & &1) == 0
+      refute_receive _any
```
PragTob, line 157: we could consider switching it to `refute_received` or adjusting the timeout, depending on timeout this might wait 100ms which we may or may not want :)

PragTob (commented):
👋  Hi I have no idea what I'm doing. But I felt like reviewing to get a feel for the code. Hence I'm also not approving, as I don't feel like I know enough to approve yet.

All looks good to me, but this is the first time I'm taking a look at the code base 😅  

You know me, but still: Hi I'm Tobi, I tend to leave a lot of comments with the intent to learn from each other. It's perfectly normal for me to leave 5+ comments but approve the PR because it's minor stuff that shouldn't stop delivery or just genuine question, interest or praise.

Here, I think the `Map.values()` one deserves a look as I hate flaky tests. The others are questions/praise/highly optional.

Image

edgurgel (changes requested):
What happens if the user gains access to read broadcasts on a new refreshed token?

filipecabaco:
@edgurgel added the logic to be processed during a refresh token

edgurgel (approved):
LGTM

lib/realtime_web/channels/realtime_channel/message_dispatcher.ex
edgurgel: Yeah that's a good point. We had a small tuple in the beginning mostly due to speed and size.

PR #2073, opened 2026-08-06
feat: broadcast muster routing

edgurgel:

* New feature flag to broadcast using Muster's routers instead of always flooding:  `use_muster_broadcast` (which __must__ have the same feature flag bucket key as `use_muster_channel_join`). Only for `broadcast` messages. Postgres Changes and Presence are not touched here.
* For intra region comms we run the Muster scope so we can naturally can find the router and use it to forward to the other nodes.
* For inter region comms we keep a set of `ExHashRing.Ring` (managed by the new `RegionRings` process) of what we expect the membership to be to calculate the router. They might not agree briefly while a rebalance is happening but if a rebalance is happening we would flood anyway and that's what happens. If we send to the wrong node it will just flood the region (and log a warning so we keep track of how often this happens)


lib/realtime/gen_rpc/pub_sub.ex
```diff
@@ -45,23 +48,87 @@ defmodule Realtime.GenRpcPubSub do
...
+    worker = worker_name(adapter_name, self())
+    my_region = Application.get_env(:realtime, :region)
+    scope = Application.get_env(:realtime, :muster_scope)
+
+    # Cross-region: route to each remote region's expected router, or flood that
+    # region when we have no ring for it yet.
+    cross_region_route(worker, my_region, tenant_id, topic, message, dispatcher)
+
+    # Intra-region: route to the node Muster says owns this tenant's occupancy.
+    case Muster.router(scope, tenant_id) do
```
leandrocp, line 98: `Muster.router/2` can return `{:error, atom}` from `ExHashRing.Ring.find_node/2` but that spec is missing and consequently also missing the case clause here. Shouldn't it fallback to `intra_region_flood` in this case or would this never happen?

lib/realtime/gen_rpc/pub_sub.ex
```diff
@@ -85,47 +177,156 @@ end
...
+    nodes =
+      case Forum.Muster.router(scope, tenant_id) do
+        # Still the router: use the authoritative occupancy set, or over-deliver
+        # to the whole region on a readiness/view-hash mismatch.
+        {:ok, router_node} when router_node == node() ->
+          targets_or_flood(scope, tenant_id, my_region, view_hash)
+
+        # Router changed somehow, we have a bug or it's rebalancing at the moment
+        _ ->
+          Logger.warning("Muster router changed during broadcast for tenant #{tenant_id}, falling back to region flood")
```
leandrocp, line 265: Do we want to differentiate `:route` from `:route_region` to know which one is failing? The warn message in `:route_region` is equal.

lib/realtime/gen_rpc/pub_sub/region_rings.ex
```diff
@@ -0,0 +1,225 @@
...
+  @spec expected_router(String.t(), Forum.group(), atom()) ::
+          {:ok, node(), non_neg_integer()} | :error
+  def expected_router(region, group, table \\ @table) when is_binary(region) do
+    with [{^region, ring_name, view_hash}] <- :ets.lookup(table, region),
+         {:ok, node} <- ExHashRing.Ring.find_node(ring_name, group) do
+      {:ok, node, view_hash}
+    else
+      _ -> :error
+    end
+  rescue
```
leandrocp, line 95: I believe it's cheaper to call `:ets.whereis/1` as first `with` statement.

lib/realtime/application.ex
leandrocp, line 99: Maybe unrelated to this PR but I'm wondering why this region naming fallbacks to `"local"` while all others just assume whatever value is defined in `Application.get_env(:realtime, :region)` including `nil` and if that could cause problems downstream in envs where that value might not be defined (self-hosted/cli maybe).

leandrocp (commented):
Just comments so far to understand how it works and clear some doubts.

lib/realtime/gen_rpc/pub_sub.ex
edgurgel: Great point

lib/realtime/application.ex
edgurgel: Good question! I went with "local" so there was no trailing `_` but will investigate! CLI won't matter as they don't become clusters so there is no intra/inter messaging anyway

But yeah self-hosted clusters might have issues if they are not consistently addressed

lib/realtime/gen_rpc/pub_sub.ex
edgurgel: Good point! Will review! Yeah I might just do `_ ->` instead of `{:rebalancing, _}` to go intra_region_flood

lib/realtime/gen_rpc/pub_sub/region_rings.ex
edgurgel: The trick is that there is a race condition where :ets.whereis can reply with a `tid` and when we do lookup it can still raise (because the table died after whereis replied) so this is just a generic catch.

I don't expect this to ever happen realistically but I went full :tinfoil_hat: there to ensure we don't blow up on the GenRpc PubSub flow.

lib/realtime/application.ex
edgurgel: Ok so after reviewing with Muster we actually route nicely when the routing works but "fails" on the fallback. And without Muster with `region=nil` today this happens:

> def region_nodes(nil), do: []

> def node_from_region(_, _), do: {:error, :not_available}

https://github.com/supabase/realtime/blob/2eda30a5c1fb339feb8f52f5b8abae04a181e157/lib/realtime/nodes.ex#L68C1-L96

Which means that this is already "broken" for people that don't set a region as no other nodes are broadcasted to.

So if we wanted to fix this we would probably have a default region name instead of having it `nil`. 

But I think this belongs to other PR as this one is big enough. WDYT?

lib/realtime/gen_rpc/pub_sub.ex
edgurgel: Addressed here https://github.com/supabase/realtime/pull/2073/changes/0ae80f5c683e1d3ee5ce825efc5aa1f799708320

lib/realtime/gen_rpc/pub_sub.ex
edgurgel: https://github.com/supabase/realtime/pull/2073/changes/0ae80f5c683e1d3ee5ce825efc5aa1f799708320

lib/realtime/gen_rpc/pub_sub/region_rings.ex
leandrocp: All right makes sense!

lib/realtime/application.ex
leandrocp: Another issue/PR for sure. It was just more clear now because of the Muster code dealing with regions but that's out of scope for this PR.

leandrocp (approved):
🚀

PR #2076, opened 2026-08-07
chore: refactor Containers into TestTenantDb

edgurgel:


filipecabaco (approved):

test/realtime/extensions/cdc_rls/subscriptions_test.exs
```diff
@@ -823,7 +823,7 @@ defmodule Realtime.Extensions.PostgresCdcRls.SubscriptionsTest do
     end
 
     test "create works for a table whose name contains a backslash", %{conn: conn} do
-      Postgrex.query!(conn, ~s|CREATE TABLE "my\\table" (id int)|, [])
+      Postgrex.query!(conn, ~s|CREATE TABLE IF NOT EXISTS "my\\table" (id int)|, [])
```
leandrocp, line 835: I think this table should be reseted/dropped at the end of the test or during setup, otherwise it can carry data and config over subsequent tests.

leandrocp (changes requested):
Just a small note, not blocking.

test/realtime/extensions/cdc_rls/subscriptions_test.exs
edgurgel: Yeah, but I was actually thinking on simply dropping all tables under public schema during our test tb setup. So even if we kill the test before running the on_exit it will clean up before things run

edgurgel: Sadly rebuilding `public` is not trivial so I will just drop this table for now.

leandrocp (approved):

PR #2077, opened 2026-08-07
feat: replay binary messages

leandrocp:
Ref REAL-960

Introduce support to replay binary messages.

Returns empty.

## What is the new behavior?

Returns the same payload as the original broadcast.

## Additional context

https://github.com/supabase/realtime/pull/1526 and https://github.com/supabase/realtime/pull/1869


filipecabaco (approved):

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -1095,6 +1095,31 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+
+    user_broadcast = %UserBroadcast{
+      topic: topic,
+      user_event: event,
+      user_payload: binary_payload,
+      user_payload_encoding: :binary,
+      metadata: meta
+    }
+
+    send(transport_pid, serializer.fastlane!(user_broadcast))
```
edgurgel: Hmm fastlane is not really supposed to be here really 🤔. Cant we push as the other message?

edgurgel: I think as long as the message looks like this it will encode fine https://github.com/supabase/realtime/blob/main/lib/realtime_web/socket/v2_serializer.ex#L110

leandrocp: Nice let me test.

edgurgel: If not we should find a way as fastlane is supposed to be used outside the channel

leandrocp: I'm looking into https://github.com/supabase/realtime/blob/744c6d60f000f721c3942b6df0e7601f54a3b69d/lib/realtime_web/socket/v2_serializer.ex#L110 but it doesn't encode `meta` so it would lose `replayed:true`

edgurgel: Oh yeah maybe we should fix this then 💡

TBH This is more of a conceptual problem rather than technical as for the original serializer "Broadcast" was only used outside the channel and inside the channel just "Push" and "Reply" were used.

So I guess it's fine to use fastlane here? WDYT?

leandrocp: Given https://github.com/supabase/supabase-js/blob/586dee38d946273c8c7e5eacd5b37136edef9c4f/packages/core/realtime-js/src/lib/serializer.ts#L152-L160 I think that's the smallest change otherwise it will require SDK changes first, but honestly no hard opinions I'm still learning the codebase 😄

edgurgel (approved):
LGTM! Thanks!

Do you think we should improve the e2e tests to include a binary payload in the replay tests https://github.com/supabase/realtime/blob/main/test/e2e/realtime-check.ts ?

leandrocp:
> LGTM! Thanks!
> 
> Do you think we should improve the e2e tests to include a binary payload in the replay tests https://github.com/supabase/realtime/blob/main/test/e2e/realtime-check.ts ?

Absolutely! Pushed https://github.com/supabase/realtime/pull/2077/changes/525c694cdd5dbe6c8b8ce8ed270b4ea609900bb0

edgurgel (approved):

PR #2079, opened 2026-08-07
fix: improve inspector usability

filipecabaco:

found multiple errors when trying to use the new inspector UI. doing another round of cleaning and improvements to it 
including better pg changes debugging, select, etc.

Also tackled a small docker bug where we setup the proper docker file cache usage

1-connected-filters-and-columns
2-editor-multiple-conditions
3-editor-rejects-invalid-is
4-editor-empty-state


edgurgel (changes requested):
Do you need to commit those priv/static files? They look like digests 🤔

assets/js/app.js
leandrocp, line 1: I think it's about time to start using a formatter and linter for JS given the amount of code we've been adding recently.

I know that https://oxc.rs/docs/guide/usage/formatter.html and https://oxc.rs/docs/guide/usage/linter works very well but no preference, just mentioning.

mix.exs
```diff
@@ -137,6 +137,7 @@ defmodule Realtime.MixProject do
   defp aliases do
     [
       setup: ["deps.get", "ecto.setup", "cmd npm install --prefix assets"],
+      "test.assets": ["cmd node --test assets/js/*.test.mjs"],
```
leandrocp: Shouldn't run these tests in CI? 🤔

leandrocp (approved):

edgurgel (approved):

PR #2080, opened 2026-08-10
fix: optimise read auth probe

edgurgel:

* Avoid using `RETURNING` when we don't care about all the rows except `id`
* `SELECT` only `id` to identify which extension was approved


PragTob (approved):
👌 

(took me a while to grog why we can't just return the ids instead when we could return all before but I think I got it 😅  Thanks for the excellent _why_ comments 👌 )

Image

leandrocp (approved):

PR #2082, opened 2026-08-11
fix: bump phoenix

edgurgel:

Update phoenix


PragTob (approved):
ooof we were at live view pre 1.0? 😬 

I'm kind of worried/surprised that there aren't more changes here as I do remember that update being more destructive.

Not sure how good our tests are for that part (or what even it is), might deserve some more manual testing (not sure how much was done).

Approving as the changes look good generally and of course this is the right direction 👌 

Image

leandrocp:
Diff: https://hex.pm/diffs?diffs[]=floki%3A0.38.0%3A0.38.4&diffs[]=phoenix%3A1.7.19%3A1.8.10&diffs[]=phoenix_live_view%3A0.20.17%3A1.2.9&diffs[]=plug_crypto%3A2.1.1%3A2.2.0

_Phoenix is approximate because it was running from `{:git, "https://github.com/supabase/phoenix.git", "7b884cc0cc1a49ad2bc272acda2e622b3e11c139", [branch: "feat/presence-custom-dispatcher-1.7.19"]}`_

mix.exs
```diff
@@ -124,7 +125,7 @@ defmodule Realtime.MixProject do
     if path = System.get_env("PHOENIX_PATH") do
       {:phoenix, path: path, override: true}
     else
-      {:phoenix, override: true, github: "supabase/phoenix", branch: "feat/presence-custom-dispatcher-1.7.19"}
+      {:phoenix, "~> 1.8"}
```
leandrocp, line 130: 1. Phoenix 1.8 introduced the option `:max_channels_per_transport` in `Phoenix.Socket` which i believe should be set by `TENANT_MAX_CHANNELS_PER_CLIENT` right? So we need to pass it into `RealtimeWeb.Socket` otherwise it will always [default to 100](https://github.com/phoenixframework/phoenix/blob/5506184f9180ef21cb99dbbcb08645329e0784a3/lib/phoenix/socket.ex#L656).

2. Missing `compilers: [:phoenix_live_view] ++ Mix.compilers()` in mix.exs to support colocated hooks/js/css

leandrocp (changes requested):

mix.exs
edgurgel: Actually for max_channels_per_transport we should have it above 100 because we have our own error message for this anyway. Good catch

> Missing compilers: [:phoenix_live_view] ++ Mix.compilers() in mix.exs to support colocated hooks/js/css

true!

PragTob: Thanks for catching all that @leandrocp - great work! 🚀

lib/realtime_web/channels/user_socket.ex
```diff
@@ -1,5 +1,9 @@
 defmodule RealtimeWeb.UserSocket do
-  use RealtimeWeb.Socket
+  # Raise Phoenix's built-in per-transport channel limit above its default of 100. Realtime
+  # enforces its own per-tenant `max_channels_per_client` limit (default 100) in RealtimeChannel
+  # with a dedicated error message, so we keep a bit of headroom above it and let that be the
+  # source of truth for the error the client sees.
+  use RealtimeWeb.Socket, max_channels_per_transport: 110
```
leandrocp: Wait now I'm a bit confused with this change 😄

Let's suppose `TENANT_MAX_CHANNELS_PER_CLIENT` is 50 and limit is reached then we return `ChannelRateLimitReached` ✅
But if we increase `TENANT_MAX_CHANNELS_PER_CLIENT` to 111 or more then we would actually return the Phoenix error because it exceeds 110, right? Which I believe is not what we want. In short I think we need to pass `TENANT_MAX_CHANNELS_PER_CLIENT` to `:max_channels_per_transport` and add some tests. WDYT?

edgurgel: `TENANT_MAX_CHANNELS_PER_CLIENT` is the default if not passed. But we use what's in the tenant row not the ENV.

So only if the tenant max_channels_per_clients is greater than 110 this will matter

Honestly we could even put a very high number or :infinity if Phoenix allows as we already take care of this 🤷

edgurgel: https://github.com/phoenixframework/phoenix/blob/52aa2cbd3b9283830ddfaf03173e20350c0b1d22/lib/phoenix/socket.ex#L724

It doesn't seem to support :infinity so I will just use a number like 99_999 😆

leandrocp: 99_999 works 😄

leandrocp (approved):

PR #2084, opened 2026-08-12
fix: bump liveview to fix bug

filipecabaco:

Bumping up the version of liveview so we tackle the bug reported in https://github.com/phoenixframework/phoenix_live_view/issues/3527


PragTob:
Wait, we already have another open PR to bump liveview haven't we?

PragTob:
@filipecabaco I believe this is mostly a duplicate of #2082 right? LV is also updated in there and found a couple of things

PR #2089, opened 2026-08-13
fix: check extension write policies lazily

### LINKED ISSUE #1953: Presence track() write-authorization probe also probes 'broadcast', logging noisy-but-benign RLS errors on Broadcast-from-Database setups

ryux1:
## Summary

- require every write-authorization probe to select exactly `:broadcast` or `:presence`
- keep the unselected extension policy unevaluated so each feature authorizes itself on first use
- pass the explicit extension from channel presence/broadcast handlers and the SingleBroadcast and BatchBroadcast paths
- add regression coverage for presence-only tracking and lazy broadcast authorization

Closes #1953

## Verification

- `mix compile --force --warnings-as-errors`
- `mix format --check-formatted`
- `mix credo` (no issues)
- `mix deps.unlock --check-unused`
- `mix sobelow --config .sobelow-conf`
- GitHub Actions database matrix requested; currently awaiting maintainer workflow approval
- `mix hex.audit` reports the locked `cowlib 2.19.0` advisory; this branch does not change `mix.lock`

### DIFF
--- a/lib/realtime/tenants/authorization.ex
+++ b/lib/realtime/tenants/authorization.ex
@@ -306,10 +306,9 @@ defmodule Realtime.Tenants.Authorization do
   @all_extensions [:broadcast, :presence]
 
   defp extensions_to_check(opts) do
-    Enum.filter(@all_extensions, fn
-      :broadcast -> Keyword.get(opts, :broadcast_enabled?, true)
-      :presence -> Keyword.get(opts, :presence_enabled?, true)
-    end)
+    if Keyword.get(opts, :presence_enabled?, true),
+      do: @all_extensions,
+      else: [:broadcast]
   end
 
   defp check_read_policies(conn, messages_by_extension, policies) do
--- a/lib/realtime/tenants/authorization.ex
+++ b/lib/realtime/tenants/authorization.ex
@@ -31,6 +31,8 @@ defmodule Realtime.Tenants.Authorization do
           :sub => binary | nil
         }
 
+  @type extension :: :broadcast | :presence
+
   @doc """
   Builds a new authorization struct which will be used to retain the information required to check Policies.
 
@@ -114,7 +116,7 @@ defmodule Realtime.Tenants.Authorization do
 
   Automatically uses RPC if the database connection is not in the same node
   """
-  @spec get_write_authorizations(Policies.t(), pid(), t(), keyword()) ::
+  @spec get_write_authorizations(Policies.t(), pid(), t(), extension()) ::
           {:ok, Policies.t()}
           | {:error, :rls_policy_error, Postgrex.Error.t()}
           | {:error, :query_canceled, Postgrex.Error.t()}
@@ -122,14 +124,13 @@ defmodule Realtime.Tenants.Authorization do
           | {:error, :increase_connection_pool}
           | {:error, :tenant_database_unavailable}
           | {:error, any()}
-  def get_write_authorizations(policies, db_conn, authorization_context, opts \\ [])
-
-  def get_write_authorizations(policies, db_conn, authorization_context, opts) when node() == node(db_conn) do
+  def get_write_authorizations(policies, db_conn, authorization_context, extension)
+      when extension in [:broadcast, :presence] and node() == node(db_conn) do
     rate_counter = rate_counter(authorization_context.tenant_id)
 
     if rate_counter.limit.triggered == false do
       db_conn
-      |> get_write_policies_for_connection(authorization_context, policies, opts)
+      |> get_write_policies_for_connection(authorization_context, policies, extension)
       |> handle_policies_result(rate_counter)
     else
       {:error, :increase_connection_pool}
@@ -137,7 +138,8 @@ defmodule Realtime.Tenants.Authorization do
   end
 
   # Remote call
-  def get_write_authorizations(policies, db_conn, authorization_context, opts) do
+  def get_write_authorizations(policies, db_conn, authorization_context, extension)
+      when extension in [:broadcast, :presence] do
     rate_counter = rate_counter(authorization_context.tenant_id)
 
     if rate_counter.limit.triggered == false do
@@ -145,7 +147,7 @@ defmodule Realtime.Tenants.Authorization do
              node(db_conn),
              __MODULE__,
              :get_write_authorizations,
-             [policies, db_conn, authorization_context, opts],
+             [policies, db_conn, authorization_context, extension],
              tenant_id: authorization_context.tenant_id,
              key: authorization_context.tenant_id
            ) do
@@ -164,8 +166,8 @@ defmodule Realtime.Tenants.Authorization do
     end
   end
 
-  def get_write_authorizations(db_conn, authorization_context),
-    do: get_write_authorizations(%Policies{}, db_conn, authorization_context)
+  def get_write_authorizations(db_conn, authorization_context, extension),
+    do: get_write_authorizations(%Policies{}, db_conn, authorization_context, extension)
 
   defp handle_policies_result(result, rate_counter) do
     case result do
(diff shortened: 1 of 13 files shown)


edgurgel (commented):
Hey thanks for the PR! 

It looks like we won't have a single place where we might pick both presence and broadcast and this function allows nothing to be selected.

How about we instead just receive as an argument what we want?

```elixir
get_write_authorizations(policies, db_conn, context, :presence)
get_write_authorizations(policies, db_conn, context, :broadcast)
```


We would also update `SingleBroadcast` and `BatchBroadcast` to pass `:broadcast`

Thoughts?

ryux1:
@edgurgel Addressed in `6108572b`. `get_write_authorizations/4` now requires exactly `:broadcast` or `:presence`; the default and combinable selector options are gone. Broadcast and presence channel callers, SingleBroadcast, and BatchBroadcast now pass the explicit extension, and the tests assert that the other extension remains unevaluated. I also rebased the branch onto current `main`; compile with warnings-as-errors, formatting, Credo, unused-dependency checks, and Sobelow pass locally. The database-backed Actions matrix has been requested and is waiting for maintainer workflow approval.

lib/realtime/tenants/authorization.ex
```diff
@@ -305,9 +306,10 @@ defmodule Realtime.Tenants.Authorization do
   @all_extensions [:broadcast, :presence]
 
   defp extensions_to_check(opts) do
-    if Keyword.get(opts, :presence_enabled?, true),
-      do: @all_extensions,
-      else: [:broadcast]
+    Enum.filter(@all_extensions, fn
+      :broadcast -> Keyword.get(opts, :broadcast_enabled?, true)
+      :presence -> Keyword.get(opts, :presence_enabled?, true)
+    end)
```
edgurgel: Is this change still necessary?

edgurgel (approved):
Added a small comment but other than that everything else looks good 👍 

Thanks for the PR!

lib/realtime/tenants/authorization.ex
ryux1: No, this is no longer necessary. With write authorization now selecting exactly one extension explicitly, `extensions_to_check/1` is only part of the existing read-authorization path. I restored its prior behavior in `16d9469c` so this PR does not broaden read selection.

ryux1:
@edgurgel Addressed the remaining review comment in `16d9469c`. The `broadcast_enabled?` read-selection change was no longer needed after making write authorization require an explicit extension, so I restored `extensions_to_check/1` to its previous read-only behavior. The explicit `:broadcast | :presence` write interface remains unchanged. Compilation with warnings as errors, formatting, Credo, the unused-dependency check, and Sobelow all pass locally. The fresh Actions runs are awaiting maintainer approval.

edgurgel:
Thanks, @ryux1  🎉

PR #2096, opened 2026-08-14
fix: contain a raising RLS policy to the subscription it came from

### LINKED ISSUE #2093: apply_rls: one subscription whose RLS evaluation raises silently kills postgres_changes delivery for ALL subscribers
## Bug report

One subscriber whose stored claims make RLS evaluation raise causes `postgres_changes` delivery to stop **for every subscriber on the project** — silently. We hit this in production on 2026-08-13 and lost all realtime delivery for ~2.5 hours, platform-wide.

### Environment

- Hosted Supabase, eu-north-1 (project ref available via support ticket on request)
- Realtime 2.125.1 (slot `supabase_realtime_replication_slot_2_125_1_*`)
- Postgres 15, RLS enabled on all published tables

### What happened

1. For ~2 hours (coinciding with a platform-side restart of our project's Auth + Realtime at ~05:04 UTC), GoTrue-issued access tokens carried the **string** `"null"` instead of JSON null in nullable `app_metadata` claims (e.g. `app_id`). Our access-token hook stamps proper JSON nulls and was byte-identical on our dev project, which was unaffected — the mangling was platform-side and stopped on its own at ~07:19 UTC. That is a separate Auth issue; it matters here only as the trigger.
2. Our RLS policies call a helper that does `NULLIF(auth.jwt() -> 'app_metadata' ->> 'app_id', '')::uuid`. With the claim being the string `"null"`, this raises `22P02 invalid input syntax for type uuid: "null"`.
3. `realtime.apply_rls()` evaluates policies per subscription using the claims stored at subscribe time. When evaluation for **one** subscription raises, the error aborts processing of the WAL batch — and **no subscriber receives anything**, including `service_role` subscriptions and subscriptions on unrelated tables.

### Observed symptoms

- WebSocket connects (101), channels reach `SUBSCRIBED`, rows appear in `realtime.subscription` with correct claims/filters — everything looks healthy.
- The replication slot stays `active` with `confirmed_flush_lsn` advancing to `pg_current_wal_lsn()` — WAL is consumed and acknowledged.
- Zero `postgres_changes` frames are delivered to any subscriber (verified with a `service_role` client subscribed with and without filters, on multiple tables).
- Reproduced directly: calling `realtime.apply_rls()` with a wal2json-shaped record while the poisoned subscriptions were present raises `22P02` with context `SQL statement "execute walrus_rls_stmt"`; after hardening our helper to not raise, the same call returns records and delivery resumed instantly for all subscribers.

### Expected behavior

An error while evaluating RLS for one subscription should disqualify **that subscription** (arguably with an error surfaced to that channel), not abort delivery for every subscriber on the project. JWT claims are external input from the subscriber's perspective of the pipeline — any upstream serializer defect (as happened here, in the platform's own Auth layer) becomes a total, silent, project-wide realtime outage that no health signal exposes: sockets up, subscriptions registered, slot advancing, nothing delivered.

### Workaround for anyone else hitting this

Make every function reachable from an RLS policy raise-proof against malformed claims — validate instead of casting:

```sql
CREATE OR REPLACE FUNCTION public.claim_uuid(p_claim text)
RETURNS uuid LANGUAGE sql IMMUTABLE SET search_path = '' AS $$
  SELECT CASE
    WHEN p_claim ~* '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
    THEN p_claim::uuid
  END;
$$;
```

Diagnosis query that found our poisoned subscriptions, in case it helps others:

```sql
SELECT jsonb_typeof(claims->'app_metadata'->'app_id'), count(*)
FROM realtime.subscription GROUP BY 1;
-- 'string' where you expect 'null' = poisoned tokens
```

Happy to provide timestamps, the project ref, and full reproduction detail through a support ticket — please reference this issue.

hamodywe:
Closes #2093.

## The problem

`realtime.apply_rls` checks each subscription against the table's policies with a bare

```sql
execute 'execute walrus_rls_stmt' into subscription_has_access;
```

Nothing catches an error there. A policy that raises on one subscription's stored claims — a malformed claim cast to `uuid` is enough — propagates out of `apply_rls` and takes down the entire `list_changes` call, so **every** subscriber on the tenant loses the batch, including subscriptions on other roles and other tables whose own policies cannot fail.

It is silent from every angle an operator would look at: sockets connect, `realtime.subscription` looks correct, the slot stays active. And the changes are not replayed — `confirmed_flush_lsn` advances during decoding even though the statement errored, so each affected batch is dropped for good. That is what turned this into a ~2.5 hour project-wide outage in #2093.

## The change

The per-subscription check goes into its own block. A subscription whose policy raises is disqualified on its own and collected into `errored_role_sub_ids`; every other subscription is evaluated and delivered normally.

Those subscriptions then get a row of their own carrying `Error 500: Internal Server Error, RLS policy evaluation failed`, with only `schema`, `table` and `type` in the payload — the same shape the existing 400 and 401 rows use — so a policy that could not be evaluated never ends up disclosing the row it was meant to gate. Silently dropping the subscription instead would have turned a loud tenant-wide outage into a silent per-subscriber one, which is the part of #2093 that made it expensive. Happy to cut that half if you would rather keep the change minimal.

The warning follows the pattern `realtime.send` already uses (`WarnSendingBroadcastMessage`), with the subscription id in it so the poisoned row can be found. Note that `list_changes` runs with `SET log_min_messages TO 'fatal'`, so in the polling path this reaches the client rather than the server log; the `errors` array is the reliable signal there.

Everything else in the function is byte-identical to `20260709120000_fix_apply_rls_filter_role_leak`.

## Verification

Elixir side, in `subscriptions_test.exs`: a poisoned subscription alongside a healthy one on the same role, and alongside a subscription on a different role. Asserts the healthy subscriptions still receive the record, the poisoned one receives the error row without a `record` key, and the batch is consumed rather than left behind.

**I could not run `mix test` locally** — this is a Windows box with no Elixir toolchain and the suite provisions tenant databases through its own Docker backend. So I verified the SQL directly instead, on `supabase/postgres:17.6.1.127` with the tenant migrations applied, and the tests are written to run on CI:

- **Reproduced first.** Three subscriptions on one table (two `anon`, one `authenticated` with `using (true)`), one `anon` carrying `app_metadata.app_id` as the string `"null"`. Pre-fix: `ERROR: invalid input syntax for type uuid: "null"` at `execute walrus_rls_stmt`, and all three subscriptions get nothing.

- **After the change**, same scenario:

  ```
             subscription_ids            |                               errors                               | delivered_details
  ----------------------------------------+--------------------------------------------------------------------+-------------------
   {11111111-...} (anon, healthy)          | {}                                                                 | hello
   {22222222-...} (anon, poisoned)         | {"Error 500: Internal Server Error, RLS policy evaluation failed"} |
   {33333333-...} (authenticated)          | {}                                                                 | hello
  ```

- **No behaviour change when nothing raises.** Ran a scenario covering INSERT/UPDATE/DELETE, a `selected_columns` subscription, two roles, and a table with no primary key (the 400 path) against both versions of the function and diffed the output: identical apart from the generated ids.

- **Cost of the extra subtransaction**, interleaved A/B on the same box, 100 `apply_rls` calls per run, best of 5, 100 subscriptions on one record:

  | round | before | after |
  |---|---|---|
  | 1 | 250.9 ms | 260.6 ms |
  | 2 | 259.1 ms | 266.3 ms |
  | 3 | 252.6 ms | 261.3 ms |

  About 3%, roughly 1 µs per subscription evaluation. If that is too much for the hot path, the alternative is to keep the loop unguarded and re-run it with per-subscription isolation only after something raises — zero cost normally, at the price of duplicating the loop. I went with the simpler version because this is a correctness fix in a function that is already hard to read; say the word and I will switch it.


hsusul:
Thanks for the clear repro and fix here, @hamodywe — the approach looks right and matches what @edgurgel confirmed on #2093.

This PR has been `CONFLICTING`/`DIRTY` against `main` since 2026-08-17, so I rebased the same migration + regression tests onto current `main` (and synced `apply_rls.sql`) in a superseding PR:

https://github.com/supabase/realtime/pull/2175

Happy to close this one in favor of #2175, or adjust if you'd rather refresh this branch yourself.

edgurgel:

PR #2097, opened 2026-08-15
fix: index subscription filters by digest so large in lists fit

### LINKED ISSUE #1670: Realtime subscription fails with large id.in(...) filters due to subscription_subscription_id_entity_filters_key btree size limit (ERROR 54000)
# Bug report

- [ x ] I confirm this is a bug with Supabase, not with my own application. 
- [ x ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When subscribing to Postgres changes with a large `id.in.(...)` filter list, Supabase Realtime fails to register the subscription. The server returns ERROR 54000 (program_limit_exceeded) because the subscription table index row size exceeds the btree limit for subscription_subscription_id_entity_filters_key. This prevents realtime updates from being delivered.

## To Reproduce

1. Create a Realtime channel and register a postgres_changes subscription with a filter containing a large list of IDs (e.g., id=in.(<60+ UUIDs>)).
2. Subscribe to the channel.
3. Observe the server error in the realtime logs.

Example supabase-js:

```
const channel = supabase.channel('db-changes');

channel.on(
  'postgres_changes',
  {
    event: '*',
    schema: 'public',
    table: 'destinations',
    filter: `id=in.(${hugeListOfUuids.join(',')})`,
  },
  payload => {
    console.log(payload);
  }
);

channel.subscribe();
```

Observed from Supabase:

```
receive error realtime:db-changes system {
  "channel": "db-changes",
  "extension": "postgres_changes",
  "message": "Unable to subscribe to changes with given parameters. An exception happened so please check your connect parameters: [schema: public, table: destinations, filters: [{\"id\", \"in\", \"{...}\"}]]. Exception: ERROR 54000 (program_limit_exceeded) index row size 3056 exceeds btree version 4 maximum 2704 for index \"subscription_subscription_id_entity_filters_key\" ...",
   "table": subscription
   "constraint": subscription_subscription_id_entity_filters_key
    "hint": Values larger than 1/3 of a buffer page cannot be indexed.
Consider a function index of an MD5 hash of the value, or use full text indexing.

Index row references tuple (7,3) in relation \"subscription\".", "status": "error"}
}
```

## Expected behavior

According to [the documentation](https://supabase.com/docs/guides/realtime/postgres-changes#contained-in-list-in) this filter is only limited to 100 values, but if there is also a size limit, this should also be documented. This error can only be surfaced by enabling realtime debug logs in the Supabase constructor.

Ideally, Supabase should either this this to accept the subscription regardless of filter size (perhaps using the MD5 fix documented in the returned error), or return a clearer error / limit guidance before attempting to insert the oversized filter into the subscription table (e.g., reject with a user‑facing limit or documented max size).

## System information

- Version of supabase-js: [e.g. 6.0.2]
- Running against Supabase Cloud (Hosted)

## Additional context

This is a React Native app. We dynamically subscribe to a growing set of destination IDs. Once the filter list grows large, realtime registration fails with the error above and no updates are delivered. If there’s a documented max filter size or a recommended workaround (e.g., server-side views, channel per subset, etc.), that would be helpful.

This can currently be worked around by batching the IDs so that they don't get near the limit (10-20 per batch)

hamodywe:
Closes #1670.

## The problem

`realtime.subscription`'s unique index covers `filters` by value:

```sql
CREATE UNIQUE INDEX ... ON realtime.subscription
  (subscription_id, entity, filters, action_filter, coalesce(selected_columns, '{}'));
```

so a subscription's whole filter array has to fit btree's 2704-byte tuple limit. An `id=in.(...)` list of uuids passes that quickly, and the subscription fails to register with `54000 index row size ... exceeds btree version 4 maximum 2704`.

Measured on `supabase/postgres:17.6.1.127` by inserting `in` filters of growing length:

```
largest in-list that inserts: 69 uuids
first failure at 70 uuids: 54000 index row size 2712 exceeds btree version 4 maximum 2704
```

**69.** Meanwhile `realtime.subscription_check_filters` raises `too many values for 'in' filter. Maximum 100`, and the docs describe the same 100. So the limit the code enforces has never been reachable on a uuid column, and because the real constraint is total byte length rather than value count, the same list works on an `int` column and fails on a `uuid` one — which is what makes it look arbitrary from the client side.

## The change

Index a digest of the array instead of the array:

```sql
create function realtime.filters_hash(filters realtime.user_defined_filter[]) returns bytea
  language sql immutable strict parallel safe
as $$ select pg_catalog.sha256(pg_catalog.convert_to(filters::text, 'UTF8')) $$;
```

and point the upsert's conflict target at the same expression. The index row is now fixed-width, uniqueness is unchanged, and the 100-value guard becomes the thing that actually stops you.

sha256 rather than the md5 the Postgres hint suggests, since the value is user-controlled and the digest is what decides whether two subscriptions are the same row; a collision would silently fold one subscription into another. `grant execute` follows the pattern of `20260706120000_grant_check_equality_op_5_arg`, since the blanket grant in `20231204144023` only covered functions that existed then.

## Why the tenant dumps are in this PR

`Realtime.Tenants.Migrations.load_db_dump/1` provisions a brand-new tenant from `priv/repo/tenant_db_dump_<major>.sql` and returns `Enum.count(migrations())` — it records every migration as applied without replaying any of them. A new tenant would therefore keep the old value index while being marked fully migrated, and the new `on conflict` target would have nothing to match, so **every** `postgres_changes` subscription on that tenant would fail, not just the large ones.

`update-tenant-db-snapshots.yml` is gated to same-repo PRs (`if: github.event.pull_request.head.repo.full_name == github.repository`), so it will not run here. I updated the three snapshot files by hand — the function, its ACL block, the index, the `schema_migrations` row, and the `tenant_schema` entries — and verified the result rather than trusting the edit: a fresh database loaded from the edited `tenant_db_dump_17.sql` reports 82 migrations (matching `migrations.ex`), has `subscription_subscription_id_entity_filters_hash_key` and `realtime.filters_hash`, and accepts a 100-uuid filter through the exact upsert statement from `subscriptions.ex`. Please regenerate them properly before merge if you would rather not carry a hand-edit.

## Verification

Tests added to `subscriptions_test.exs`: a 100-value `in` filter is accepted, and resubscribing the same large filter updates the existing row instead of duplicating it (that one guards the conflict-target change).

**I could not run `mix test` locally** — Windows box, no Elixir toolchain, and the suite provisions tenant databases through its own Docker backend. I verified the SQL directly instead, against a real tenant schema on `supabase/postgres:17.6.1.127`:

- **Before:** 69 uuids maximum, 70 fails with 54000.
- **After:** 100 uuids insert; 101 fails with `P0001 too many values for 'in' filter. Maximum 100` — the intended guard, reached for the first time.
- **Upsert identity preserved:** resubscribing the same 100-value filter leaves one row with updated claims; a genuinely different 100-value filter on the same channel is still its own row; the new index is the one enforcing both.
- **End to end:** with two subscriptions carrying different 100-value lists, an insert whose id is in the first list is delivered only to that subscription.


edgurgel:

PR #2099, opened 2026-08-15
fix: validate a filter's operator against the column type by evaluating it

### LINKED ISSUE #2098: subscription_check_filters accepts filter operators apply_rls cannot evaluate, and refuses some it can
## Describe the bug

`realtime.subscription_check_filters` decides whether a filter's operator can be applied to a column type by looking the operator up in `pg_catalog.pg_operator`. That lookup disagrees with what `check_equality_op` actually does at WAL time, in both directions.

**Accepted at subscribe time, raises at WAL time.** The comparison operators are not checked against the column type at all — the `else` branch only checks that the value casts:

```sql
else
    -- eq/neq/lt/lte/gt/gte: value must be coercable to the type
    perform realtime.cast(filter.value, col_type);
end if;
```

So `body=eq.1` on a `json` column registers happily, and `apply_rls` then raises `operator does not exist: json = json`. `is_visible_through_filters` runs in `apply_rls`' subscription loop, so this aborts the whole `list_changes` call: every subscription on the tenant loses the batch, including subscriptions on other tables with no filters at all, and the WAL is not replayed. One client is enough to stop `postgres_changes` for a whole project.

`ilike` on a `bytea` column has the same effect for a different reason: the guard checks `oprname = '~~'` for both `like` and `ilike`, and `bytea` has `~~` but no `~~*`.

**Refused at subscribe time although it works.** `varchar` carries no `~~`/`~`/`~*` entry of its own — it resolves them through `text` — so the `oprleft = col_type` test fails and `like`, `ilike`, `match`, `imatch` are rejected on `varchar` columns that handle them perfectly well.

## To Reproduce

On `supabase/postgres:17.6.1.127` with the tenant migrations applied:

```sql
create table public.docs (id int primary key, body json);
create table public.notes (id int primary key, details text);
grant all on table public.docs, public.notes to anon;
create publication supabase_realtime_test for all tables;

-- one filtered subscription on the json column, two ordinary ones
insert into realtime.subscription (subscription_id, entity, claims, filters) values
  ('11111111-1111-1111-1111-111111111111', 'public.docs'::regclass, '{"role":"anon"}',
   array[row('body', 'eq', '{"a":1}', false)::realtime.user_defined_filter]),   -- accepted
  ('22222222-2222-2222-2222-222222222222', 'public.docs'::regclass, '{"role":"anon"}', '{}'),
  ('33333333-3333-3333-3333-333333333333', 'public.notes'::regclass, '{"role":"anon"}', '{}');

select pg_create_logical_replication_slot('blast_slot', 'wal2json');
insert into public.docs  (id, body) values (1, '{"a":1}');
insert into public.notes (id, details) values (1, 'unrelated table');

select subscription_ids, errors from realtime.list_changes('supabase_realtime_test', 'blast_slot', 100, 1048576);
```

```
ERROR:  operator does not exist: json = json
QUERY:  select '{"a": 1}'::json = ('{"a":1}'::json)
CONTEXT:  PL/pgSQL function realtime.check_equality_op(realtime.equality_op,regtype,text,text,boolean) line 53 at EXECUTE
          SQL function "is_visible_through_filters" statement 1
          PL/pgSQL function realtime.apply_rls(jsonb,integer) line 191 at FOR over SELECT rows
          SQL function "list_changes" statement 1
```

Nobody receives anything — not the unfiltered subscription on `docs`, and not the one on `notes`.

## Full sweep

I ran every operator against a table with 17 column types, comparing what `subscription_check_filters` accepts against what `check_equality_op` can evaluate:

| Column type | Operators accepted that raise at WAL time |
|---|---|
| `json` | `eq` `neq` `lt` `lte` `gt` `gte` `isdistinct` |
| `xml` | `eq` `neq` `lt` `lte` `gt` `gte` `isdistinct` |
| `point` | `eq` `lt` `lte` `gt` `gte` `isdistinct` |
| `bytea` | `ilike` |

| Column type | Operators refused although they evaluate fine |
|---|---|
| `character varying` | `like` `ilike` `match` `imatch` |

**21 accepted-but-raises, 4 refused-but-works.**

## Expected behavior

A filter that registers should be evaluable, and a filter that is evaluable should register. Anything that can raise inside `apply_rls` needs to be caught at subscribe time, where it can be reported to the one client responsible instead of taking down the tenant's stream — which is the reason the `is` and `match` guards were added in the first place.

## System information

- `supabase/postgres:17.6.1.127`, tenant migrations through `20260709120000`
- Reproduced against `realtime.list_changes` directly

I have a fix ready and will open a PR against this issue.

hamodywe:
Closes #2098.

## The problem

`subscription_check_filters` answers "can this operator be applied to this column type?" with a `pg_operator` lookup. `check_equality_op` answers it by building SQL and letting the planner resolve it. Those two answers differ, in both directions.

- **`varchar` carries no `~~` of its own** — it resolves one through `text` — so `oprleft = col_type` fails and `like`/`ilike`/`match`/`imatch` are refused on varchar columns that handle them fine.
- **`bytea` carries `~~` but no `~~*`**, and the guard checks `~~` for both operators, so `ilike` on a bytea column is accepted and raises at WAL time.
- **The comparison operators are not checked at all.** The `else` branch only verifies the value casts, so `eq` on a `json`, `xml` or `point` column registers and raises inside `apply_rls`.

The last one is the expensive case. `is_visible_through_filters` is evaluated in `apply_rls`' subscription loop, so the raise aborts the whole `list_changes` call: every subscription on the tenant loses the batch — other tables and other roles included — and because the slot's `confirmed_flush` advances during decoding, the batch is not replayed. One client subscribing to `body=eq.x` on a json column stops `postgres_changes` for the entire project. That is the same failure shape as #2093, reached through the filter path rather than the RLS path, and it is not covered by the containment in #2096 because `is_visible_through_filters` runs outside that block.

## The change

Stop asking the catalog and ask the operator. The branch now probes `check_equality_op` — the exact call `apply_rls` will make — inside its own block:

```sql
begin
    perform realtime.check_equality_op(filter.op, col_type, filter.value, filter.value, filter.negate);
exception when others then
    raise exception 'operator % is not supported on column % of type %: %',
        filter.op::text, filter.column_name, col_type::text, sqlerrm;
end;
```

Because the probe *is* the runtime path, subscribe-time acceptance and WAL-time evaluation cannot drift apart — which no catalog predicate can guarantee, since operator resolution goes through implicit and binary-coercible casts that `pg_operator` does not record. The two `pg_operator` guards are removed as a result; the `is` keyword checks and the eager regex compile stay, the latter so a bad pattern keeps its own precise error instead of arriving as "unsupported operator".

The client now gets, at subscribe time:

```
operator eq is not supported on column body of type json: operator does not exist: json = json
```

## Verification

I swept every operator against a table covering 17 column types, comparing what `subscription_check_filters` accepts against what `check_equality_op` can evaluate.

| | before | after |
|---|---|---|
| accepted at subscribe, raises at WAL time | **21** | **0** |
| refused at subscribe, evaluates fine | **4** | **0** |

The 21: `json` and `xml` on `eq neq lt lte gt gte isdistinct`, `point` on `eq lt lte gt gte isdistinct`, `bytea` on `ilike`. The 4: `varchar` on `like ilike match imatch`.

Also verified end to end, on `supabase/postgres:17.6.1.127` with the tenant migrations applied:

- **Before:** the three-subscription scenario from #2098 delivers nothing to anyone.
- **After:** the json filter is refused at subscribe time; the unfiltered subscriptions on `docs` and on `notes` both receive their rows; and a `label=like.rele%` filter on a `varchar` column — previously refused outright — is accepted and correctly gates delivery.

Tests added to `subscriptions_test.exs` cover all three: the json rejection, the bytea `ilike` rejection, varchar `like` being accepted, and a varchar `like` subscription gating rows end to end through `list_changes`.

**The tenant dumps are in this PR** for the same reason as #2097: `Realtime.Tenants.Migrations.load_db_dump/1` provisions a new tenant from `priv/repo/tenant_db_dump_<major>.sql` and then records every migration as applied without replaying any, so a stale dump would leave new tenants with the old trigger and the bug intact. `update-tenant-db-snapshots.yml` is gated to same-repo PRs so it will not run here; I edited the three snapshot files by hand and checked the result rather than the edit — a fresh database loaded from the edited `tenant_db_dump_17.sql` reports 82 migrations, matching `migrations.ex`, and re-running the full matrix against it gives 0 and 0. Please regenerate them properly before merge if you would rather not carry a hand-edit.

**I could not run `mix test` locally** — Windows box with no Elixir toolchain, and the suite provisions tenant databases through its own Docker backend. Everything above was verified at the SQL level instead; the Elixir tests are written to the file's existing patterns and run on CI.

One note on ordering: this and #2097 both add a line to the migration list in `lib/realtime/tenants/migrations.ex` and both touch the dumps, so whichever merges second needs a trivial rebase.


edgurgel:

PR #2100, opened 2026-08-16
fix(docker): only sudo to nobody when running as root

### LINKED ISSUE #2063: Error running on Kubernetes with runAsNonRoot: true and allowPrivilegeEscalation: false
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The use of sudo in run.sh prevent the realtime container from running on Kubernetes with securityContext settings runAsNonRoot: true and allowPrivilegeEscalation: false

## To Reproduce

Create a deployment on kubernetes with:
```
Pod security context:
  runAsNonRoot: true
  runAsUser: 1000
  runAsGroup: 1000
  seccompProfile:
    type: RuntimeDefault
Container securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
```
Or use helm chart from https://github.com/supabase-community/supabase-kubernetes with values:
```
deployment:
  realtime:
    podSecurityContext:
      runAsNonRoot: true
      runAsUser: 1000
      runAsGroup: 1000
      seccompProfile:
        type: RuntimeDefault
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
          - ALL
```
The pods will fail to start.

## Expected behavior

Realtime docker containers should allow running as non-root:

## System information

- OS: Linux

## Additional context

One way to fix this (not tested) is to use sudo only if running as root, otherwise run the migrations as the user running the container:

```
if (( EUID == 0 )); then
    sudo -E -u nobody -- /app/bin/migrate
else
    /app/bin/migrate
fi

if [ "${SEED_SELF_HOST-}" = true ]; then
    echo "Seeding selfhosted Realtime"
    if (( EUID == 0 )); then
        sudo -E -u nobody /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'
    else
        /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'
    fi
fi
```

hamodywe:
Closes #2063.

`run.sh` drops to `nobody` with an unconditional `sudo` before running migrations and the self-host seed. The image sets no `USER`, so the container normally starts as root and that works.

It breaks under an orchestrator that already starts the container as a non-root user. With Kubernetes `runAsNonRoot: true` and `allowPrivilegeEscalation: false`, the kernel ignores sudo's setuid bit, so the call fails and the container never starts.

This guards both calls on the effective uid: when we are root, drop to `nobody` exactly as before; when we are already unprivileged, run the command directly, since there is nothing to drop. Since the image sets no `USER`, this only changes behaviour for orchestrators that already override the runtime UID.

**Verification**

Built an image mirroring this one (sudo installed, `/app` owned by `nobody`, no `USER`) and ran both paths:

- Started as root: migrate runs as `nobody` (uid 65534) both before and after the change — behaviour unchanged.
- Started with `--user 1000:1000 --security-opt no-new-privileges --cap-drop ALL`: the current code fails with `sudo: The "no new privileges" flag is set, which prevents sudo from running as root` (exit 1), while the guarded version runs as uid 1000 (exit 0).

`bash -n run.sh` is clean.


hsahmed:
@edgurgel can you please review this PR. This is the last change required to allow Supabase to run on restricted Kubernetes clusters.

edgurgel:
Hey team have you managed to test this in your environment?

I'm not an expert on this subject but wouldn't it be enough to add 

```
USER nobody
```

just before the `ENTRYPOINT` definition in the Dockerfile?

WDYT?

Then drop the sudo and the new function completely?

hsahmed:
I have tested adding the USER nobody to the Dockerfile and everything is working fine. Tested the following:

Build the container
Deploy it and test subscribe and broadcast functionality from a Python client (using supabase-py)

There are no errors and the migration and seed scripts run correctly. I have also tested running it on a restricted Kubernetes environment using the following settings:

      realtime:
        podSecurityContext:
          runAsNonRoot: true
          runAsUser: 65534
          runAsGroup: 65534
          seccompProfile:
            type: RuntimeDefault
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
              - ALL

I have submitted a new PR with the changes at https://github.com/supabase/realtime/pull/2207

edgurgel:
Merged https://github.com/supabase/realtime/pull/2207 and a new release is out. Thanks, team

PR #2104, opened 2026-08-17
fix: support jsonb expression filters

### LINKED ISSUE #1090: Postgres changes with jsonb filters
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When using realtime postgres_changes, I noticed that filtering on jsonb columns doesn't work. I *maybe* have a bug somewhere but without the filter or with a filter on a regular text column works.

```ts
const channel = supabase
  .channel(`queue:${organization.id}`)
  .on(
    "postgres_changes",
    {
      event: "*",
      schema: "pgboss",
      table: "job",
      filter: `data->>organization_id=eq.${organization.id}`,
    },
    (payload) => {
      console.log("queue/pgboss/job", payload);
    }
  )
  .subscribe();
```
## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a table with a jsonb column (and enable realtime/rls/etc on that table)
2. Create realtime subscription for postgres_changes and target that table and filter on a jsonb field value you know is present.
3. Make a change to the table 
4. See no changes are published

## Expected behavior

I would expect the jsonb->>field=eq.value to work as it does with supabase-js or regular postgrest requests.

## System information

- OS: macOS
- Version of supabase-js: 2.39.7
- Version of Node.js: 20

qtjg:
## Summary

This PR adds support for filtering Postgres Changes subscriptions on a JSONB text-extraction expression such as `data->>organization_id=eq.org_123`.

Fixes #1090.

## What changed

- Validate a single `jsonb_column->>key` filter expression during subscription creation.
- Require the base column to have the `jsonb` type and validate the comparison value as text.
- Extract the requested JSONB key with PostgreSQL’s `->>` operator during row-visibility evaluation.
- Add a `data jsonb` field to the shared integration fixture.
- Add an end-to-end regression test proving that a matching JSONB value is delivered and a different value is ignored.

The implementation keeps the existing filter wire format and equality-operator machinery. It does not use dynamic SQL for the JSONB key; the base column and key are parsed and validated before the value is passed to the existing comparison function.

## Testing

- The standalone Elixir parser check for `data->>organization_id=eq.org_123` passes.
- `git diff --check` passes.
- The integration test is included but the full Mix suite was not run in this environment because the repository pins Elixir 1.19.5 / OTP 28, while the available runtime is Elixir 1.14 / OTP 25.

## Coordination note

A previous comment on issue #1090 says that `krishvsoni` wanted to work on this issue. I am opening this PR transparently so maintainers can confirm whether the work is still available. If that contributor is actively implementing the fix, I am happy to defer or adapt this branch based on maintainer guidance.

## Review focus

Please review the accepted expression syntax, behavior for missing JSONB keys and `NULL`, and interactions with `not.`, `in`, `is null`, and delete/update old-row filtering.

### DIFF
--- a/priv/repo/tenant_schema/schemas/realtime/functions/is_visible_through_filters.sql
+++ b/priv/repo/tenant_schema/schemas/realtime/functions/is_visible_through_filters.sql
@@ -14,8 +14,14 @@ create or replace function realtime.is_visible_through_filters (
             and sum(
                 realtime.check_equality_op(
                     op:=f.op,
-                    type_:=coalesce(col.type_oid::regtype, col.type_name::regtype),
-                    val_1:=col.value #>> '{}',
+                    type_:=case
+                        when position('->>' in f.column_name) > 0 then 'text'::regtype
+                        else coalesce(col.type_oid::regtype, col.type_name::regtype)
+                    end,
+                    val_1:=case
+                        when position('->>' in f.column_name) > 0 then col.value ->> btrim(split_part(f.column_name, '->>', 2), ' "')
+                        else col.value #>> '{}'
+                    end,
                     val_2:=f.value,
                     negate:=coalesce(f.negate, false)
                 )::int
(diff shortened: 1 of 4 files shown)


edgurgel:

leandrocp:
Hi @qtjg can you please create a new migration in `lib/realtime/tenants/repo/migrations/` and append into the `@migrations` list in lib/realtime/tenants/migrations.ex? Then CI will re-generate the files in priv/repo/tenant_schema/*
Thanks!

edgurgel:
And please add a test to `test/walrus`! These are standalone tests just for postgres changes functions. The existing tests should be a good example of what you might need

PR #2106, opened 2026-08-18
fix: statuslive view can crash from high memory usage

PragTob:
👋 

The main goal here is to stop the status page from crashing. 

The most important bit for the crash is:

```
Process: #PID<0.19739425.8> on node :"realtime@2600:1f1c:c01:4900:87b8:cc8e:10c4:1fb4"
Context: maximum heap size reached
Max Heap Size: 125000000
Total Heap Size: 149485546
Kill: true
Error Logger: true
Message Queue Len: 366999
```

I.e. the message queue is overly long.

The root cause here is that on every broadcast from `latency` we subbed to we used to do a re-render due to the change tracking via `assign`. The rendering would take too long, hence the messages would pile up and then 💥 . There's a lot of traffic since basically we get a message for every node pinging every other node once every 15 seconds. So, say 100 nodes - that's 100 * 100 / 15 = ~666 messages/s aka each render could only take ~1.5ms to keep up.

The fix applied here, stores the messages/changes triggered but renders them out only every second. That's still very fresh, but without doing all the additional work of rendering multiple times a second.

That change is implemented in df3f93e6eed502bd0402b318c020782db1754ece and is much smaller than the rest of these changes.

The remainder of the changes here are small conveniences and one big one: a script to load test this view, to verify the problem and the fix. The script sends fake messages to load test and polls the process state checking in on queue length and Memory consumption.

### Before Fix

**node_count=100**

```
[load] t=59.2s  #PID<0.1064.0>  sent=38860  actual=656.6/s mem= 35.9 MB queue_len= 35969
[load] t=60.2s  #PID<0.1064.0>  sent=39530  actual=656.6/s mem= 36.0 MB queue_len= 36623
[load] t=61.2s  #PID<0.1064.0>  sent=40200  actual=656.6/s mem= 36.1 MB queue_len= 37276
[load] ================= SUMMARY =================
[load] N=100  target_rate=666.7 msgs/s  duration=60.0s (ran 61.2s)
[load] messages sent: 40200  (actual throughput: 656.6 msgs/s)
[load] #PID<0.1064.0>:
[load]   queue_len: start=67  peak=37276  final=37276
[load]   memory:    start=  2.6 MB  peak= 51.0 MB  final= 36.1 MB
[load] =============================================
```

_I'll spare us the higher node counts_

### After Fix

**node_count=100**

```
[load] t=60.3s  #PID<0.1130.0>  sent=38860  actual=644.1/s mem=  9.6 MB queue_len=     0
[load] t=61.4s  #PID<0.1130.0>  sent=39530  actual=643.9/s mem=  9.6 MB queue_len=     0
[load] t=62.4s  #PID<0.1130.0>  sent=40200  actual=643.8/s mem=  9.6 MB queue_len=     0
[load] ================= SUMMARY =================
[load] N=100  target_rate=666.7 msgs/s  duration=60.0s (ran 62.4s)
[load] messages sent: 40200  (actual throughput: 643.8 msgs/s)
[load] #PID<0.1130.0>:
[load]   queue_len: start=0  peak=67  final=0
[load]   memory:    start=  4.7 MB  peak= 13.2 MB  final=  9.6 MB
[load] =============================================
```

**node_count=150**

```
[load] t=60.7s  #PID<0.1064.0>  sent=87000  actual=1433.1/s mem= 22.5 MB queue_len=     0
[load] t=61.8s  #PID<0.1064.0>  sent=88500  actual=1433.2/s mem= 16.4 MB queue_len=     0
[load] t=62.8s  #PID<0.1064.0>  sent=90000  actual=1433.3/s mem= 22.5 MB queue_len=     0
[load] ================= SUMMARY =================
[load] N=150  target_rate=1.5e3 msgs/s  duration=60.0s (ran 62.8s)
[load] messages sent: 90000  (actual throughput: 1433.2 msgs/s)
[load] #PID<0.1064.0>:
[load]   queue_len: start=0  peak=450  final=0
[load]   memory:    start=  7.6 MB  peak= 24.0 MB  final= 22.5 MB
[load] =============================================
```

**node_count=200**

```
[load] t=61.9s  #PID<0.1064.0>  sent=157530  actual=2546.4/s mem= 30.7 MB queue_len=   267
[load] t=62.9s  #PID<0.1064.0>  sent=160200  actual=2546.4/s mem= 32.8 MB queue_len=   267
[load] ================= SUMMARY =================
[load] N=200  target_rate=2666.7 msgs/s  duration=60.0s (ran 62.9s)
[load] messages sent: 160200  (actual throughput: 2546.3 msgs/s)
[load] #PID<0.1064.0>:
[load]   queue_len: start=0  peak=1069  final=267
[load]   memory:    start= 11.3 MB  peak= 45.6 MB  final= 32.8 MB
[load] =============================================
```

(notably this one once went up to 800 but down to 0 again so might still stabilize)

## Note-worthy

* I put the load testing script into a new `scripts` folder - arguably it could/should be in `bench` (although not a benchmark) - happy for feedback/thoughts
* The way I got the nodes to render in the load test is slightly sus/has an effect on prod code but should be negligible and without it the view renders basically nothing --> not a realistic test

### DIFF
--- a/lib/realtime_web/live/status_live/index.ex
+++ b/lib/realtime_web/live/status_live/index.ex
@@ -8,11 +8,11 @@ defmodule RealtimeWeb.StatusLive.Index do
   alias Realtime.Nodes
   alias RealtimeWeb.Endpoint
 
-  @ping_interval 15_000
+  @ping_interval to_timeout(second: 15)
   @stale_after @ping_interval * 2
-  @warn_above_ms 1_000
+  @warn_above_ms to_timeout(second: 1)
   @problem_limit 200
-  @flush_interval 1_000
+  @flush_interval to_timeout(second: 1)
   @state_key :node_info
 
   @impl true
--- a/lib/realtime_web/live/status_live/index.ex
+++ b/lib/realtime_web/live/status_live/index.ex
@@ -12,28 +12,33 @@ defmodule RealtimeWeb.StatusLive.Index do
   @stale_after @ping_interval * 2
   @warn_above_ms 1_000
   @problem_limit 200
+  @flush_interval 1_000
+  @state_key :node_info
 
   @impl true
   def mount(_params, _session, socket) do
     if connected?(socket) do
       Endpoint.subscribe("admin:cluster")
       schedule_staleness_check()
+      schedule_flush()
     end
 
     node_ids = all_nodes()
+    pair_status = default_pair_status(node_ids)
 
     socket =
       socket
       |> assign(:active_nav, :status)
       |> assign(:node_ids, node_ids)
       |> assign(:node_regions, %{})
-      |> assign(:pair_status, default_pair_status(node_ids))
+      |> assign(:pair_status, pair_status)
       |> assign(:selected_node, nil)
       |> assign(:region_filter, "")
       |> assign(:node_sort, "node")
       |> assign(:pair_sort, "severity")
       |> assign(:drill_sort, "node")
       |> assign(:mounted_at, DateTime.utc_now())
+      |> put_state(%{pair_status: pair_status, node_regions: %{}})
 
     {:ok, socket}
   end
@@ -68,11 +73,26 @@ defmodule RealtimeWeb.StatusLive.Index do
   @impl true
   def handle_info(%Phoenix.Socket.Broadcast{payload: %Payload{} = payload}, socket) do
     {pair, entry} = pair_entry(payload)
+    state = state(socket)
+
+    state = %{
+      state
+      | pair_status: Map.put(state.pair_status, pair, entry),
+        node_regions: Map.put(state.node_regions, payload.from_node, payload.from_region)
+    }
+
+    {:noreply, put_state(socket, state)}
+  end
+
+  # scheduled flushes/draws to avoid triggering rerenders too frequently
+  def handle_info(:flush, socket) do
+    schedule_flush()
+    state = state(socket)
 
     socket =
       socket
-      |> update(:pair_status, &Map.put(&1, pair, entry))
-      |> update(:node_regions, &Map.put(&1, payload.from_node, payload.from_region))
+      |> assign(:pair_status, state.pair_status)
+      |> assign(:node_regions, state.node_regions)
 
     {:noreply, socket}
   end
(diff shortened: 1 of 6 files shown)


lib/realtime_web/live/status_live/index.ex
```diff
@@ -123,9 +153,18 @@ defmodule RealtimeWeb.StatusLive.Index do
...
+  # to delay rerendering until the next `flush`
+  defp state(socket), do: socket.private[@state_key]
+  defp put_state(socket, state), do: put_private(socket, @state_key, state)
 
   defp all_nodes do
-    [Node.self() | Node.list()] |> Enum.map(&Nodes.short_node_id_from_name/1)
+    # Load-test-only escape hatch to have realistic rendering.
+    # Practically `[]` on prod and so shouldn't have a meaningful impact.
+    extra_ids = Application.get_env(:realtime, __MODULE__, []) |> Keyword.get(:extra_node_ids, [])
+    extra_ids ++ ([Node.self() | Node.list()] |> Enum.map(&Nodes.short_node_id_from_name/1))
```
PragTob, line 167: Not sure how we feel about this, it changes prod code for a load testing script to work.

But: It's only done on mount and then it's just an empty list, so I think it's ok but I'd understand if people would rather not have this around.

filipecabaco: no worries, I think it's an ok think to have mainly in this view but worth checking how we can do load testing in the future during development with less injecty code (find a pattern basically)

filipecabaco (approved):

lib/realtime_web/live/status_live/index.ex
PragTob: I couldn't find one easily here sadly, as it literally uses `Node.list()` and that's what we need to change unless we want to boot 100+ VMs.

I mean I guess:
* we could use mimic here in the dev/test env only? 🤔 
* We could further gate it by environment or some config option, so it's even more gated
* another way would be to rewrite the code - this was only necessary of the way that rendering traverses the nodes it knows and then writes statistics for them, if we started with the events themselves this might not be necessary but 🤷

leandrocp:
> I put the load testing script into a new scripts folder - arguably it could/should be in bench (although not a benchmark) - happy for feedback/thoughts

@PragTob scripts/ works for me but another idea would be to store such files into dev/ as dev/bench and dev/scripts (just an idea, no hard opinions)

PragTob:
@leandrocp good callout, I think I'll do it in a separate PR though (as it'd move the bench scripts and potentially reformat them due to the formatter config).

PR #2107, opened 2026-08-18
Fix: snabkaffex inclusion in root mix.lock

PragTob:
Quick follow up to: [supabase/realtime#2105](https://linear.app/supabase/review/fix-use-published-snabbkaffex-lib-de41059c6c47) - running `mix deps.get` now leaves changes to `mix.lock` in the root `mix.exs` as `forum` is a dependency of it.


leandrocp (approved):

edgurgel:
Thanks @PragTob !

PR #2109, opened 2026-08-18
feat: pg changes able to wait on subscription

filipecabaco:

adds a new opt in wait for pg changes to accept the connection.

```js
channel.join({
  config: {
    postgres_changes: [],
    postgres_changes_options: {
      wait: true,
      timeout: 15000 // optional, ms; clamped server-side to `postgres_changes_wait_max_timeout`
    }
  }
})
```

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,6 +14,7 @@ config :realtime,
   replication_watchdog_interval: :timer.minutes(5),
   replication_watchdog_timeout: :timer.minutes(1),
   replication_ready_timeout: :timer.minutes(1),
+  # Blocks the transport process, so keep it under the client heartbeat interval (30s)
   postgres_changes_wait_max_timeout: :timer.seconds(20)
 
 # Configures the endpoint
--- a/config/config.exs
+++ b/config/config.exs
@@ -13,7 +13,8 @@ config :realtime,
   version: Mix.Project.config()[:version],
   replication_watchdog_interval: :timer.minutes(5),
   replication_watchdog_timeout: :timer.minutes(1),
-  replication_ready_timeout: :timer.minutes(1)
+  replication_ready_timeout: :timer.minutes(1),
+  postgres_changes_wait_max_timeout: :timer.seconds(20)
 
 # Configures the endpoint
 config :realtime, RealtimeWeb.Endpoint,
--- a/lib/realtime_web/channels/realtime_channel.ex
+++ b/lib/realtime_web/channels/realtime_channel.ex
@@ -35,8 +35,8 @@ defmodule RealtimeWeb.RealtimeChannel do
 
   @confirm_token_ms_interval :timer.minutes(5)
   @replication_ready_check_interval 500
-  @postgres_subscribe_first_retry 100
-  @postgres_subscribe_max_retry 1_000
+  @postgres_subscribe_backoff_min 100
+  @postgres_subscribe_backoff_max 1_000
   @postgres_subscribe_fatal_reasons [:malformed_subscription_params, :subscription_insert_failed]
   @postgres_subscribe_error_code "RealtimeDisabledForConfiguration"
   @fullsweep_after Application.compile_env!(:realtime, :websocket_fullsweep_after)
@@ -953,12 +953,22 @@ defmodule RealtimeWeb.RealtimeChannel do
         :ok
 
       timeout ->
-        deadline = System.monotonic_time(:millisecond) + timeout
-        await_postgres_subscribe(socket, tenant, pg_change_params, deadline, @postgres_subscribe_first_retry)
+        await_postgres_subscribe(socket, %{
+          tenant: tenant,
+          pg_change_params: pg_change_params,
+          timeout: timeout,
+          deadline: System.monotonic_time(:millisecond) + timeout,
+          backoff:
+            Backoff.new(
+              backoff_min: @postgres_subscribe_backoff_min,
+              backoff_max: @postgres_subscribe_backoff_max,
+              backoff_type: :rand_exp
+            )
+        })
     end
   end
 
-  defp await_postgres_subscribe(socket, tenant, pg_change_params, deadline, interval) do
+  defp await_postgres_subscribe(socket, %{tenant: tenant, pg_change_params: pg_change_params} = wait) do
     case postgres_subscribe_attempt(tenant, pg_change_params) do
       {:ok, _response} ->
         send(self(), :postgres_changes_subscribed)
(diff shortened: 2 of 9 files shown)


lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -935,11 +927,115 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        Logger.metadata(metadata)
+
+        try do
+          retry_postgres_subscribe(socket, tenant, pg_change_params, subscriber, @postgres_subscribe_first_retry)
+        rescue
+          e -> log_error(socket, postgres_subscribe_error_code(:connect_failed), Exception.message(e))
+        catch
+          :exit, reason -> log_error(socket, postgres_subscribe_error_code(:connect_failed), reason)
+        end
+      end)
```
edgurgel: why do we need a task here? Could we just subscribe and wait?

filipecabaco: still testing out ideas but honestly not needed and removed it as it would very likely break a ton of things due to a stupid amount of task creation

config/config.exs
```diff
@@ -13,7 +13,8 @@ config :realtime,
   version: Mix.Project.config()[:version],
   replication_watchdog_interval: :timer.minutes(5),
   replication_watchdog_timeout: :timer.minutes(1),
-  replication_ready_timeout: :timer.minutes(1)
+  replication_ready_timeout: :timer.minutes(1),
```
edgurgel, line 16: Why 1 minute? I guess we need to document clients that the server has a max wait?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -935,11 +924,76 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+        maybe_log_warning(socket, @postgres_subscribe_error_code, error)
+        sleep_and_retry_postgres_subscribe(socket, tenant, pg_change_params, deadline, interval)
+    end
+  end
+
+  defp sleep_and_retry_postgres_subscribe(socket, tenant, pg_change_params, deadline, interval) do
+    case deadline - System.monotonic_time(:millisecond) do
+      remaining when remaining > 0 ->
+        Process.sleep(min(interval, remaining))
+        next_interval = min(interval * 2, @postgres_subscribe_max_retry)
```
edgurgel: Should we use `DBConnection.Backoff` like we have been using in other places (including in this file)

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -308,6 +308,74 @@ defmodule RealtimeWeb.RealtimeChannelTest do
       # It will try again in the future
       assert socket.assigns.pg_sub_ref != nil
     end
+
+    test "wait rejects the join when the subscription is not established in time", %{tenant: tenant} do
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
```
edgurgel, line 314: `s/stub/expect` to ensure this is actually called? You can specify how many times this will be called if necessary

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -308,6 +308,74 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+    end
+
+    test "wait does not start a connect attempt once the timeout has passed", %{tenant: tenant} do
+      test_pid = self()
+
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ ->
+        send(test_pid, :connect_attempt)
+        Process.sleep(300)
+        nil
+      end)
```
edgurgel: you can use `expect` and it will ensure that it's called just once no need to use send/receive

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -308,6 +308,74 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+
+    test "wait lets an in-flight subscription attempt finish before rejecting the join", %{tenant: tenant} do
+      test_pid = self()
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> {:ok, {test_pid, test_pid}} end)
+
+      stub(Extensions.PostgresCdcRls, :handle_after_connect, fn _, _, _, _ ->
+        send(test_pid, :after_connect_attempt)
+        Process.sleep(300)
+        {:error, :boom}
+      end)
```
edgurgel, line 340: same here on expect

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -308,6 +308,74 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+        {:error, "Too many database timeouts"}
+      end)
+
+      expect(Extensions.PostgresCdcRls, :handle_after_connect, fn _, _, _, _ -> {:ok, []} end)
+
+      assert {:ok, _reply, _socket} = join_waiting_for_postgres_changes(tenant, %{"wait" => true, "timeout" => 5_000})
+    end
+
+    test "wait does not gate a join without postgres_changes bindings", %{tenant: tenant} do
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
```
edgurgel: `s/stub/reject` ?

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -308,6 +308,74 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+
+      assert {:ok, %{postgres_changes: []}, _socket} =
+               join_waiting_for_postgres_changes(tenant, %{"wait" => true, "timeout" => 100}, [])
+    end
+
+    test "join is not gated when wait is not requested", %{tenant: tenant} do
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
```
edgurgel: `s/stub/reject` ?

mise.lock
```diff
@@ -74,6 +74,25 @@ checksum = "sha256:b01b0830076539b15e63e0bcecd71b38be7bd3784c4bcd0bf46339d579549
 url = "https://github.com/erlang/otp/releases/download/OTP-28.5.0.4/otp_win64_28.5.0.4.zip"
 url_api = "https://api.github.com/repos/erlang/otp/releases/assets/491644305"
 
+[[tools.erlang]]
```
edgurgel: any idea why this file has changed?

edgurgel (changes requested):

config/config.exs
filipecabaco: probably the server timeout time for new conns?

config/config.exs
```diff
@@ -13,7 +13,8 @@ config :realtime,
...
+  postgres_changes_wait_max_timeout: :timer.seconds(20)
```
edgurgel, line 18: Sorry I commented on the wrong one. Why 20 seconds here?

What happens when a client tries to use 1 minute? Will they get an error saying that the max timeout is 20 seconds instead?

edgurgel: I'm asking because I think 20 seconds is not a huge number and I thought of this max here as a "catch" to not wait forever etc. So maybe a more generous one like 2 minutes is enough? Not sure!

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -310,34 +310,31 @@ defmodule RealtimeWeb.RealtimeChannelTest do
     end
 
     test "wait rejects the join when the subscription is not established in time", %{tenant: tenant} do
+      expect(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
       stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
```
edgurgel, line 314: Do you still need this stub?

edgurgel (approved):
Added some comments. Other than those LGTM! Nice one

PR #2110, opened 2026-08-18
feat: simultaneous instances of dev/test nodes

leandrocp:

Improve dev and test env to allow running multiple instances at the same time, automatically. In short:

```
# terminal 1
❯ mise run dev
12:04:08.785 [info] Access RealtimeWeb.Endpoint at http://127.0.0.1:4000
[info] Forum[pink@127.0.0.1|users] Received DISCOVER request from node pink2@127.0.0.1

# terminal 2
❯ TENANT=foo mise run dev
[info] [libcluster:dev] connected to :"pink@127.0.0.1"
[info] Forum[pink2@127.0.0.1|users] Received DISCOVER request from node pink@127.0.0.1
[info] Access RealtimeWeb.Endpoint at http://127.0.0.1:4001
```

I've been struggling to code review other PRs locally and/or work on more than one branch at the same time, for example we can't run `mix test` to verify the test suite (which takes a while to finish) and run tests on another branch at the same time, or even run two Phx servers at the same time, or simulate multiple nodes. That's a major productivity killer. This PR fixes those problems.

## Main changes
- `mix setup` does only deps and assets setup so it runs first and doesn't depend on database
- add `TENANT` env to configure DB prefix, compose prefix, etc - defaults to `realtime-dev` so no changes to single tenant setups (current workflow)
- break db containers into realtime-db and tenant-db
- introduce `TestEnv` to coordinate port attributions and node naming

## Testing

```shell
# deps and any initial local env setup
❯ mix setup

# start realtime-dev tenant
❯ mise task run db-rm
❯ mise task run db-start
...
  tenant             realtime-dev
  realtime database  127.0.0.1:5432/postgres
  tenant database    127.0.0.1:5434/postgres
❯ mise run dev
...
[info] Access RealtimeWeb.Endpoint at http://127.0.0.1:4000

# start "another" tenant
❯ TENANT=another mise task run db-rm
❯ TENANT=another mise task run db-start
...
  tenant             another
  realtime database  127.0.0.1:5432/postgres
  tenant database    127.0.0.1:5435/postgres
❯ TENANT=another mise run dev
...
[info] Access RealtimeWeb.Endpoint at http://127.0.0.1:4001
```

Open http://localhost:4000/status and http://localhost:4001/status (or whatever ports were assigned).

Then run tests with `mix test` (default `realtime-dev` tenant) or `TENANT=another mix test`.

## Notes

There's no cluster isolation, every new node started connects to running nodes.


config/runtime.exs
```diff
@@ -49,8 +49,45 @@ gen_rpc_socket_recbuf = Env.get_integer("GEN_RPC_SOCKET_RECEIVE_BUFFER")
...
+
 gen_rpc_tcp_client_port = Env.get_integer("GEN_RPC_TCP_CLIENT_PORT", 5369)
-gen_rpc_tcp_server_port = Env.get_integer("GEN_RPC_TCP_SERVER_PORT", 5369)
+
+port_free? = fn port ->
+  case :gen_tcp.listen(port, [:inet, ip: {0, 0, 0, 0}, reuseaddr: true, active: false]) do
+    {:ok, socket} -> :gen_tcp.close(socket) == :ok
+    {:error, _reason} -> false
+  end
+end
```
PragTob: I know I'm more hardcore with this than "normal" people, but I think I'd prefer to have `port_free?/1` as part of `Env` or another helper module. It's inherently the same between `runtime` and `dev` (if I change it in one, I'm pretty sure I wanna change it in the other) and so I think should have one source of truth :) 

edit: same for `test.exs`

editedit: I reminded myself that we can't really use libcode in the normal envs (dev/test/prod) - so might still be useful but may not be _as_ important.

config/runtime.exs
```diff
@@ -49,8 +49,45 @@ gen_rpc_socket_recbuf = Env.get_integer("GEN_RPC_SOCKET_RECEIVE_BUFFER")
...
+
+    {:dev, port} ->
+      if port_free?.(port) do
+        port
+      else
+        raise "GEN_RPC_TCP_SERVER_PORT=#{port} is set but it's already in use."
+      end
+
+    {_env, port} ->
+      port || 5369
```
PragTob: might be nice to use a variable for `5369` (`@default_port` or whatever it is)

config/test.exs
```diff
@@ -17,20 +62,31 @@ for repo <- [
...
     hostname: "127.0.0.1",
+    port: db_port,
     pool: Ecto.Adapters.SQL.Sandbox
 end
 
-http_port = if partition, do: 4002 + String.to_integer(partition), else: 4002
+# 16 ports each, one per peer in TestEnv's slots, in bands of their own so no run's peers can
+# land on another run's endpoint or gen_rpc port.
+peer_http_base = get_integer.("TEST_PEER_PORT_BASE", nil) || 10_000 + offset * 16
+peer_gen_rpc_base = get_integer.("TEST_PEER_GEN_RPC_PORT_BASE", nil) || 26_000 + offset * 16
```
PragTob: 16 to me sounds more like it could/should be some constant 🤔  Might have missed what the peers are but it also sounds like a value that _may_ change

lib/realtime/gen_rpc.ex
```diff
@@ -208,6 +208,28 @@ defmodule Realtime.GenRpc do
...
+
+  gen_rpc calls this once per node while opening a client connection (`gen_rpc_client.erl:252`),
+  and expects `{driver, port}` or `{:error, reason}`.
+  """
+  @spec get_config(node) :: {:tcp | :ssl, :inet.port_number()} | {:error, term()}
+  def get_config(node) do
+    driver = Application.get_env(:gen_rpc, :default_client_driver, :tcp)
+    key = if driver == :ssl, do: :ssl_server_port, else: :tcp_server_port
+
+    case :erpc.call(node, :application, :get_env, [:gen_rpc, key], :timer.seconds(5)) do
```
PragTob: nit: [`to_timeout/1`](https://elixir.hexdocs.pm/Kernel.html#to_timeout/1)  is better interface than `:timer.fun` imo :) (`to_timeout(second: 5)`)

lib/realtime/gen_rpc.ex
```diff
@@ -208,6 +208,28 @@ defmodule Realtime.GenRpc do
     end)
   end
 
+  @doc """
+  Tells gen_rpc which port to dial `node` on, by asking that node over Erlang distribution.
+
+  Only `config/runtime.exs` in dev wires this in, as
+  `config :gen_rpc, client_config_per_node: {:external, __MODULE__}`: several nodes on one machine
+  cannot share a port, so each takes a free one. Everywhere else every node listens on the same
+  port and gen_rpc's own default answers for them.
```
PragTob, line 217: good/important docs 👌

test/support/test_tenant_db/backend/docker.ex
```diff
@@ -101,42 +101,94 @@ defmodule TestTenantDb.Backend.Docker do
...
     :crypto.strong_rand_bytes(length)
     |> Base.url_encode64()
     |> binary_part(0, length)
   end
 
+  # This run's containers carry its run tag, so a run only ever tears down its own.
+  def container_prefix do
+    case TestEnv.run_tag() do
+      "" -> @container_prefix
+      "_" <> tag -> "#{@container_prefix}-#{tag}"
```
PragTob: nit: but if `run_tag` returned "-tag" (or we didn't care about underscore vs. dash) we wouldn't need branches here and just do "#{@container_prefix}#{run_tag}" (or  `<>` whatever)

test/support/test_tenant_db/backend/docker_test.exs
```diff
@@ -0,0 +1,69 @@
+defmodule TestTenantDb.Backend.DockerTest do
+  use ExUnit.Case, async: false
```
PragTob, line 4: a comment on why we `async: false` may be nice :) (I suspect it's because we're interacting witht he global docker, but still may be nice)

edit: Oh and `env` stuff

test/support/clustered.ex
```diff
@@ -158,18 +163,33 @@ defmodule Clustered do
...
+    Another test run or a dev server is bound to it. Set #{env_var} to move this run to a free block.
+    """
+  end
+
+  defp await_port_available!(port, label, env_var, attempts) do
+    if TestEnv.port_available?(port) do
+      :ok
+    else
+      Process.sleep(@port_wait_delay_ms)
+      await_port_available!(port, label, env_var, attempts - 1)
```
PragTob: Isn't this essentially `eventually` with a customizable error message? :)

test/support/test_env.ex
```diff
@@ -0,0 +1,94 @@
+defmodule TestEnv do
```
PragTob, line 1: it's arguably taste, but I think I'd prefix `Realtime` or `RealtimeTest` to the module name - just to namespace and avoid conflicts (with a potential future `test_env` package).

test/support/test_env.ex
```diff
@@ -0,0 +1,94 @@
...
+  @moduledoc false
+  # The ports and node names this run owns. config/test.exs claims the ports; names carry its
+  # run tag, so concurrent runs never collide.
+
+  # One slot per peer, offset from both peer port ranges, so no test has to know the
```
PragTob: are those the peers we earlier talked about? I think so but iirc there we said there are 16 but here there are 12 🤔

test/support/test_env.ex
```diff
@@ -0,0 +1,94 @@
...
+  # reuseaddr the real listeners use, so a port left in TIME_WAIT by a peer that just
+  # stopped reads as available while a live listener — ours or another run's — does not.
+  # Inherently a snapshot: use it to report a collision, never to allocate a port.
+  @spec port_available?(:inet.port_number()) :: boolean()
+  def port_available?(port) do
+    case :gen_tcp.listen(port, [:inet, ip: {0, 0, 0, 0}, reuseaddr: true, active: false]) do
+      {:ok, socket} -> :gen_tcp.close(socket) == :ok
+      {:error, _reason} -> false
+    end
+  end
```
PragTob: another implementation 👀

PragTob (approved):
👋 

Looks good to me, but I'm also not that familiar with our whole setup tenants and all - so may be good to have someone else review this as well.

But overall, looks good to me and I admire the effort of enabling is to run a lot of stuff in parallel 🚀  A+ on that alone!

I ran 2 nodes/config locally while running tests and that all worked great ✅ 

The comments are are all optional and broadly:

* minor: centralizing `port_free?` somewhere - I'm a bit hardcore and like a single source of knowledge if it's inherently the same, but yeah with the config files it's a bit harder :(
* nit: some constants/speaking variables for some numbers may be nice (and the `slots` had me a bit confused, also 16 vs. 12 🤷  but I may just be missudnerstanding)
* nit: fanboying `to_timeout/1`
* nit: `async: false` comments
* minor: name spacing for test support modules maybe
* minor: I think one of the helpers could be `eventually` with a custom error emssage

But well, all looks good from here. Would also appreciate having this in, as it changes a lot of how tenants etc. are run and their DBs in test which is one of my highest suspects for our current flaky tests and I hate merge conflicts 😅 

Image

config/runtime.exs
leandrocp: This reminds me of https://www.youtube.com/watch?v=Cpx8vigaGW4

leandrocp: I also prefer a single source of truth but that would be possible only if we move all the code from test.exs and dev.exs to runtime.exs so I'm not sure it's worth it just because of a couple small functions 🤔

config/runtime.exs
leandrocp: Moved to `gen_rpc_default_port` 👍🏻

config/test.exs
leandrocp: Peers are the nodes started by `Clustered` and they simulate different scenarios like a bad tcp port, regions, etc, you can see the mapping in `TestEnv` and they may change but not very often.

> could/should be some constant

Moved to `peer_ports_per_run` 👍🏻

lib/realtime/gen_rpc.ex
leandrocp: I've been using `:timer` for a very long time so it's just habit. No preference really 😄

test/support/test_env.ex
leandrocp: Usually I do prefix by app name as well but most of realtime test support modules do _not_ prefix `Realtime.`. Tbh I don't know the reason and I just followed the established convention. I think we can do a follow-up PR renaming such modules if everyone agrees with that pattern. Wdyt?

test/support/test_env.ex
leandrocp: Just to leave some room, I added more comments to make it more clear.

config/runtime.exs
PragTob: yeah yeah absolutely, that's what my edit edit was supposed to mean, I kinda forgot we couldn't. It's not worth it. But I think there's still `runtime.exs` and the other usage in a test that could be made the same 🤔 

edit: the other one is the `TestEnv` module

lib/realtime/gen_rpc.ex
PragTob: for me that was just a moment at... I think goatmire where Andrea low key made fun of Sasa for still using `:timer` - so since then I changed as it's also a funny memory :)

PragTob: (plus solving issues with data vs. functions which I admire but besides the point)

test/support/test_env.ex
PragTob: Absolutely then that's the right call and let's do it in a separate PR :)

test/support/clustered.ex
leandrocp: Good point! Fixed

config/runtime.exs
leandrocp: Made some small changes to reuse more code 👍🏻

PR #2111, opened 2026-08-19
fix: Circumvent phoenix join_ref regression

PragTob:
Avoid the regression from phoenix 1.8.3 that drops incoming messages with stale or missing join refs.

Verify against a release by asking Guillaume to run the affected tests. Also verify locally.

Decided to upgrade to 1.8.11 but undo the bugfix/regression so that we can fix the CVEs while we're at it, `mix deps.get` complained. The change seemed minimal enough: supabase/phoenix#52


leandrocp (approved):

PragTob:
```
 1) test delete tenant deletes chosen tenant (RealtimeWeb.TenantControllerTest)
Error:      test/realtime_web/controllers/tenant_controller_test.exs:306
     Expected truthy, got false
     code: assert Realtime.Tenants.ReplicationConnection.ready?(tenant.external_id)
     arguments:

         # 1
         "emdrk4fxuiihrqk7ajd6osehvxsvbxv2"

     stacktrace:
       test/realtime_web/controllers/tenant_controller_test.exs:310: (test)
```

😬  (rerunning)

PR #2116, opened 2026-08-19
feat: encryption on upsert

filipecabaco:

Encrypt with new cypher on upsert.


lib/realtime/api/tenant.ex
```diff
@@ -46,8 +46,8 @@ defmodule Realtime.Api.Tenant do
     timestamps()
   end
 
-  @doc false
-  def changeset(tenant, attrs) do
+  @doc "`opts` are forwarded to `Realtime.Crypto.encrypt!/2`."
+  def changeset(tenant, attrs, opts \\ []) do
```
PragTob: feels less like an `@doc` and more like a comment to me :) (aka it's a "why do we do this" - doesn't describe the whole fucntion).

lib/realtime/api/tenant.ex
```diff
@@ -112,11 +113,59 @@ defmodule Realtime.Api.Tenant do
...
+  def encrypt_jwt_secret(%Ecto.Changeset{valid?: true, changes: %{jwt_secret: plaintext}} = changeset, opts)
       when is_binary(plaintext),
-      do: put_change(changeset, :jwt_secret, Crypto.encrypt!(plaintext))
+      do: put_change(changeset, :jwt_secret, Crypto.encrypt!(plaintext, opts))
+
+  def encrypt_jwt_secret(changeset, _opts), do: changeset
+
+  @doc """
+  Keeps `gcm_migrated_at` in step with the ciphers the pending write leaves behind: stamped once
+  every encrypted value is AES-256-GCM, cleared again as soon as one goes back to the legacy cipher.
```
PragTob: omega minor: José once taught me that `@doc` is at best used with a one line summary in the first line and then the body (kind of like a commit message) and that the tools are optimized for that style :)  So first line, what is this function here for overall, body has all the details.

lib/realtime/tenants/encryption.ex
```diff
@@ -0,0 +1,33 @@
...
+  Per-tenant rollout of the move from AES-128-ECB to AES-256-GCM.
+
+  Both ciphers stay readable throughout: `Realtime.Crypto.decrypt!/1` picks one from the value
+  itself, so a tenant can be moved either way without a deploy.
+
+  Temporary: deleted along with `Realtime.Tenants.EncryptionReconciler` and the legacy branch in
+  `Realtime.Crypto` once no tenant is left on ECB.
+  """
+
+  # TODO: delete this module once no tenant is left on AES-128-ECB.
```
PragTob, line 12: minor: I always like to put this into a ticket in the backlog and reference the ticket here.

more minor: One of the side projects I wanna build, is something that notifies you about these :)

lib/realtime/tenants/encryption.ex
```diff
@@ -0,0 +1,33 @@
...
+  @doc """
+  Cipher a tenant's values should be written with: GCM only while `:db_enc_write_gcm` is set and the
+  #{@flag} flag covers the tenant, the legacy cipher otherwise. Turning either off is enough to put
+  the tenant back on ECB from its next write onwards.
+
+  Reads the tenant cache, so it must not be called from inside that cache's fallback - see
+  `Realtime.Tenants.EncryptionReconciler.reconcile/1`.
+  """
+  @spec cipher_for(binary() | nil) :: Crypto.cipher()
+  def cipher_for(external_id) when is_binary(external_id) do
```
PragTob, line 28: really like extracting this as a central source of truth like you did, well done 👌

test/realtime/api/tenant_test.exs
```diff
@@ -3,15 +3,33 @@ defmodule Realtime.Api.TenantTest do
...
+    "db_host" => "127.0.0.1",
+    "db_port" => "5432",
+    "db_name" => "postgres",
+    "db_user" => "postgres",
+    "db_password" => "postgres"
+  }
+
+  describe "encrypt_jwt_secret/2 via changeset/3" do
+    test "encrypts jwt_secret with the cipher the caller asked for" do
+      for cipher <- [:gcm, :ecb] do
```
PragTob: minor: I'd make that the outer loop (around `test`) and have it be 2 different tests interpolationg `cipher` into the test description :)

test/realtime/api/tenant_test.exs
```diff
@@ -35,6 +53,84 @@ defmodule Realtime.Api.TenantTest do
...
+
+      changeset = Tenant.changeset(%Tenant{extensions: []}, attrs, cipher: :gcm)
+
+      assert {:ok, %DateTime{}} = stamp(changeset)
+    end
+
+    test "does not stamp while jwt_secret is on the legacy cipher" do
+      changeset = tenant(jwt_secret: cipher_text(:ecb), extensions: [extension(:gcm)]) |> change() |> mark()
+
+      assert stamp(changeset) == :error
```
PragTob: Could we change it to an error tuple? Just getting back `:error` and seeing it somewhere in an error will be harder to track back vs. `{:error, :legacy_cypher}` or something 🤔

test/realtime/api/tenant_test.exs
```diff
@@ -35,6 +53,84 @@ defmodule Realtime.Api.TenantTest do
...
+      changeset = %Tenant{external_id: "tenant", jwt_secret: cipher_text(:gcm)} |> change() |> mark()
+
+      assert stamp(changeset) == :error
+    end
+
+    test "leaves an already stamped tenant alone" do
+      changeset =
+        tenant(jwt_secret: cipher_text(:gcm), extensions: [extension(:gcm)]) |> stamped() |> change() |> mark()
+
+      assert stamp(changeset) == :error
```
PragTob: should this be `:error`? I.e. I like the idempotency mindset of "this is already in the state that I'm supposed to bring it to so I no-op since the intent of my call was already achieved`

PragTob: Ah I just notice that this is due to the implementation of `stamp/1` being a `Map.fetch` - to me that's a bit surprising reading the tests as it just means we didn't add any change. 

Maybe [`changed?/3`](https://ecto.hexdocs.pm/Ecto.Changeset.html#changed?/3) is what we're looking for here to see if a change was made or not?

test/realtime/api/tenant_test.exs
```diff
@@ -72,4 +168,27 @@ defmodule Realtime.Api.TenantTest do
...
+
+  defp encrypted_settings(cipher) do
+    for key <- Extensions.encrypted_settings_keys("postgres_cdc_rls"), reduce: @settings do
+      acc -> Map.put(acc, key, Crypto.encrypt!(@settings[key], cipher: cipher))
+    end
+  end
+
+  defp cipher_text(cipher), do: Crypto.encrypt!("a-secret", cipher: cipher)
+
+  defp stamped(tenant), do: %{tenant | gcm_migrated_at: DateTime.utc_now(:second) |> DateTime.add(-1, :day)}
```
PragTob: `DateTime.shift/2` is preferred these days: https://elixir.hexdocs.pm/DateTime.html#add/4

(I believe I added a check to credo for this even 😅 )

PragTob (approved):
👋 

Looks really good, thank you!

Most of them are minor comments. 

* good job on the extractions 👍 
* there are some musings on the format of `@doc` comments which are naturally ingorable
* some thoughts on generating tests instead of looping inside a test 😇 
* the way `stamp` worked in the tests surprised me/wonder if we can make that more obvious
* `DateTime.shift/2` nitpick

But nothing major found, thanks for your work! 🙌 

Image

lib/realtime/api/tenant.ex
leandrocp: I feel the opposite even tho the doc was only documenting a single arg. The "why do we do this" is actually documentation and `opts` is part of that function, thus `@doc` makes more sense to me :)
Besides that comments are hidden from hexdocs, introspection, etc so all tooling lose that context.

lib/realtime/api/tenant.ex
leandrocp: I think we should [redact](https://ecto.hexdocs.pm/Ecto.Schema.html#module-redacting-fields) this field to avoid exposing it in logs.


```suggestion
    field(:jwt_secret, :string, redact: true)
```

leandrocp: Same here:


```suggestion
    field(:jwt_jwks, :map, redact: true)
```

leandrocp (approved):
LGTM, just a non-blocking suggestion to redact jwt fields.

PR #2117, opened 2026-08-20
feat: muster graceful shutdown

edgurgel:

Make the Muster scope leave before the node shuts down to avoid receiving broadcasts messages where the shutting down node is a router. 

The Muster scope notifies all other peers that it's leaving. Once every node acked it waits `settle_ms` for any inflight broadcast messages or RPCs that were in-flight and then it considers drained to continue the termination.

### DIFF
--- a/forum/README.md
+++ b/forum/README.md
@@ -78,8 +78,9 @@ The split keeps the hot path (joins/claims) sharded while the rare, node-wide wo
 ### Public API
 
 ```elixir
-Forum.Muster.join(scope, group, pid)        # :ok | {:error, :rpc_failed | :not_local | ...}
+Forum.Muster.join(scope, group, pid)        # :ok | {:error, :rpc_failed | :not_local | :draining | ...}
 Forum.Muster.leave(scope, group, pid)       # :ok | {:error, term}
+Forum.Muster.drain(scope, opts)             # graceful cluster-leave; :ok | {:timeout, [node]}
 Forum.Muster.router(scope, group)           # {:ok, node} | {:rebalancing, [node]}
 Forum.Muster.targets(scope, group, sender_view_hash)  # (on router) {:ok, [node]} | {:error, :flood}
 Forum.Muster.occupancy(scope, group)        # (on router) raw held [node], NOT barrier-gated
@@ -242,6 +243,17 @@ Because `:view_hash` is content-derived (`phash2` of the sorted node set) rather
 
 Because the ring is consistent-hashed, a 100-node rolling deploy that grows the new cluster 1 → 100 moves ~1 group out of every 100 candidates per node-add, on average. A `phash2(group, length(members))` scheme would instead remap ~all groups on every node-add. Consistent hashing keeps the per-event announce-set small and the rebalance window short, which in turn keeps the broadcast-fan-out window short.
 
+### Graceful shutdown / draining (`Forum.Muster.drain/2`)
+
+**Draining is outbound-only.** While leaving, the node suppresses every *outbound self-assertion* (heartbeat announce, re-discovery, discover-ack, nodeup-discovery) so no peer can re-pair it after it announced leaving. Every *inbound coordination RPC* (`:occupied`, `:vacant_batch`, snapshot/delta/transition applies, `:rebalance_marker`) keeps being serviced through the whole drain + settle window — the node stays a correct, responsive router right up to exit; it just stops advertising itself. `join/3` on a draining node returns `{:error, :draining}` so a stray join fails loudly instead of creating an unroutable member.
+
+* Its **source role** (holding local members) needs no Muster mechanism on shutdown. As the host app disconnects its clients, membership drops to 0 and the normal vacant-flush (see *How leave works*) retracts occupancy on the current router. That is the host app's job.
+* Its **router role** is the one only a *pre-death* handoff can protect. By consistent hashing this node is the elected router for ~1/N of *all* groups cluster-wide. If it dies abruptly, broadcast messages routed to this node will be dropped.
+
+`drain/2` closes that window while the node is **still alive**. It does *not* rebalance itself; instead it broadcasts `{:muster_leaving}` and lets each **peer** rebalance it out of that peer's ring. The instant a peer swaps its ring, its subsequent claims/routing go to the newly-elected router, and its own rebalance re-announces the groups it holds to that router. Nothing the dying router "owned" is lost: every occupancy row was sourced from some node's local members, and the live sources re-announce during the same rebalance. The only rows that vanish are the ones sourced from the dying node itself which are going away regardless.
+
+**`drain` is terminal.** It leaves the node draining with its heartbeat off, so a node that calls `drain` but does not actually die (e.g. an aborted deploy) is stranded until it restarts. Call it only when truly shutting down.
+
 ---
 
 ## Failure scenarios
@@ -280,11 +292,12 @@ Snapshot dispatch is fire-and-forget on the sender (the coordinator does not wai
 
 After a crash the supervisor restarts the **coordinator**, whose `init/1`:
 
-* Resets the member list to `[node()]`, the `:status` persistent_term to `:ready`, and `:view_hash` to `phash2([node()])` (forgetting the cluster view). A sender that still holds the pre-crash multi-node `:view_hash` will mismatch this single-node hash, so `can_decide?/2` returns false and broadcasts to the restarted router fan out to all nodes until it re-converges.
+* Resets the member list to `[node()]`, the `:status` persistent_term to `:converging` (not `:ready`, even at members `[node()]`: a restarted local sender would otherwise agree with that one-node view and trust incomplete occupancy before scope discovery re-pairs; the bounded `:singleton_promotion_timeout_ms` restores `:ready` if the scope truly downsized to one node), and `:view_hash` to `phash2([node()])` (forgetting the cluster view). A sender that still holds the pre-crash multi-node `:view_hash` will mismatch this single-node hash, so `can_decide?/2` returns false and broadcasts to the restarted router fan out to all nodes until it re-converges.
 * Walks the shards' membership entries (which survive a coordinator death because they live in `:public, :named_table` ETS owned by the Supervisor) and re-asserts (monotonic upsert) its router-role **occupancy self-rows** for every group with live local members. The occupancy table is itself owned by the Supervisor, so it is **not** recreated on a coordinator restart: it survives intact under whichever shard processes write it directly. Any foreign-source rows the dead incarnation held are harmless: the restart resets `:view_hash` (below), so `can_decide?/2` is false and callers flood until each source re-snapshots (replacing its rows) and the `:ready` sweep prunes the rest. A coordinator crash also restarts every shard (see [Coordinator crash](#coordinator-shard-or-ring-crash-for-other-reasons)), so this same walk-and-reassert always runs against freshly-restarted shards, one reset story, not two. The per-group state machine is not lost by this: it lives in the shards' Supervisor-owned durable claim-state table, which each shard's own restart rebuilds from; remote sources' rows are refilled when they re-snapshot after re-discovery.
 * Re-broadcasts the discovery message; peers respond; `recompute_members` runs again. This `init` broadcast is one-shot, but it is **not** the only chance to re-pair: because the node restarted in place (its dist connection never dropped, so no `:nodeup` re-fires and peers won't reach back out), a lost `init` discovery would otherwise strand it forever, so the view heartbeat re-offers `:muster_discover` to every connected non-member each tick (see [the re-discovery backstop](#router-readiness-barrier)), bounding worst-case stranding to one interval.
+* Reopens the join gate (`:accepting_joins` persistent_term to `true`). `drain/2` is the *only* writer that closes it (to `false`), and that term outlives the coordinator process. So a coordinator crash *during* a drain — before the node actually dies — would otherwise leave the gate stuck closed forever: the restarted incarnation re-pairs and routes normally, yet `join/3` would keep returning `{:error, :draining}` for the life of the OS process. A fresh `init` means "not leaving", so it reopens the gate; the in-memory `leaving` state is dropped by the restart anyway, so the drain is effectively aborted and the node rejoins as a normal member.
 
-The supervisor restart strategy ensures the cluster eventually re-converges. During the restart window, between the crash and `init/1` running, the `:status` persistent_term is left at `:rebalancing`, so `Muster.router/2` returns `{:rebalancing, members}` and callers correctly fan out. (See `test/forum/muster_distributed_test.exs`.)
+The supervisor restart strategy ensures the cluster eventually re-converges. The `:status` and `:view_hash` persistent_terms both **survive the crash unchanged** — nothing resets them as the coordinator dies — so during the brief window between the crash and `init/1` running they hold their pre-crash values (whatever `:status` was: usually `:ready` or `:converging`, `:rebalancing` only if the node happened to crash mid-rebalance). That window is nonetheless safe, because the occupancy table is likewise unmutated (the coordinator that would rebalance it is down), so a router answering `targets/3` from it returns the same authoritative set it would have before the crash. What actually protects correctness *across* the restart is `init/1` resetting `:view_hash` to `phash2([node()])` and `:status` to `:converging` (above): from then on `can_decide?/2` fails for any sender still holding the pre-crash multi-node view, so callers flood rather than trust the restarted router's shrunken view until it re-converges. (See `test/forum/muster_distributed_test.exs`.)
 
 ### Coordinator, shard, or ring crash for other reasons
 
--- a/forum/docs/muster-broadcast.md
+++ b/forum/docs/muster-broadcast.md
@@ -129,5 +129,18 @@ sequenceDiagram
 - **Safety during change.** The `:rebalancing` fan-out trades a brief burst of
   extra traffic for never missing a node whose router assignment is mid-flight.
 
+### Graceful shutdown
+
+A router that is about to leave the cluster must not simply vanish: a peer that
+still hashes a group to it would fire an `:occupied` claim (or route a broadcast)
+to a corpse and silently lose it. `Forum.Muster.drain/2` evacuates the router
+role *before* death — peers rebalance the leaver out of their rings and
+re-announce the groups they hold to the newly-elected routers while the leaver is
+still alive to service in-flight RPCs. During the drain the leaver stays a
+correct router (it answers `targets/3` throughout, including a mandatory settle
+window for broadcasts already in flight to it), but stops advertising itself, and
+`join/3` on it returns `{:error, :draining}`. See the *Graceful shutdown /
+draining* section in the top-level [`README.md`](../README.md).
+
 See the top-level [`README.md`](../README.md) for the join, leave/cooldown,
 rebalance, and failure-handling details that keep the occupancy table accurate.
--- a/forum/lib/forum/muster.ex
+++ b/forum/lib/forum/muster.ex
@@ -236,19 +236,97 @@ defmodule Forum.Muster do
   the table against the live member counts, and never forgets an outstanding router
   assertion, so a shard crash cannot orphan a `{group, node}` entry on a (local or
   remote) router.
+
+  Returns `{:error, :draining}` if this node has begun a graceful cluster-leave
+  (`drain/2`): once leaving, the node is being rebalanced out of every peer's
+  ring, so a new first-member join could produce a local member with no occupancy
+  row on any router (silently missing from every broadcast). The join fails loudly
+  instead. The host app should stop accepting new joins at SIGTERM anyway (see
+  `drain/2`); this is defense-in-depth for a stray join that races the drain.
   """
-  @spec join(atom, group, pid) :: :ok | {:error, :not_local | :rpc_failed | term}
+  @spec join(atom, group, pid) :: :ok | {:error, :not_local | :draining | :rpc_failed | term}
   def join(_scope, _group, pid) when is_pid(pid) and node(pid) != node(),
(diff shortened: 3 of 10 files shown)


forum/docs/muster-broadcast.md
```diff
@@ -129,5 +129,18 @@ sequenceDiagram
...
+A router that is about to leave the cluster must not simply vanish: a peer that
+still hashes a group to it would fire an `:occupied` claim (or route a broadcast)
+to a corpse and silently lose it. `Forum.Muster.drain/2` evacuates the router
+role *before* death — peers rebalance the leaver out of their rings and
+re-announce the groups they hold to the newly-elected routers while the leaver is
+still alive to service in-flight RPCs. During the drain the leaver stays a
+correct router (it answers `targets/3` throughout, including a mandatory settle
+window for broadcasts already in flight to it), but stops advertising itself, and
+`join/3` on it returns `{:error, :draining}`. See the *Graceful shutdown /
+draining* section in the top-level [`README.md`](../README.md).
```
PragTob: thanks for being thoughtful about adding these docs 👌

forum/lib/forum/muster/scope.ex
```diff
@@ -728,6 +750,58 @@ defmodule Forum.Muster.Scope do
...
 
+  # Graceful cluster-leave: evacuate the ROUTER role before this node dies (see
+  # Forum.Muster.drain/2). We do NOT rebalance ourselves; instead we broadcast
+  # {:muster_leaving} and let each PEER rebalance us out of its ring and ack. The
+  # drain/2 caller is parked in `from` and replied to asynchronously (via
+  # GenServer.reply from the :leave_settle / :leave_deadline handlers), so this
+  # loop never blocks on the remote acks -- preserving the "Scope's loop never
+  # blocks on a remote RPC" invariant.
+  #
+  # Start draining immediately (leaving: true suppresses every outbound
```
PragTob: minor: using code ticks aka `leaving: true` would help readability ever so slightly (similar for other code snippets in the comment, but this one was a bit harder to parse as it was less clear that this was code)

forum/lib/forum/muster/scope.ex
```diff
@@ -728,6 +750,58 @@ defmodule Forum.Muster.Scope do
...
+  # self-assertion) and stop accepting joins. A singleton (no peers) has nobody
+  # to evacuate to and nothing routed to it, so it replies :ok at once. Otherwise
+  # we arm the ack deadline and wait; inbound coordination RPCs keep being
+  # serviced throughout.
+  def handle_call({:drain, timeout, settle_ms}, from, %State{} = state) do
+    :persistent_term.put({Forum.Muster, state.scope, :accepting_joins}, false)
```
PragTob, line 769: minor: not necessary or everyone knows `:persistent_term` better than me, but the information that `false` is a simple enough term not to trigger GC here _might_ be helpful, _might_ be considered general knowledge (I didn't know before 😅 )

forum/lib/forum/muster/scope.ex
```diff
@@ -728,6 +750,58 @@ defmodule Forum.Muster.Scope do
...
+    expected = state.peers |> Map.keys() |> Enum.map(&node/1) |> MapSet.new()
+
+    tp(:muster_drain_begin, %{
+      scope: state.scope,
+      node: node(),
+      expected: MapSet.to_list(expected),
+      settle_ms: settle_ms
+    })
+
+    if MapSet.size(expected) == 0 do
```
PragTob: While obvs. not super important I was suspicious of `MapSet.size() == 0` but it turned out to be a [good call](https://github.com/PragTob/elixir_playground/blob/main/bench/map_set_empty.exs)

forum/lib/forum/muster/scope.ex
```diff
@@ -728,6 +750,58 @@ defmodule Forum.Muster.Scope do
...
+      Logger.info(
+        "Muster[#{node()}|#{state.scope}] Draining: singleton (no peers), nothing to hand off; draining immediately"
+      )
+
+      {:reply, :ok, %{state | leaving: true}}
+    else
+      Logger.info(
+        "Muster[#{node()}|#{state.scope}] Draining: broadcasting leave, awaiting acks from " <>
+          "#{MapSet.size(expected)} peer(s) #{inspect(MapSet.to_list(expected))} " <>
```
PragTob: very minor: I'd put `MapSet.size(expected)` into a variable to not calculate it twice. It has practically no perf impact, but reading code when I see the same value calculated I always think I'm missing something 😅

forum/lib/forum/muster/scope.ex
```diff
@@ -890,76 +988,109 @@ defmodule Forum.Muster.Scope do
...
+  # peer's eventual real :DOWN is a no-op (the guard above rejects it).
+  def handle_info({:muster_leaving, peer_pid, _seq}, %State{} = state) do
+    case Map.get(state.peers, peer_pid) do
+      nil ->
+        {:noreply, state}
+
+      ref ->
         Logger.info(
-          "Muster[#{node()}|#{state.scope}] peer down: #{inspect(peer_node)}, dropping occupancy/view data attributable to this incarnation and rebalancing"
+          "Muster[#{node()}|#{state.scope}] peer draining: #{inspect(node(peer_pid))}, rebalancing it out of the ring and acking"
```
PragTob: very minor inconsistency: further up you use just `#{node()}` (no `inspect` call) - may be nice to have them all conform.

forum/test/forum/muster_distributed_test.exs
```diff
@@ -5317,6 +5326,274 @@ defmodule Forum.MusterDistributedTest do
...
+          # source row intact -- no missed broadcast across the handoff.
+          {:ok, r} = Muster.router(scope, g)
+          assert r != c_node
+          assert a_node in occupancy_on(r, scope, g)
+
+          # The eventual real :DOWN (process actually stops) is a no-op: C was
+          # already departed + demonitored, so membership stays put.
+          :ok = stop_supervised({:peer, c_name})
+          Process.sleep(300)
+          assert Muster.members(scope) == Enum.sort([t_node, a_node])
```
PragTob: Is `members` always sorted? If we don't have it yet (we should!) I'd prefer an assertion that compares lists independent of order (or use Devon's assertions library) or use MapSet or something.

(introducing library probably for a separate PR)

forum/test/forum/muster_distributed_test.exs
```diff
@@ -5317,6 +5326,274 @@ defmodule Forum.MusterDistributedTest do
...
+          # The group C routed is reachable on the newly elected router, with A's
+          # source row intact -- no missed broadcast across the handoff.
+          {:ok, r} = Muster.router(scope, g)
+          assert r != c_node
+          assert a_node in occupancy_on(r, scope, g)
+
+          # The eventual real :DOWN (process actually stops) is a no-op: C was
+          # already departed + demonitored, so membership stays put.
+          :ok = stop_supervised({:peer, c_name})
+          Process.sleep(300)
```
PragTob: not a fan of `Process.sleep/1` in tests (can wait too long so slows down tests, or produce flakies), usually it's avoidable by waiting for a message or busy waiting for some condition. I don't know this half-way well enough to say which is which and if it's possible but _maybe_ an `eventually fn -> not Process.alive?(..) end` could be used here?

forum/test/forum/muster_distributed_test.exs
```diff
@@ -5317,6 +5326,274 @@ defmodule Forum.MusterDistributedTest do
...
+          assert Muster.members(scope) == Enum.sort([t_node, a_node])
+          assert status(scope) == :ready
+        end,
+        fn _trace -> :ok end
```
PragTob, line 5838: Why are we running `check_trace` if we ignore the trace in the end? 🤔

test/integration/muster_drain_shutdown_test.exs
```diff
@@ -0,0 +1,51 @@
+defmodule Realtime.Integration.MusterDrainShutdownTest do
+  use ExUnit.Case, async: false
```
PragTob, line 3: a comment for why `async: false` would be appreciated :)

test/integration/muster_drain_shutdown_test.exs
```diff
@@ -0,0 +1,51 @@
...
+
+  alias Forum.Muster
+
+  @moduletag timeout: 60_000
```
PragTob: pretty sure 60 seconds is the default timeout, no need to set it imo

test/integration/muster_drain_shutdown_test.exs
```diff
@@ -0,0 +1,51 @@
...
+  defp await(fun, attempts \\ 100)
+  defp await(_fun, 0), do: flunk("condition never became true within timeout")
+
+  defp await(fun, attempts) do
+    if fun.() do
+      :ok
+    else
+      Process.sleep(100)
+      await(fun, attempts - 1)
+    end
```
PragTob: we do have an `eventually` helper for this: https://github.com/supabase/realtime/blob/94246b530f6cd838abca86a5e7368bd6c230b4e6/test/support/test_helpers.ex#L16-L32

PragTob (commented):
👋  Looks real good, although I wouldn't trust myself to understand all of it hence Comment vs. Approve as I think someone who understands Muster better than I do should probably give this one a read. Maybe we can cache in another José/dashbit review token 😓 

Really want to compliment you on the amount of helpful documentation you added in this PR, it's really helpful and thoughtful 💚 💡 

This may be the first proper review I'm giving one of your PRs, so my usual spiel: I may leave a lot of comments on a PR, that doesn't mean I don't like it - often quite the opposite! I think this is a fantastic PR for instance. I just like to ask questions for my understanding, praise something (like the docs!), or also ask questions to understand decisions and so we can align & learn. I try to make it clear in each one of my comments (and in a summary) how important they are and I also default to approving PRs as long as I think they make the code base better (and we shouldn't block that). I trust you as the author to make the calls, and so not implementing my minor suggestions is perfectly fine and reasonable. 10 comments, an approval and no changes really needed is almost a normal PR review from me (I would appreciate answers to my questions to learn though 😁 ).  And of course, separate follow ups to not hold this up are also great imo. Oh and while it's harder for the 2 of us, I'm also always available for a "pairing review" to make turnaround on comments quicker.

So this one:
* some praise 🚀 

things I'd like to see (or understand) but not blocking:
* comment on `async: false`
* why we only sort one half of the comparison
* why we run `check_trace` without checking the trace 🤔 
* getting rid off `Process.sleep` if possible (or an easy follow up)
* using our `eventually` helper instead of `await` would be really nice unless I'm missing something

minor optional things:
* minor comments around using code ticks in the comments for better readability (very optional)
* my benchmark brain kicking in but seeing you have the overall fastest variant 🚀 
* some minor things on reusing variables or consistent representation in log messages


Image

forum/lib/forum/muster/scope.ex
edgurgel: Good point! I will add a comment 🫡

forum/lib/forum/muster/scope.ex
edgurgel: Fair enough!

forum/lib/forum/muster/scope.ex
edgurgel: yeah I think it's safe to drop inspect tbh. I usually add inspect everywhere to be safe but `node` is always an atom 🤷

forum/test/forum/muster_distributed_test.exs
edgurgel: Hmm I might try using `monitor_nodes` as this is killing a peer/node 🤔

forum/test/forum/muster_distributed_test.exs
edgurgel: Oh it's mostly because there is no other function to just "run a trace" under a snabbkaffe trace. Which is a good point to have one when there is no need to check anything 🤔 I might add this to the library later

forum/README.md
```diff
@@ -78,8 +78,9 @@ The split keeps the hot path (joins/claims) sharded while the rare, node-wide wo
 ### Public API
 
 ```elixir
-Forum.Muster.join(scope, group, pid)        # :ok | {:error, :rpc_failed | :not_local | ...}
+Forum.Muster.join(scope, group, pid)        # :ok | {:error, :rpc_failed | :not_local | :draining | ...}
```
leandrocp: Nit: the spec is `@spec join(atom, group, pid) :: :ok | {:error, :not_local | :draining | :rpc_failed | term}` so this could be:


```suggestion
Forum.Muster.join(scope, group, pid)        # :ok | {:error, :not_local | :draining | :rpc_failed | term}
```

forum/README.md
```diff
@@ -242,6 +243,17 @@ Because `:view_hash` is content-derived (`phash2` of the sorted node set) rather
 
 Because the ring is consistent-hashed, a 100-node rolling deploy that grows the new cluster 1 → 100 moves ~1 group out of every 100 candidates per node-add, on average. A `phash2(group, length(members))` scheme would instead remap ~all groups on every node-add. Consistent hashing keeps the per-event announce-set small and the rebalance window short, which in turn keeps the broadcast-fan-out window short.
 
+### Graceful shutdown / draining (`Forum.Muster.drain/2`)
+
+**Draining is outbound-only.** While leaving, the node suppresses every *outbound self-assertion* (heartbeat announce, re-discovery, discover-ack, nodeup-discovery) so no peer can re-pair it after it announced leaving. Every *inbound coordination RPC* (`:occupied`, `:vacant_batch`, snapshot/delta/transition applies, `:rebalance_marker`) keeps being serviced through the whole drain + settle window — the node stays a correct, responsive router right up to exit; it just stops advertising itself. `join/3` on a draining node returns `{:error, :draining}` so a stray join fails loudly instead of creating an unroutable member.
```
leandrocp: > `join/3` on a draining node returns `{:error, :draining}`

Just a question for my own understanding. In that case who is responsible for retrying until it finds the correct router?

forum/README.md
```diff
@@ -242,6 +243,17 @@ Because `:view_hash` is content-derived (`phash2` of the sorted node set) rather
...
+
+* Its **source role** (holding local members) needs no Muster mechanism on shutdown. As the host app disconnects its clients, membership drops to 0 and the normal vacant-flush (see *How leave works*) retracts occupancy on the current router. That is the host app's job.
+* Its **router role** is the one only a *pre-death* handoff can protect. By consistent hashing this node is the elected router for ~1/N of *all* groups cluster-wide. If it dies abruptly, broadcast messages routed to this node will be dropped.
+
+`drain/2` closes that window while the node is **still alive**. It does *not* rebalance itself; instead it broadcasts `{:muster_leaving}` and lets each **peer** rebalance it out of that peer's ring. The instant a peer swaps its ring, its subsequent claims/routing go to the newly-elected router, and its own rebalance re-announces the groups it holds to that router. Nothing the dying router "owned" is lost: every occupancy row was sourced from some node's local members, and the live sources re-announce during the same rebalance. The only rows that vanish are the ones sourced from the dying node itself which are going away regardless.
```
leandrocp: Nit: I read this first and found it odd to have a single value tuple like `{:muster_leaving}` but the actual code is `{:muster_leaving, ...}` so maybe something like "broadcasts a `:muster_leaving` message [...]"

edgurgel:
I've addressed the PR feedback and made a fix to ensure that if a peer dies mid-drain we progress draining

PragTob (approved):
Pinged others to take a look but approving so Eduardo isn't blocked as to the best of my ability this looks good - I just don't know muster as well.
IMG_20191011_074648

test/integration/muster_drain_shutdown_test.exs
```diff
@@ -0,0 +1,42 @@
...
+
+    # Terminate the drainer child on the peer: this runs Realtime.MusterDrainer's
+    # terminate/2 -> Forum.Muster.drain/2 while the peer's coordinator is alive.
+    assert :ok = :erpc.call(peer, Supervisor, :terminate_child, [Realtime.Supervisor, Realtime.MusterDrainer], 30_000)
+
+    # drain/2 stopped the peer from accepting new joins.
+    assert :erpc.call(peer, :persistent_term, :get, [{Forum.Muster, scope, :accepting_joins}, true]) == false
+
+    # The survivor rebalanced the peer out of its ring...
+    eventually(fn -> Enum.sort(Muster.members(scope)) == [local] end)
```
leandrocp: `eventually` only returns a boolean, it does not assert. So you need to explicitly assert or reject the returned value:

```suggestion
    assert eventually(fn ->
      Enum.sort(Muster.members(scope)) == Enum.sort([local, peer]) and Muster.status(scope) == :ready
    end)

    # Terminate the drainer child on the peer: this runs Realtime.MusterDrainer's
    # terminate/2 -> Forum.Muster.drain/2 while the peer's coordinator is alive.
    assert :ok = :erpc.call(peer, Supervisor, :terminate_child, [Realtime.Supervisor, Realtime.MusterDrainer], 30_000)

    # drain/2 stopped the peer from accepting new joins.
    assert :erpc.call(peer, :persistent_term, :get, [{Forum.Muster, scope, :accepting_joins}, true]) == false

    # The survivor rebalanced the peer out of its ring...
    assert eventually(fn -> Enum.sort(Muster.members(scope)) == [local] end)
```

forum/lib/forum/muster.ex
```diff
@@ -251,6 +262,76 @@ defmodule Forum.Muster do
...
+    4. Call `drain/2`.
+    5. **Only then** halt.
+
+  **`drain` is terminal**: it leaves the node draining with its heartbeat off,
+  so a node that calls `drain` but does not actually die (e.g. an aborted deploy)
+  is stranded until it restarts. Call it only when truly shutting down.
+
+  ## Options
+
+    * `:timeout_ms` (default `5_000`): how long to wait for every peer to ack
```
leandrocp, line 320: Just a appreciation note for using the `_ms` suffix. :feels-good:

forum/test/forum/muster_distributed_test.exs
```diff
@@ -5317,6 +5330,324 @@ defmodule Forum.MusterDistributedTest do
...
+    # stay out of T's membership. C keeps its own full view. Ot never rebalances
+    # itself.
+    test "a drained node is not resurrected by discovery", %{scope: scope} do
+      t_node = node()
+      c_name = ~c"muster_drain_lame_#{System.unique_integer([:positive])}"
+      c_node = :"#{c_name}@127.0.0.1"
+
+      {:ok, pc, ^c_node} = Peer.start(name: c_name, aux_mod: @aux_mod)
+      start_remote_muster(pc, scope)
+      wait_until(fn -> Enum.sort(Muster.members(scope)) == Enum.sort([t_node, c_node]) end)
```
leandrocp: `wait_until` and `eventually` are very similar, should we merge into just one function?

forum/test/forum/muster_distributed_test.exs
```diff
@@ -5317,6 +5330,324 @@ defmodule Forum.MusterDistributedTest do
...
+
+      assert :ok = :peer.call(pc, MusterPeerAux, :drain, [scope, [settle_ms: 200]])
+      wait_until(fn -> Muster.members(scope) == [t_node] end)
+
+      # T re-offers discovery ~3x over this window; C must ignore every one.
+      Process.sleep(1_000)
+      assert Muster.members(scope) == [t_node]
```
leandrocp: I think we can save some time here:


```suggestion
      assert eventually(fn -> Muster.members(scope) == [t_node] end)
```

leandrocp (changes requested):
Only some small suggestions and a small fix on test assertion. I've not found anything critical 🚀

forum/test/forum/muster_distributed_test.exs
edgurgel: eventually is on `realtime` I think? `wait_on`, `wait_until` are scoped just for forum tests

forum/test/forum/muster_distributed_test.exs
edgurgel: Here we are actually interested in waiting the window for ~3 announcements from T to C and C must not join the cluster.

If you check the lines above we do `wait_until` first, and then just ensure that after 1 seconds (~3 announcements of membership) we haven't changed.

test/integration/muster_drain_shutdown_test.exs
edgurgel: Good catch oops

forum/README.md
edgurgel: Yeah I went back and forth with the "leaving" message and forgot to update the docs! Good catch. Will use the opportunity to add a bit more docs

leandrocp (approved):

PragTob:
`mix format` needs a lil run :)

edgurgel:
Oh yeah I'm also fixing one thing that I found while rewriting the docs! :loading:

lib/realtime/muster_drainer.ex
```diff
@@ -0,0 +1,77 @@
...
+    scope = Keyword.get(opts, :scope) || Application.fetch_env!(:realtime, :muster_scope)
+    drain_opts = Keyword.get(opts, :drain_opts, [])
+    {:ok, %{scope: scope, drain_opts: drain_opts}}
+  end
+
+  @impl true
+  def terminate(_reason, %{scope: scope, drain_opts: drain_opts}) do
+    Logger.info("#{__MODULE__}: draining Muster router role for scope #{inspect(scope)}")
+
+    case Forum.Muster.drain(scope, drain_opts) do
```
leandrocp, line 55: I'm assuming `:already_draining` would not be triggered here so no need to match it?

```elixir
@spec drain(atom, keyword) :: :ok | {:timeout, [node]} | {:error, :already_draining}
```

leandrocp (approved):

lib/realtime/muster_drainer.ex
edgurgel: Yeah if it happens then we are in a real problem and things should just blow up

leandrocp (approved):
🚀

PR #2119, opened 2026-08-20
chore: add automatic dependency updates with Dependabot

PragTob:
Automatic version updates via dependabot to help us stay up to date.

Choice fell on dependabot over renovate as renovate at least seemed unable to deal with the way our phoenix dependency is declared (that said, wasn't able to verify if dependabot _can_).

It's also slightly lower friction to set it up and easy to change, should we want to.

The setup also has a cooldown specified, this works together with our configured hex cool down (as hex applies it to all deps operations, such as manual updates or transitive dependency updates).

Otherwise, also chose to limit the number of open PRs so we're not flooded by PRs all at once and so also keep the release/accumulated changes risk a bit lower.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,45 @@
+# https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference
+
+version: 2
+updates:
+  - package-ecosystem: "mix"
+    directories:
+      - "/"        # root app
+      - "/forum"   # in-repo path dep
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "06:00"
+
+    # Let's not flood ourselves with PRs
+    open-pull-requests-limit: 3
+
+    # May widen `~>` constraints, but only when the latest resolvable version
+    # does not already satisfy them: in-range bumps touch mix.lock only.
+    versioning-strategy: "increase-if-necessary"
+
+    cooldown:
+      # same as in our hex config
+      default-days: 7
+      # major versions are more risky --> more cooldown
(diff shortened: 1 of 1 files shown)


filipecabaco (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,45 @@
...
+      # same as in our hex config
+      default-days: 7
+      # major versions are more risky --> more cooldown
+      semver-major-days: 21
+
+    commit-message:
+      # By default, it changes behavior and so is a fix
+      prefix: "fix"
+      # dev only dependencies don't affect prod
+      prefix-development: "chore"
```
leandrocp, line 31: My only _very minor_ suggestion would be to use (deps) to make it more explicit as:


```suggestion
      prefix: "fix(deps)"
      # dev only dependencies don't affect prod
      prefix-development: "chore(deps)"
```

But I'm not sure if that works with dependabot and don't mind if you prefer to keep the current version :)

leandrocp (approved):
Great!

.github/dependabot.yml
PragTob: It's my understanding that dependabot adds these by itself, but let me double check :)

PragTob: Ah yes, that's what the `include: "scope"` is supposed to accomplish:

> [include](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#include)
> Supports only the value scope
> When defined any prefix is followed by the type of dependencies updated in the commit: deps or deps-dev.

PR #2120, opened 2026-08-20
feat: remove shutdown node from nodes list

filipecabaco:

remove shutdown node from nodes list

### DIFF
--- a/lib/realtime/region_membership.ex
+++ b/lib/realtime/region_membership.ex
@@ -1,0 +1,39 @@
+defmodule Realtime.RegionMembership do
+  @moduledoc """
+  Owns this node's membership of the syn `RegionNodes` group.
+
+  Started last in `Realtime.Application` so this node is only advertised once the
+  rest of the tree can serve the work it will be handed.
+
+  Membership carries a `draining` flag rather than being dropped on shutdown: a
+  node that is going away must stop being *elected* for new work
+  (`Realtime.Nodes.eligible_region_nodes/1`) while peers keep *delivering* to the
+  clients still connected to it (`Realtime.Nodes.region_nodes/1`).
+  """
+  use GenServer
+  require Logger
+
+  def start_link(region), do: GenServer.start_link(__MODULE__, region, name: __MODULE__)
+
+  @doc "Marks this node as draining. No-op when the process is not running."
+  @spec drain() :: :ok
+  def drain do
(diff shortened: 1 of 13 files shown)


lib/realtime/region_membership.ex
```diff
@@ -0,0 +1,30 @@
...
+  def init(region) do
+    Process.flag(:trap_exit, true)
+    :ok = :syn.join(RegionNodes, region, self(), node: node())
+    {:ok, region}
+  end
+
+  @impl true
+  def terminate(_reason, region) do
+    Logger.info("Leaving RegionNodes group #{inspect(region)}")
+    :syn.leave(RegionNodes, region, self())
```
edgurgel: I feel like this is too extreme leaving completely as we might have places (I don't know!) where we implicitly expect to see this information for any connected node while it's alive. 

Should we instead update the metadata to be `state: :shutting_down`" or something like this and in places where we pick a node to start Connect or Postgres Changes we filter them out? WDYT?

lib/realtime_web/controllers/page_controller.ex
```diff
@@ -11,8 +13,9 @@ defmodule RealtimeWeb.PageController do
   end
 
   def healthcheck(conn, _params) do
-    conn
-    |> put_status(:ok)
-    |> text("ok")
+    case SignalHandler.shutdown_in_progress?() do
```
edgurgel, line 16: I think this might bring more trouble than we think as it might show as unhealthy and quickly be removed instead of following the normal process of shutting down a node 🤔

lib/realtime/region_membership.ex
filipecabaco: probably that is a stronger idea, unsure still exploring it

lib/realtime_web/controllers/page_controller.ex
filipecabaco: potentially but this nodes in ECS will be marked for shutdown so it should be ok? I know ECS has a logic like that where the nodes marked to shutdown won't enter a restart loop

edgurgel: Yeah but it might be removed more quickly if it's not healthy anymore. 

While in the normal process it gets a SIGTERM and it waits for the node to shutdown. Which in our case the Phoenix endpoint will stop all websockets etc.

PR #2123, opened 2026-08-21
chore: dependabot versioning strategy

PragTob:
odd one, docs say the option is supported but it fails and says only this one is supported:

```
The property '#/updates/0/versioning-strategy' value "increase-if-necessary" did not match one of the following values: lockfile-only, auto
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,9 +14,8 @@ updates:
     # Let's not flood ourselves with PRs
     open-pull-requests-limit: 3
 
-    # while the docs say versioning-strategy is supported by hex, seemingly only the default
-    # option is
-    versioning-strategy: "lockfile-only"
+    # not setting versioning strategy seemingly gives us, what we want,
+    # see: https://github.com/supabase/realtime/pull/2123#discussion_r3832772340
 
     cooldown:
       # same as in our hex config
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,9 +14,9 @@ updates:
     # Let's not flood ourselves with PRs
     open-pull-requests-limit: 3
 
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -14,9 +14,9 @@ updates:
     # Let's not flood ourselves with PRs
     open-pull-requests-limit: 3
 
-    # May widen `~>` constraints, but only when the latest resolvable version
-    # does not already satisfy them: in-range bumps touch mix.lock only.
-    versioning-strategy: "increase-if-necessary"
+    # while the docs say versioning-strategy is supported by hex, seemingly only the default
+    # option is
+    versioning-strategy: "lockfile-only"
```
leandrocp: @PragTob I also use dependabot on my projects and legit wanted to have `increase-if-necessary` so I did some digging into https://github.com/dependabot/dependabot-core and turns out the default is already `increase-if-necessary` so we just need to remove it:

```suggestion
```

Here's a real example https://github.com/hexpm/hexpm/pull/1545

leandrocp (dismissed):

.github/dependabot.yml
PragTob: it's so strange that it then won't allow it... thanks 💚

PragTob: Did the manual fix tor remove the comment anda lso add a new comment :)

filipecabaco (approved):

.github/dependabot.yml
leandrocp: From what I understood the hex implementation in dependabot-core doesn't support options from gh actions (the client) and just defaults to `increase-if-necessary` and that's the reason it's not "allowed".

PR #2149, opened 2026-08-31
chore: Better flaky digest report

PragTob:
* truncate better (by splitting our budget on multiple blocks and being more precise in our estimation)
* include links to concrete jobs with the failures (which is surprisingly hard) to be able to link to full traces
* strip random successful test output from the middle
* visually should be easier to parse now


PragTob:
Fixed previous failure and also rebased against `main`.

filipecabaco (approved):

PR #2150, opened 2026-08-31
feat: update pgdelta

leandrocp:
Closes REAL-1029


lib/mix/tasks/realtime.export_tenant_db_dump.ex
```diff
@@ -53,6 +54,24 @@ defmodule Mix.Tasks.Realtime.ExportTenantDbDump do
 
   defp dump_path(pg_major), do: Application.app_dir(:realtime, "priv/repo/tenant_db_dump_#{pg_major}.sql")
 
+  @doc false
+  def banner(pg_major) do
```
leandrocp, line 58: To inform why we also need the dump files

edgurgel (approved):

PR #2152, opened 2026-09-01
chore: add ARCHITECTURE.md

edgurgel:

Explain how Realtime is architectured with some highlights.

It's not an exhaustive description of everything!

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -147,9 +147,9 @@ That's cheap on the wire per hop, but every node in every region receives every 
 
 ### Highlight: Muster, so a broadcast only reaches nodes that care
 
-[`Forum.Muster`](forum/README.md) (vendored in `forum/`) answers one question: *which nodes hold at least one member of group G?* Realtime uses `tenant_id` as the group and the client's `transport_pid` as the member.
+[`Forum.Muster`](forum/README.md) (vendored in `forum/`) answers one question: *which nodes hold at least one member of a certain group?* Realtime uses `tenant_id` as the group and the client's `transport_pid` as the member.
 
-- **On join** (`RealtimeChannel.join/3`, flag `use_muster_channel_join`): the channel registers its transport pid into the region's Muster scope.
+- **On join** (`RealtimeChannel.join/3`, flag `use_muster_channel_join`): the channel registers its transport pid (with `tenant_id` as the group) into the region's Muster scope.
 - **On broadcast** (`GenRpcPubSub.broadcast/4`, flag `use_muster_broadcast`): instead of flooding, ask Muster for the tenant's **router** node.
 
 ```mermaid
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,0 +1,203 @@
+# Realtime Architecture
+
+A high-level tour of how Realtime is put together, plus a few highlights of the less obvious parts.
+
+## The Big Picture
+
+Realtime is a multi-tenant Phoenix app where every node in the cluster runs the same code. A tenant is identified by its `external_id`.
+
+Clients talk to Realtime over a single WebSocket that multiplexes many channels. Almost everything else in this document exists to serve one job: get a message published on a tenant's topic.
+
+---
+
+## The Main Flow
+
+### 1. Websocket & Channels
+
+`RealtimeWeb.UserSocket.connect/3`, mounted at `/socket` in `RealtimeWeb.Endpoint`:
+
+- The **tenant** comes from the request host: `Database.get_external_id/1` turns `<external_id>.realtime.example.com` into an `external_id`.
+- The **token** is the `x-api-key` header, falling back to the `apikey` param. It's a JWT verified against the tenant's `jwt_secret` / `jwt_jwks`.
+
+The process owning the WebSocket is the transport pid. It shows up everywhere below: it's the unit counted for `max_concurrent_users`, the member registered into Muster, and the pid that ultimately receives the encoded frame.
+Channel topics look like `realtime:<sub_topic>` and are all handled by `RealtimeWeb.RealtimeChannel`. One socket can hold many of them (`max_channels_per_client`, default 100).
+
+### 2. What a channel carries
+
+| Feature | Client sends | Produced by |
+| --- | --- | --- |
+| **Broadcast** | `"broadcast"` | another client, the HTTP API, or an insert into `realtime.messages` |
+| **Presence** | `"presence"` | other clients' track/untrack, diffed and synced per topic |
+| **Postgres Changes** | configured at join | the tenant's WAL, via `PostgresCdcRls` |
+
+### 3. Fan-out and receive
+
+`Realtime.PubSub` reaches other nodes through `Realtime.GenRpcPubSub`, an adapter built on [`gen_rpc`](https://github.com/emqx/gen_rpc) (wrapped locally by `Realtime.GenRpc`) instead of Erlang distribution, so nodes can have more than one TCP connection between them for more bandwidth. How it picks which nodes to send to is the subject of [Broadcast fan-out](#broadcast-fan-out).
+
+On each receiving node the local PubSub registry calls `MessageDispatcher.dispatch/3` with the topic's subscribers.
+
+### The two per-tenant singletons
+
+Two things are started once per tenant for the whole cluster on a node chosen by the placement rules in the next section.
+
+`Realtime.Tenants.Connect` owns everything that needs the *tenant's own Postgres*. It provides:
+
+- a **DBConnection pool** to the tenant database, published in `:syn` metadata as `conn:`, so any node can call `Connect.lookup_or_start_connection/1` and use that pool over RPC;
+- **broadcast from the database**: the `ReplicationConnection` streams inserts into `realtime.messages` off a logical replication slot and republishes each row as a broadcast on the tenant topic;
+- **lifecycle**: it stops itself when the tenant has had no connected users for a while or when it notices it's running in the wrong region.
+
+`Extensions.PostgresCdcRls` is the Postgres Changes driver. Per tenant it runs a `WorkerSupervisor` with a `SubscriptionManager` and a `ReplicationPoller`. What it provides:
+
+- turning a channel's `postgres_changes` config (schema / table / filter) into rows in the tenant's `realtime.subscription` table;
+- polling the tenant's WAL and delivering each change, already filtered per subscriber by `realtime.list_changes/4`, to only the nodes holding matching subscribers.
+
+See [Postgres Changes](#postgres-changes) for the details.
+
+---
+
+## Tenant placement
+
+The two singletons above (`Tenants.Connect` and the `PostgresCdcRls` `WorkerSupervisor`) have to exist exactly once per tenant across the whole cluster. Two things make that work: a deterministic node picker, and `:syn` as the registry that enforces uniqueness.
+
+```mermaid
+sequenceDiagram
+    participant CH as Channel (any node)
+    participant SYN as :syn registry
+    participant NODES as Realtime.Nodes
+    participant OWNER as Chosen node
+
+    CH->>SYN: lookup(Connect, tenant_id)
+    alt already registered
+        SYN-->>CH: pid + metadata (conn: db_conn)
+    else not registered
+        CH->>NODES: get_node_for_tenant(tenant)
+        NODES-->>CH: {:ok, node, region}
+        CH->>OWNER: GenRpc call Connect.connect/3
+        OWNER->>SYN: register {Connect, tenant_id}, conn: nil
+        OWNER->>OWNER: migrations, DB pool, ReplicationConnection
+        OWNER->>SYN: update metadata conn: pid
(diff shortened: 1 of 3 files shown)


ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+## The Main Flow
+
+### 1. Websocket & Channels
+
+`RealtimeWeb.UserSocket.connect/3`, mounted at `/socket` in `RealtimeWeb.Endpoint`:
+
+- The **tenant** comes from the request host: `Database.get_external_id/1` turns `<external_id>.realtime.example.com` into an `external_id`.
+- The **token** is the `x-api-key` header, falling back to the `apikey` param. It's a JWT verified against the tenant's `jwt_secret` / `jwt_jwks`.
+
+The process owning the WebSocket is the transport pid. It shows up everywhere below: it's the unit counted for `max_concurrent_users`, the member registered into Muster, and the pid that ultimately receives the encoded frame.
```
seangeo: I wonder if a glossary might be useful here? I remember reading about frame encodings in the Realtime docs* while playing around with it a couple months ago but it might be worth defining it, and maybe linking to those docs?

* Unless I hallucinated that.

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+Channel topics look like `realtime:<sub_topic>` and are all handled by `RealtimeWeb.RealtimeChannel`. One socket can hold many of them (`max_channels_per_client`, default 100).
+
+### 2. What a channel carries
```
seangeo, line 25: I'm a bit confused about what "Client" refers to here. Does it mean a web socket client? Or are you referring to what Realtime sends back on the socket? I think the confusion is from the third heading "Produced by", which seems to reference some events that will trigger some data being sent back to the subscriber (which is the client).

Or is "Client sends" the type of subscription the client is creating?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+
+| Feature | Client sends | Produced by |
+| --- | --- | --- |
+| **Broadcast** | `"broadcast"` | another client, the HTTP API, or an insert into `realtime.messages` |
+| **Presence** | `"presence"` | other clients' track/untrack, diffed and synced per topic |
+| **Postgres Changes** | configured at join | the tenant's WAL, via `PostgresCdcRls` |
+
+### 3. Fan-out and receive
+
+`Realtime.PubSub` reaches other nodes through `Realtime.GenRpcPubSub`, an adapter built on [`gen_rpc`](https://github.com/emqx/gen_rpc) (wrapped locally by `Realtime.GenRpc`) instead of Erlang distribution, so nodes can have more than one TCP connection between them for more bandwidth. How it picks which nodes to send to is the subject of [Broadcast fan-out](#broadcast-fan-out).
```
seangeo: I read this as what we're doing here is using direct RPC between nodes for higher bandwidth, instead of using the usual Erlang cluster communication? Am I getting that right?

Do we use Erlang cluster distribution for anything?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+
+On each receiving node the local PubSub registry calls `MessageDispatcher.dispatch/3` with the topic's subscribers.
+
+### The two per-tenant singletons
+
+Two things are started once per tenant for the whole cluster on a node chosen by the placement rules in the next section.
+
+`Realtime.Tenants.Connect` owns everything that needs the *tenant's own Postgres*. It provides:
```
seangeo, line 43: Does this mean that there's one connection pool per tenant in the entire cluster?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+
+- a **DBConnection pool** to the tenant database, published in `:syn` metadata as `conn:`, so any node can call `Connect.lookup_or_start_connection/1` and use that pool over RPC;
+- **broadcast from the database**: the `ReplicationConnection` streams inserts into `realtime.messages` off a logical replication slot and republishes each row as a broadcast on the tenant topic;
```
seangeo, line 46: Is this how Postgres Changes are published or is this for providing the message broadcasting feature? Or maybe both?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+- **lifecycle**: it stops itself when the tenant has had no connected users for a while or when it notices it's running in the wrong region.
+
+`Extensions.PostgresCdcRls` is the Postgres Changes driver. Per tenant it runs a `WorkerSupervisor` with a `SubscriptionManager` and a `ReplicationPoller`. What it provides:
+
+- turning a channel's `postgres_changes` config (schema / table / filter) into rows in the tenant's `realtime.subscription` table;
+- polling the tenant's WAL and delivering each change, already filtered per subscriber by `realtime.list_changes/4`, to only the nodes holding matching subscribers.
```
seangeo, line 52: How is RLS enforced here? As I understand it WAL has no concept of RLS at all since it's not aware of the user, so something would need to take a change that we got from the WAL, figure out the subscribers that care about that change, and then apply relevant RLS policies right?

Does the config that we store in realtime.subscription for a channel include the details of the user who created the subscription so the RLS policies can be applied.

Very likely I'm missing something in all this. :D

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+    B -->|one| D["that node"]
+    B -->|many| E["seed RNG with phash2({tenant_id, time_bucket})"]
+    E --> F["pick 2 candidate nodes<br/>(same result on every node in the bucket)"]
+    F --> G["node_load/1 on both<br/>:cpu_sup.avg5, cached per node"]
+    G -->|both known| H["least loaded wins"]
+    G -->|not enough data| I["phash2(tenant_id) over sorted nodes"]
+```
+
+- Region membership comes from `:syn.members(RegionNodes, region)`, sorted for stability. Tenant regions are mapped to the nearest Realtime region by `platform_region_translator/1` (overridable via `REGION_MAPPING`).
+- The time bucket (default 60s, `:node_selection_time_bucket_seconds`) is the trick: every node computing in the same window picks the same two candidates, so concurrent requests agree. The next window picks a different pair, so placement spreads out over time.
```
seangeo: How do these time buckets work? Is anything done to prevent issues like a node asking to launch for a tenant in the next time bucket when a launch request for the previous time bucket hasn't finished and been registered?

ARCHITECTURE.md
seangeo: Also wondering if that has caused issues with a single point of failure or bottlenecking? Any thought been given to one node per availability zone in the closest region to the tenant's DB, or something like that?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Highlight: syn as the cluster-wide registry
+
+[`:syn`](https://github.com/ostinelli/syn) gives us a distributed process registry with metadata and conflict resolution. Realtime uses three kinds of scopes, set up in `Realtime.Application`:
+
+| Scope | What's registered | Used for |
+| --- | --- | --- |
+| `RegionNodes` | one process per node, joined to a group named after its region | region membership → node selection, region flood |
+| `Realtime.Tenants.Connect` | the `Connect` GenServer, keyed by `tenant_id` | find the tenant's DB pool from any node |
+| `realtime_postgres_cdc_<shard>` | the CDC `WorkerSupervisor`, keyed by `tenant_id` | find the tenant's poller; sharded (`:postgres_cdc_scope_shards`) so one scope's table isn't a hot spot |
```
seangeo, line 119: What are the shards based on in this context? does that actually mean there's multiple CDC supervisors per tenant?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+    O -->|"ftr: one representative"| Y1["node Y1, region Y"]
+    Y1 -->|"re-floods its region"| Y2["node Y2"]
+    Y1 --> Y3["node Y3"]
+```
+
+That's cheap on the wire per hop, but every node in every region receives every tenant's broadcast, even nodes with zero connections for that tenant.
+
+### Highlight: Muster, so a broadcast only reaches nodes that care
+
+[`Forum.Muster`](forum/README.md) (vendored in `forum/`) answers one question: *which nodes hold at least one member of group G?* Realtime uses `tenant_id` as the group and the client's `transport_pid` as the member.
```
seangeo: Nitpick "G" as in "group G" isn't used anywhere else ;)

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+
+- **On join** (`RealtimeChannel.join/3`, flag `use_muster_channel_join`): the channel registers its transport pid into the region's Muster scope.
```
seangeo: I assume here it's registering its transport pid against the group identified by the tenant id?

ARCHITECTURE.md
```diff
@@ -0,0 +1,203 @@
...
+    RP -->|"realtime.list_changes(...)"| DB
+    RP -->|"direct_broadcast to only the nodes<br/>holding matching subscribers"| SUB["Subscriber nodes"]
+    SUB --> CH["Channel fastlane → client"]
+```
+
+---
+
+## Presence
+
+Presence is per tenant-topic state synced between nodes over the same PubSub. Each channel tracks the client's transport pid; joins/leaves and periodic syncs are diffed and broadcast to the topic. For private channels, presence writes are authorized via `Tenants.Authorization` policies evaluated against the tenant DB.
```
seangeo, line 228: Does Presence do the same tenant based routing optimisations?

What happens when a client connects to a node in a region that doesn't have their tenant's presence information?  

Or is presence global at the moment?

seangeo (commented):
This is excellent!

I've just added some questions and clarifications to make sure I understand things.

I think it would be really good to see a deployment architecture view too, I'm assuming there's some load balancer that routes web socket clients to nodes in their nearest region or something like that?

Thanks so much for putting this together, very timely. :D

leandrocp (approved):
This is 👌🏻

ARCHITECTURE.md
edgurgel: Oh yeah "frame" here is just the term that WebSockets refer to as the sort of the basic unit of data sent. Can clarify!

ARCHITECTURE.md
edgurgel: We use Erlang cluster distribution! Syn & Muster uses Erlang distribution and some other bits and pieces like the Phoenix Live Dashboard uses `erpc` under the hood I think. 

We try to send "user data" through `gen_rpc` as we have more control of the bandwidth as Erlang distribution will only hold 1 connection which can impact the connection clustering.

ARCHITECTURE.md
edgurgel: > Does this mean that there's one connection pool per tenant in the entire cluster?

Exactly!

> Also wondering if that has caused issues with a single point of failure or bottlenecking? Any thought been given to one node per availability zone in the closest region to the tenant's DB, or something like that?

The tricky bit is that if we have multiple pools we need to ensure that they are evenly split as we have a max number of connections we establish. But definitely worth exploring.

ARCHITECTURE.md
edgurgel: Providing message broadcasting from the database. That's a different replication! We have 1 replication slot just for `realtime.messages` that goes into "broadcast" messages. See https://supabase.com/docs/guides/realtime/broadcast?queryGroups=language&language=js#broadcast-from-the-database


Postgres Changes replicates data from any table that was added the to publication. That's a separate replication slot

Will make it clearer

ARCHITECTURE.md
edgurgel: It just means that we use separate Syn registries because Syn doesn't have the concept of partitions like the Elixir Registry has. One tenant goes always to a specific shard.

ARCHITECTURE.md
edgurgel: yeah! will clarify.

ARCHITECTURE.md
edgurgel: Global!

ARCHITECTURE.md
edgurgel: Good question. We accept the possibility of overlap when the request is being fired. If we have different nodes being selected Syn has a conflict resolution process! Will add this

ARCHITECTURE.md
edgurgel: Made it clear that it's a Websocket frame

ARCHITECTURE.md
edgurgel: I've updated the section. Let me know if it's better now

ARCHITECTURE.md
edgurgel: I've added a note on Erlang distribution

ARCHITECTURE.md
edgurgel: Added a sentence talking about both slots

ARCHITECTURE.md
edgurgel: I've tried to explaing a bit better how it behaves in the edges

seangeo (approved):
Excellent doc! Thanks @edgurgel!

PR #2154, opened 2026-09-02
chore: Extend ARCHITECTURE.md a bit

PragTob:
* add a glossary/extend some wording that I think helps understandingog
* mention the multi-node/multi-region charateristic upfront along with self-hosting.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -6,7 +6,7 @@ A high-level tour of how Realtime is put together, plus a few highlights of the 
 
 Realtime is a multi-tenant Phoenix app where every node in the cluster runs the same code. A tenant is identified by its `external_id`.
 
-Realtime is deployed across multiple regions in a single cluster. Realtime talks to its own database as well as to individual tenant databases.
+Realtime is designed to run across multiple regions as a single inter-connected global cluster. Realtime talks to its own database as well as to individual tenant databases.
 Realtime doesn't have to be multi-node and multi-region though, it can also be run self-hosted for example in a single node environment.
 
 Clients talk to Realtime over a single WebSocket that multiplexes many channels. Almost everything else in this document exists to serve one job: get a message published on a tenant's topic.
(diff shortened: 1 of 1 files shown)


ARCHITECTURE.md
```diff
@@ -6,10 +6,23 @@ A high-level tour of how Realtime is put together, plus a few highlights of the
...
 Clients talk to Realtime over a single WebSocket that multiplexes many channels. Almost everything else in this document exists to serve one job: get a message published on a tenant's topic.
 
 ---
 
+## Glossary
+
+* [`Muster`](forum/README.md) - a library which helps answer "Which nodes hold clients of a tenant?"
+* [OID](https://www.postgresql.org/docs/current/datatype-oid.html) - object identifiers, for instance for relations
+* postgres_changes entry - a subscription on particular set of postgres changes (table, action, filters etc.)
+* [replication slot](https://www.postgresql.org/docs/current/warm-standby.html#STREAMING-REPLICATION-SLOTS) - ensures retention of relevant WAL entries
```
edgurgel: I think a better description would be: Postgres Change Data Capture (CDC) streams ?

edgurgel (approved):

ARCHITECTURE.md
PragTob: For replication slots? That in itself I think I wouldn't understand 😅 

I basically tried rewording and shortening the first line from the Postgres docs:

> Replication slots provide an automated way to ensure that the primary server does not remove WAL segments until they have been received by all standbys, and that the primary does not remove rows which could cause a [recovery conflict](https://www.postgresql.org/docs/current/hot-standby.html#HOT-STANDBY-CONFLICT) even when the standby is disconnected.

edgurgel: My comment was more around that "retention" is not the most important aspect here because we don't use a permanent replication slots but temporary ones. 

The reason why Realtime uses (logical) replication slots is to process stream of changes coming from the WAL and translating them into events that can be subscribed to.

And for standbys replication is AFAIK "physical" and not logical replication.

PR #2156, opened 2026-09-02
fix: don't log negative `DatabaseLackOfConnections`

leandrocp:
Closes REAL-495


test/realtime/database_test.exs
```diff
@@ -83,6 +83,19 @@ defmodule Realtime.DatabaseTest do
              end) =~ ~r/Only \d+ available connections\. At least 125 connections are required/
     end
 
+    test "guard against negative available connnections count", %{tenant: tenant} do
+      expect(Postgrex, :query!, 2, fn _conn, query, _params ->
```
edgurgel: I'm wondering if we could fake connections with applicatin name for `autovacuum` etc to simulate this for real. 🤔

edgurgel: Mostly because I would like to avoid copying Postgrex as even though Mimic is pretty fast it adds a tiny bit of overhead on things that are copied

leandrocp: Makes sense! I'm trying to add a real db test without introducing a flaky test otherwise @PragTob is gonna have more work to do 😄

PragTob: I appreciate both 😁 

iirc the Mimic benchmarks it's not that bad - since we're dealing with databases a lot I don't think it'd add meaningfully to the execution time (as every call to the DB is just SO MUCH slower) of our test suite. We're talking about micro seconds per call I think so for that to add up for even a second.

And since mimic works in `async: true` - if we can switch some tests to `async: true` because of using it, I think it'd be an overall gain.

But that said, stilla gree with keeping it limited.

Sorry, I don't think I'm being very helpful but wanted to weigh in as it's an interesting discussion 😅

leandrocp: Added `@tag db_args: ["-c", "max_connections=10"]` to control number of conns to simulate the problem in a test against real DB.

lib/realtime/database.ex
```diff
@@ -166,8 +166,10 @@ defmodule Realtime.Database do
   SELECT count(*)::int FROM realtime.schema_migrations
   """
 
+  # Autovacuum workers, walsenders and background workers are listed here without holding a
```
edgurgel, line 169: Looking at the docs it says:


>Note that the number of running workers does not count towards [max_connections](https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-MAX-CONNECTIONS) or [superuser_reserved_connections](https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-SUPERUSER-RESERVED-CONNECTIONS) limits.

https://www.postgresql.org/docs/current/routine-vacuuming.html#AUTOVACUUM

So what is the actual issue here?

Would it help to filter by `backend_type`?

leandrocp: Yes! Good catch. Fixed it.

test/realtime/database_test.exs
```diff
@@ -83,6 +86,27 @@ defmodule Realtime.DatabaseTest do
              end) =~ ~r/Only \d+ available connections\. At least 125 connections are required/
     end
 
+    @tag db_args: ["-c", "max_connections=10"]
+    test "counts only the backends holding a connection slot", %{tenant: tenant} do
```
edgurgel: Can you explain why do we need to decrease `max_connections` instead of bumping the pool_size that a tenant wants to grab?

leandrocp: That was due to my concern to have a deterministic number to assert on tests to avoid flaky tests but it introduced many other changes to pass along `postgres_args`, so I removed to keep changes minimal. Bumping db_pool to a high number also works, but the test must find the available conns to assert the actual log entry which I think is fine for this PR.

edgurgel (approved):

PR #2158, opened 2026-09-02
chore: define `TENANT` to isolate dev and tests

leandrocp:
This is a small improvement to https://github.com/supabase/realtime/pull/2110

- Derive `TEST_RUN` from `TENANT` so that's the only var we need to define to create an isolated dev or test env
- Define `TENANT` from current dir so it "just works" but keep `realtime-dev` on `realtime` dir (default)

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -33,14 +33,29 @@ node_suffix =
     _ -> "_#{port}"
   end
 
-# Names this run's database and containers. TEST_RUN labels a run ("pr_1234"), and otherwise the
-# endpoint port does, which also lets a later run tell whether the run that left something
-# behind is gone.
+# Names this run's database and containers. One knob covers a whole environment: TENANT names a
+# test run just as it names the dev stack, so `TENANT=review mise run db-start` and
+# `TENANT=review mix test` are the same review environment. The default tenant is the unmarked
+# run, so a plain `mix test` keeps the names it always had. TEST_RUN overrides TENANT for a run
+# that wants a label without a tenant behind it ("pr_1234").
+default_tenant = "realtime-dev"
+
+run_name =
+  case {System.get_env("TEST_RUN", ""), System.get_env("TENANT", "")} do
+    {"", ^default_tenant} -> ""
+    {"", tenant} -> tenant
+    {run, _} -> run
+  end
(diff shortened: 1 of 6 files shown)


config/test.exs
```diff
@@ -33,14 +33,29 @@ node_suffix =
...
-# behind is gone.
+# Names this run's database and containers. One knob covers a whole environment: TENANT names a
+# test run just as it names the dev stack, so `TENANT=review mise run db-start` and
+# `TENANT=review mix test` are the same review environment. The default tenant is the unmarked
+# run, so a plain `mix test` keeps the names it always had. TEST_RUN overrides TENANT for a run
+# that wants a label without a tenant behind it ("pr_1234").
+default_tenant = "realtime-dev"
+
+run_name =
+  case {System.get_env("TEST_RUN", ""), System.get_env("TENANT", "")} do
```
edgurgel, line 44: Do you know if config/test.exs is always loaded when the test runs or when the project is compiled on MIX_ENV=test? I thought only runtime.exs would necessarily run for every mix test

leandrocp: AFAIK it's always evaluated, on compile and run.Are you concerned  `System.get_env` might return wrong values?

edgurgel: Yeah I was concerned that it wouldn't re-evaluate

leandrocp: I _think_ it's fine but now you made me doubt myself 😄
Let me check.

edgurgel (approved):

config/test.exs
leandrocp: Yeah it's fine:

```elixir
# config/text.exs
System.get_env("TENANT", "") |> IO.inspect()
```

```shell
❯ export TENANT="FOO"

❯ MIX_ENV=test mix compile
"FOO"

❯ mix test
"FOO"
18:09:38.312 [info] == Running 20210706140551 Realtime.Repo.Migrations.CreateTenants.change/0 forward
...
```

edgurgel: Great!

leandrocp:
Merging despite the failed actions because tests are passing and this PR doesn't update any schema.

PR #2159, opened 2026-09-02
fix: normalize Realtime `:region`

leandrocp:
Closes REAL-965

Follow up of https://github.com/supabase/realtime/pull/2073#discussion_r3754629473


lib/realtime/nodes.ex
```diff
@@ -61,6 +61,16 @@ defmodule Realtime.Nodes do
...
 
+  @doc """
+  The region this node registers under in the `RegionNodes` group.
+
+  Deployments that leave `:region` unset (self-hosted, CLI) still form a cluster,
+  so they share `"local"` as a canonical group name. Both the `:syn.join` and
+  every read of this node's region must go through here.
+  """
+  @spec region() :: String.t()
+  def region, do: Application.get_env(:realtime, :region) || "local"
```
edgurgel: not against this but what if we have this default set on `runtime.exs` and just do fetch_env! here?

leandrocp: Probably better, the change I made was more based on:

```elixir
region = Application.get_env(:realtime, :region)
muster_scope = :"realtime_channels_#{region || "local"}"
```

But yeah `runtime.exs` should handle that.

test/realtime/nodes_test.exs
```diff
@@ -38,6 +38,30 @@ defmodule Realtime.NodesTest do
     end
   end
 
+  describe "region/0" do
+    setup do
+      previous_region = Application.get_env(:realtime, :region)
+      on_exit(fn -> Application.put_env(:realtime, :region, previous_region) end)
```
PragTob: :old-man-screams-at-global-state:

We can leave at is, but alternatives I'm aware of:

1. stub it with mimic
2. actually make the call as a default arg (`def region(node \\ Application.fetch_env!(:realtime, :region), do: node`) that way you can control it for a test
3. not test it - we're just wrapping `Application.fetch_env!` which I think is fine to leave untested as it's a lib function we trust and I'm rather sure the function is indirectly exercised (although it feels weird to have a public function that isn't tested). Could also just test that it returns some string 🤷

PragTob (approved):
Looks good to me 👌 

Only comment is on `Application.put_env`/global state in the tests which I'd love to avoid :D Not mandatory though, but left 3 paths I know of to (potentially) avoid it.

Image

test/realtime/nodes_test.exs
leandrocp: I don't like global state either and since `region/1` is extensively tested (indirectly throughout the entire codebase) I think it's better to just remove the those tests (they are not adding any value to our test suite anyway).

edgurgel (approved):

edgurgel:
FYI https://github.com/orgs/supabase/discussions/50055#discussioncomment-18341260

PR #2161, opened 2026-09-03
fix: remove failed janitor tasks from state

### LINKED ISSUE #2160: Janitor retains failed maintenance tasks in state
### Summary

The tenant Janitor retains failed maintenance task references in its GenServer state after receiving a terminal `:DOWN` message.

Successful tasks are removed with `Map.pop/2`, but the `:killed` handler reads the tenants with `Map.get/2` and writes the unchanged task map back to state. Other non-normal exit reasons fall through to the catch-all handler and are retained as well.

Because every scheduled run merges new task refs into this map, repeated maintenance failures can grow the Janitor state indefinitely and retain each associated tenant batch.

### Reproduction

```elixir
ref = make_ref()
state = %Realtime.Tenants.Janitor{tasks: %{ref => ["tenant"]}}

{:noreply, result} =
  Realtime.Tenants.Janitor.handle_info(
    {:DOWN, ref, :process, self(), :killed},
    state
  )

Map.has_key?(result.tasks, ref)
#=> true
```

### Expected behavior

Any tracked task should be removed from `state.tasks` when its terminal `:DOWN` message arrives, whether it completed normally or failed.

### Environment

- Realtime `main` at 1aec648a
- Elixir 1.19.5
- OTP 28

hsusul:
## Summary

- remove tracked Janitor task refs after non-normal terminal `:DOWN` messages
- preserve catch-all handling for unrelated monitor messages
- extend the existing Janitor integration test to inspect the running GenServer state

## Why

Successful tasks were removed from the GenServer state, but failed tasks remained in `state.tasks`. Repeated maintenance failures could therefore grow the map indefinitely and retain tenant batches.

Closes #2160.

## Testing

- `mix format --check-formatted lib/realtime/tenants/janitor.ex test/realtime/tenants/janitor_state_test.exs`
- `MIX_ENV=test mix compile --warnings-as-errors` under Elixir 1.19.5 / OTP 28
- the existing `JanitorTest` integration suite runs in CI with its required Postgres services
- `git diff --check`


test/realtime/tenants/janitor_state_test.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Realtime.Tenants.JanitorStateTest do
```
edgurgel: Could we instead add a test to existing JanitorTest file? We can use `:sys.get_state` to inspect the GenServer state if that's necessary. WDYT?

edgurgel (changes requested):

test/realtime/tenants/janitor_state_test.exs
hsusul: Yes, that makes sense. I initially kept it separate as a database-independent unit test, but testing the running GenServer is more representative. I’ll move the regression into `JanitorTest`, trigger the failed task path, and use `:sys.get_state/1` to verify the completed task reference is removed.

edgurgel (approved):

edgurgel:
Thank you! ❤️

PR #2167, opened 2026-09-04
fix: ranch charlist log filter

### LINKED ISSUE #2163: LogFilter misses Ranch :killed reports because Ranch emits a charlist format
## Bug report

`Realtime.LogFilter` does not suppress the Ranch `:killed` report produced by Ranch 2.2.1 because the filter matches a binary format string while Ranch emits a charlist.

Realtime v2.129.13 defines:

```elixir
@ranch_format "Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"
def filter(%{msg: {:format, @ranch_format, [_, _, _, :killed]}}, _), do: :stop
```

Ranch 2.2.1 calls `ranch:log/4` with an Erlang double-quoted string in `ranch_conns_sup.erl`, so Logger receives the format as a charlist. `ranch:log/4` forwards `Format` unchanged. The rendered message looks identical, but the exact pattern does not match.

The current test helper uses the same Elixir binary module attribute as the implementation, so it does not reproduce the runtime event shape.

## Reproducer

```elixir
binary_format = "Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"
charlist_format = String.to_charlist(binary_format)
event = %{msg: {:format, charlist_format, [:listener, :cowboy_clear, self(), :killed]}, meta: %{}}

assert :stop = Realtime.LogFilter.filter(event, [])
```

This assertion fails on v2.129.13 and the event falls through to the primary logger.

## Suggested fix

Accept the Ranch format in both binary and charlist form, or normalize the format before comparing it. Please add the charlist-shaped event as the regression test; keeping only the binary fixture will continue to miss the shape Ranch actually emits.

This matters operationally because Realtime intentionally terminates transports with no joined channels, so the failed filter can turn routine tracker cleanup into large error-log bursts.

mandar1045:
## TL;DR
Handles the charlist log format emitted by `ranch` 2.2.1+ so the `Realtime.LogFilter` actually stops the `:killed` error logs from surfacing. 

## reference
- closes #2163

### Before (Test Fails on Charlist Event)
Before Fix

### After (Tests Pass)
After Fix 1
After Fix 2
After Fix 3


lib/realtime/log_filter.ex
```diff
@@ -28,8 +28,11 @@ defmodule Realtime.LogFilter do
 
   def filter(%{meta: %{mfa: {DBConnection.Connection, _, _}}}, _), do: :stop
 
-  @ranch_format "Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"
+  @ranch_format ~c"Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"
+  @ranch_format_binary "Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"
+
   def filter(%{msg: {:format, @ranch_format, [_, _, _, :killed]}}, _), do: :stop
+  def filter(%{msg: {:format, @ranch_format_binary, [_, _, _, :killed]}}, _), do: :stop
```
leandrocp: I believe the root cause here is that extra `:format` causing a pattern mismatch. The fix is actually more simple:

```elixir
@ranch_format ~c"Ranch listener ~p had connection process started with ~p:start_link/3 at ~p exit with reason: ~0p~n"

def filter(%{msg: {@ranch_format, [_, _, _, :killed]}}, _), do: :stop
```

Then adapt the tests for that format.

Ideally we need tests that actually open and close connections instead of simple mocked events, but not required in this PR :)

leandrocp (changes requested):
Hi @mandar1045 thanks that's a real confirmed bug! But I think the test is actually just fixing the pattern format.

lib/realtime/log_filter.ex
mandar1045: Thanks @leandrocp That’s a much cleaner fix. I’ll update the pattern match to drop the `:format` wrapper and simplify the tests around it. I’ll push the changes shortly.

edgurgel: Also can we get a test that actually test that killing a cowboy process logs this and it's filtered? Maybe under `Integration.TrackerTest` ? Ideally the test should fail without the fix I guess? Otherwise we won't notice if this happens again

leandrocp: @edgurgel I agree but I'd suggest it in a following PR to get this fix merged. I'm fine with either, wdyt @mandar1045?

edgurgel: Happy either way!

mandar1045: Thanks @leandrocp and @edgurgel  I’ve made the suggested changes and updated the tests. We can do the integration test in a follow-up PR and get this one merged first. I can take that up after this .

leandrocp (approved):

leandrocp:
Thanks @mandar1045 !

mandar1045:
> Thanks @mandar1045 !

i'll proceed with the integration tests now , thanks for reviewing ❤️

PR #2168, opened 2026-09-04
fix: use runuser to drop privileges to nobody in run.sh

### LINKED ISSUE #6462: supabase start fails since v2.112.0: realtime container exits with 'sudo: account validation failure' for the nobody account
### Describe the bug

Starting with **CLI v2.112.0**, `supabase start` fails deterministically in our containerised CI runners while initialising the schema. The `realtime` container exits because `sudo` cannot switch to the `nobody` account:

```
Starting database...
Initialising schema...
+ export ERL_CRASH_DUMP=/tmp/erl_crash.dump
+ echo 'Running migrations'
+ sudo -E -u nobody /app/bin/migrate
sudo: account validation failure, is your account locked?
sudo: a password is required
Stopping containers...
error running container: exit 1
```

Without `--debug` the only output is `error running container: exit 1`, which is why this took a while to pin down — it is indistinguishable from unrelated container failures.

### To Reproduce

In a Linux container with Docker available and a project with local migrations:

```sh
supabase stop --no-backup
supabase start        # fails at "Initialising schema"
```

### Bisection

We bisected across stable releases on a single machine, 3 `supabase start` attempts per version, same project, same host, same Docker:

| CLI version | `supabase start` |
|---|---|
| 2.116.0 | **0/3** |
| 2.115.0 | **0/3** |
| 2.114.0 | **0/3** |
| 2.113.0 | **0/3** |
| 2.112.0 | **0/3** |
| **2.111.0** | **3/3 OK** |
| 2.109.1 | 6/6 OK |

**The regression is introduced in 2.112.0** and is still present in 2.116.0. All five failing versions produce the identical `sudo: account validation failure` message, so this looks like a change in the bundled `realtime` image rather than in the CLI binary itself.

### Expected behavior

`supabase start` completes, or at minimum surfaces the underlying container error without requiring `--debug`.

### System information

- OS: Ubuntu (Linux kernel 7.0.0-30-generic), x86_64
- Docker: 29.1.3 (client and server), storage driver `overlayfs`, cgroup driver `systemd` v2
- Supabase CLI: installed from the official `.deb` release artifacts
- Postgres image pulled by 2.116.0: `public.ecr.aws/supabase/postgres:17.6.1.165`
- Runs as a non-root user inside an unprivileged system container

### Additional context

We have pinned to **2.111.0** as a workaround. Happy to provide more of the `--debug` output or test a patched image if that helps.

mandar1045:
small fix : sudo -u nobody fails when the nobody account is locked (PAM checks it, runuser doesn't). Switched to runuser -m -u nobody — same env passthrough as before, tested on both root and non-root.

closes supabase/cli#6462

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -99,7 +99,7 @@ run_as_nobody() {
         if command -v runuser >/dev/null 2>&1; then
             runuser -u nobody -- "$@"
         elif command -v su >/dev/null 2>&1; then
-            su -s /bin/sh nobody -c '"$@"' -- "$@"
+            su -s /bin/sh nobody -c 'exec "$@"' _ "$@"
         else
             sudo -E -u nobody "$@"
         fi
--- a/run.sh
+++ b/run.sh
@@ -94,12 +94,26 @@ if [[ -n "${GENERATE_CLUSTER_CERTS:-}" ]] ; then
     generate_certs
 fi
 
+run_as_nobody() {
+    if [ "$(id -u)" = "0" ]; then
+        if command -v runuser >/dev/null 2>&1; then
+            runuser -u nobody -- "$@"
+        elif command -v su >/dev/null 2>&1; then
+            su -s /bin/sh nobody -c '"$@"' -- "$@"
+        else
+            sudo -E -u nobody "$@"
+        fi
+    else
+        "$@"
+    fi
+}
+
 echo "Running migrations"
-sudo -E -u nobody /app/bin/migrate
+run_as_nobody /app/bin/migrate
 
 if [ "${SEED_SELF_HOST-}" = true ]; then
     echo "Seeding selfhosted Realtime"
-    sudo -E -u nobody /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'
+    run_as_nobody /app/bin/realtime eval 'Realtime.Release.seeds(Realtime.Repo)'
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -94,12 +94,20 @@ if [[ -n "${GENERATE_CLUSTER_CERTS:-}" ]] ; then
     generate_certs
 fi
 
+run_as_nobody() {
+    if [ "$(id -u)" = "0" ]; then
+        runuser -m -u nobody -- "$@"
```
7ttp, line 99: tbh i don't think this is the right fix

sudo checks the account of whoever runs it, which in run.sh is root, 
not the `nobody` user it's switching to, and `nobody` isn't locked in any of these images anyway (v2.120.3 through v2.130.0 all look the same)

 so this just skips that check instead of telling us why it's failing on that runner....

7ttp:
imo this is pretty much environment specific,
 so i'll follow up with the reporter on the cli issue to see what's really going on...

thanks anyways for the effort!

mandar1045:
> imo this is pretty much environment specific, so i'll follow up with the reporter on the cli issue to see what's really going on...
> 
> thanks anyways for the effort!

> Thanks for looking into it! I’ll close this and follow up on the CLI issue to find the root cause.

PR #2184, opened 2026-09-09
refactor: split e2e tests into modules

seangeo:

Refactored the e2e test script. This was a single 2112 line file, I've split this up into separate modules, where runner and fixtures are in their own modules, and each test suite is in a separate file under `src/suites`. This should make it a bit easier to reason about and to add new test suites over time.

This is a refactor only, the behaviour and tests are exactly the same.

Single large file for all tests.

## What is the new behavior?

Multiple smaller files, with each test suite in its own file.

The test suite still does the same thing, I've tested it locally compared to the suite on main and they both produce the same output.

## Additional context

It's a pretty large change, but each step has been done in a single commit, in case that makes it easier to review.

Would be great if someone can test some of the ways they use this test suite too, particularly with Otel, just to verify some other use cases.

Part of REAL-1037


test/e2e/README.md
```diff
@@ -131,6 +131,16 @@ bun run nix
 
 ---
 
+## Source layout
```
PragTob, line 134: Thanks for adding this 👍

PragTob (approved):
👋 

Looks great to me and thanks for the work! Looks like a great improvement.

Only thing I'm not too sure about is the way it's split right now - that may partially be too fine grained. I.e. there are 4 `broadcast{-*}` files. I guess they could all be one `broadcast.ts` - but no strong opinion here. 

There another `load-broacast` where I'm also in general not too certain what the `load-` prefix is. Looking at the tests I suppose it's a mini load test/behaviour under load. I think that's good to separate out (as in: good distinction to make). I'd go one step further here and put them into a `load` subfolder (potentially with a README for people like me to understand that these test the feature under load).

I think the separation of load/no load is good anyhow, as we might want to exclude them/run them with different parameters.

Image

seangeo:
Thanks for the review @PragTob! The split into suites is the same as it was originally I’ve just moved them into separate files, agree that would could think more about the structure there, hopefully it’s easier to do now!

PR #2186, opened 2026-09-09
ci: migrate portable packaging to the shared Nix flake

jgoux:
The Nix portable packaging job fails before building because slim-services removed `nix-build-with-derived-hashes.sh` during its root-flake migration. This updates the workflow to call `build-artifact-from-nix.sh`, which resolves dependency hashes and builds the portable runtime from the exact Realtime commit under test.

- Pin slim-services to `4a4073b1e57f4422d203863decc94baa76640cf7` and check out Realtime at the source path expected by its recipe.
- Pass the checked-out commit as `SOURCE_REF`, update the Nix cache inputs, and run the existing crypto smoke against the exported runtime.
- Remove the generic failure message that incorrectly suggested an Elixir/OTP or dependency failure.

Validation: actionlint 1.7.7 (including ShellCheck) and `git diff --check` pass. The full portable build and smoke will run in this PR's Packaging workflow.

Original failure: https://github.com/supabase/realtime/actions/runs/34129031064/job/101764429406

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -59,36 +59,39 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
-          path: realtime
+          repository: supabase/slim-services
+          ref: 4a4073b1e57f4422d203863decc94baa76640cf7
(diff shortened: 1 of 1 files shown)


PragTob (approved):
Image

PR #2187, opened 2026-09-09
feat(telemetry): log request path in HTTP response logs

YorviArias:
Feature: 
- HTTP response logs showed only status and duration, not which request they were for
- Added the method and path to the log line
- Removed Phoenix's default line for the same event, so each response logs once

Testing:
- Added a test checking the new line has the path and status
- Added a test checking Phoenix's old line is gone
- Checked live against a local server: works for normal responses (401 on `/api/ping`).

Resolves REAL-793

Please review thoroughly. Hoping things are correct and actually tackled what we wanted.

### DIFF
--- a/test/realtime/telemetry/logger_test.exs
+++ b/test/realtime/telemetry/logger_test.exs
@@ -86,4 +86,21 @@ defmodule Realtime.Telemetry.LoggerTest do
       assert {:noreply, []} = TelemetryLogger.handle_info(:unexpected, [])
     end
   end
+
+  describe "phoenix endpoint stop events" do
+    test "logs the request method and path alongside the response" do
+      start_link_supervised!({TelemetryLogger, handler_id: "telemetry-logger-test"})
+
+      conn = %Plug.Conn{method: "GET", request_path: "/api/tenants/abc/health", status: 401, state: :sent}
+
+      log =
+        capture_log(fn ->
+          :telemetry.execute([:phoenix, :endpoint, :stop], %{duration: 1_000_000}, %{conn: conn, options: []})
+        end)
+
+      assert log =~ "GET /api/tenants/abc/health"
+      assert log =~ "401"
+      refute log =~ ~r/^Sent /
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


test/realtime/telemetry/logger_test.exs
```diff
@@ -86,4 +86,21 @@ defmodule Realtime.Telemetry.LoggerTest do
...
+      start_link_supervised!({TelemetryLogger, handler_id: "telemetry-logger-test"})
+
+      conn = %Plug.Conn{method: "GET", request_path: "/api/tenants/abc/health", status: 401, state: :sent}
+
+      log =
+        capture_log(fn ->
+          :telemetry.execute([:phoenix, :endpoint, :stop], %{duration: 1_000_000}, %{conn: conn, options: []})
+        end)
+
+      assert log =~ "GET /api/tenants/abc/health"
```
edgurgel: Should the assertion here check for the exact new log?

`asssert log =~ "GET /api/tenants/abc/health - Sent 401 in "`

WDYT?

PragTob:
Bunch of tests are failing 👀  They refute producing a log for some actions. I suppose we probably have some log filter that now doesn't take effect any more.

I wonder if rather than adding our own logs we could modify the log format emitted by phoenix to achieve the intent of the PR 🤔

mandar1045:
I went through the code and found why the CI tests are failing. It looks like `DISABLE_HEALTHCHECK_LOGGING=true` makes `Endpoint.log_level` return `false`, but this case is not handled right now.

Maybe we can fix it by checking `metadata[:options][:log]` against `conn` in `handle_event/4` and return `:ok` when it is `false`. I think this might fix the CI failures.

PragTob:
@mandar1045 yup, that's the way. We need to respect `log_level` as setup here: https://github.com/supabase/realtime/blob/ee315caec3a1388cafd1889930d5d01820245e45/lib/realtime_web/endpoint.ex#L72-L88

that also removes some of the code there.

mandar1045:
Exactly. Since `Endpoint.log_level` already handles the status levels (`:error` / `:warning`), I think we can just remove the `cond` block and call `Logger.log(level, message)` directly.

YorviArias:
Made changes to this pr based on the comments.

PragTob:
pg17 partition 2 were flakies, rerunning

PragTob (approved):
👌 

@YorviArias looks great to me. I did the tiniest micro adjustment but should be good to go.

PR #2192, opened 2026-09-10
Chore/parallelize e2e tests

seangeo:

Adds support for running the e2e test suite in parallel.

Previously the test suite would only run sequentially across all test suites. This meant the suite would take about 80 seconds to run.

## What is the new behavior?

This change adds a `--parallel` command line argument. When this is set the test suites will run in parallel. By default, parallelism happens at the suite level, so multiple suites will run in parallel, but the test cases within a suite will run sequentially. Specific suites can override this by setting `parallel: true` in their `SuiteDescriptor`, when this is set and the `--parallel` flag is passed in, the test cases within the suite will also run in parallel. This has been set on the `postgres-changes`, `postgres-changes-filters`, and `authorisation` suites, since they are the largest or slowest.

The result is now that with the `--parallel` flag set and `--wait-time` set to 10 milliseconds the suite will complete in ~5s locally.

I've gone with an "opt-in to parallelism" approach here to avoid breaking things, especially the CUFs, behaviour will be the same before if you don't pass in `--parallel`.

## Additional context

The main changes across each suite in this PR were to ensure that each suite uses its own Supabase client instance.

```bash
$  mise run e2e -- --parallel --wait-time 10
<snip>
52 passed, 0 failed  total 5.86s
```

Closes REAL-1037


test/e2e/src/runner.ts
```diff
@@ -79,15 +77,37 @@ export type SuiteDescriptor = {
   // (e.g. "broadcast extension", "authorization check") differ from their --test category key.
   label: string;
   needsDb: boolean;
+  // Defaults to false: a suite's own tests stay sequential even under --parallel until
+  // someone has actually checked it's safe (e.g. load suites that measure
+  // throughput/latency may not want concurrent noise from their own tests) and set this
+  // to true. Suites don't need to know about this at all — a suite's `run()` body
+  // always just writes `await test(...)` in sequence; whether that call blocks is
+  // decided here, not by the suite.
+  parallel?: boolean;
```
PragTob: so yeah that is my concern here, but not just load suites with themselves but also load suite vs. another load suite or just a load suite running at the same time as a normal test.

Sounds like if we run them in parallel with anything we might unnecessarily create noisy neighbours for ourselves.

Maybe this just works and I'm being overcautious. My instinct would be run all the load ones with nothing in parallel, maybe eve in their own job.

PragTob: other thing: naming this parameter the same as the CLI argument is a bit confusing to me. Esp. when mixing it with expectations from elixir (where a suite is only run in parallel when `async: true`) as any `--parallel` already means all suits run in parallel. 

so maybe something like `parallel_tests` or something for the suite argument _might_ help :)

PragTob (approved):
🚀 

Looks good and is a big improvement 👍  Evidently it works 👌 

I do have a couple of concerns though that we may address now, later or say they're not important enough 😁 

* parallelism right now seems unbound (i.e. equivalent of `--max-cases`), which I think it would be good to be able to limit
* `--parallel` vs. `parallel: true` use the same word to mean different things (run suites in parallel vs. run tests inside a suite in parallel) which is a bit confusing and I think we could opt for other wording (there's also an inline comment about this)
* we seem to be running all against the same database, which is legitimately amazing that it works/the tests are always keyed against the same thing/don't seem to interfere
*  I don't believe the "load" scenarios should run parallel to anything, esp. not in parallel to other load scenarios - i.e. an opt out of parallel behavior in general might be good to have imo (which would then be good usage of `parallel: false` 😁 ) - there's also an inline comment on that

As things work right now, easy to do as followups (or not, as it works - which kinda surprises me for the load ones tbh 😅 )

Image

seangeo:
Thanks for the review @PragTob. Yeah I was very surprised that it Just Worked too. I was expecting to have to resolve a bunch of issues related to tests stomping on each other's state in the DB, but it worked first time. I even went through and intentionally broke tests just to make sure that the change wasn't no-oping everything.

On your suggestion about max tasks, I think this would introduce quite a lot of complexity, the runner would need to track how many tests are inflight and you couldn't using things like `Promise.all` without jumping through a lot of hoops. Given we haven't seen any issues maybe it's not worth it for now. I also wondering if the Node runtime is throttling it a bit in helpful way too, since things aren't actually running in parallel, it's just single threaded async.

I'll see if I can fix that naming issue.

seangeo:
Two new commits here, separated out the load tests so they're never run in parallel and renamed the `parallel` flag on SuiteDescriptor to `runCasesInParallel` - seems clearer.

PragTob (approved):
👌 

Great point about the node run time, so it actually is concurrent not parallel (ok the SQL in the DB is parallel) (concurrent vs. parallel is one of my pet peeves 😇 ).

Thanks heaps!

Image

PragTob (approved):

PR #2193, opened 2026-09-10
Chore/e2e tests GitHub action

seangeo:

Add a CI job to run e2e tests.

e2e tests aren't run as part of CI.

## What is the new behavior?

Runs the e2e tests, this PR will show us if they work in CI or not.

## Additional context

I had to change how policies were setup in text fixtures because they were causing deadlocks by being run in a parallel on a fresh database. 

This is also not doing any caching of docker images or build artifacts, we might want that.

Part of REAL-1037


PragTob (approved):
🚀 

Image

.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Install bun and the Supabase CLI via mise
+        uses: jdx/mise-action@1648a7812b9aeae629881980618f079932869151 # v4.0.1
+        with:
+          working_directory: test/e2e
+          install_args: "bun supabase"
+
+      - name: Run e2e suite against a Realtime image built from this checkout
+        working-directory: test/e2e
+        run: mise run e2e-local -- --test functional --parallel --wait-time 10
```
PragTob, line 45: not running the load tests because of CI noise? 🤔

edgurgel: yeah probably worth running all here I think?

PragTob:
Looks like `parallel` may have caused a timeout due to strain here:

```
Failed:
  ✗ postgres-changes-filters / gt: delivers row greater than the value
    Timed out waiting for gt event (8022ms)
  ✗ postgres-changes-filters / imatch: matches the regex case-insensitively
    Timed out waiting for imatch event (8028ms)
```

.github/workflows/e2e_tests.yml
seangeo: I can change this now that I’ve separated the load tests to sequential.

PragTob: Ah nice! I wasn't sure if that was already here (I'm lacking exp with stacked PRs 😅 )

seangeo:
> Looks like `parallel` may have caused a timeout due to strain here:
> 
> ```
> Failed:
>   ✗ postgres-changes-filters / gt: delivers row greater than the value
>     Timed out waiting for gt event (8022ms)
>   ✗ postgres-changes-filters / imatch: matches the regex case-insensitively
>     Timed out waiting for imatch event (8028ms)
> ```

I’ve seen that in series too. 

I wonder if these should retry automatically?

PragTob:
@seangeo maybe, but then sounds to me like something may be wrong with the test (but usually triggers more often on CI/noisy env). I'd hope removing parallel would make it less likely but that's me guessing 🤷  Time wise should still be ok on CI for now (as we have slower jobs).

I guess if a retry is easy we might do it here, as fixing the tests might just delay this a bit further/we can do it after it landed. If it retried it'd be awesome if it added itself to the flaky report but that's also for the future 😅

edgurgel:
Hmm weird that these 2 tests seem to fail easily...

seangeo:
I'll see if I can figure this out, otherwise I'll drop the parallel in CI.

seangeo:
Ok, after a bit of whack-a-mole I've added the following two changes:

1. Introduced a way to isolate tests that use the pg_changes table but adding an opt-in run_id column. Some the tests have negation or range filters, which means they would get changes from any other test that writes to the table. 
2. Made it so the test suites that use pg_changes aren't run in parallel to one another. This seemed to be creating a bit of contention on the tests and causing timeouts.

After these changes, we have the tests passing three times in a row. :shipit: ?

PragTob (approved):
@seangeo in general sounds good to me! My only question is, since the change to not run the `pgChanges` jobs concurrently to each other (imo opting them out in total of parallel would also be ok) do we still need the other changes?

I.e. adding the `run_id` column makes me feel a bit uneasy (as it's different from production then) and the `isolated` helper you need to think of.

If running them sequentially alone is enough (which it should 🤔 ) I'd prefer that but I also trust your judgement! And in any case, it's easy to change.

Image

seangeo:
Thanks @PragTob, good question. The way I understand it, the pg_changes table that this adds run_id to is the equivalent of a tenant’s table used for the test suite so when this is run for our CUFs tests for example, it should add the run_id column there. I could be wrong, maybe @edgurgel or @filipecabaco can chime in on my understanding. 

I think for correctness, if we want to be able run any parallel tests, we need to isolate their database state, this is _one_ way to do it. 

Also not that the test cases for pg_changes themselves do still run in parallel, just the suites don’t. We could not run in parallel in CI, but then in will take probably 10-15 times longer for the suite to run. 

I kinda do like the idea that we can test in parallel, feels like something we should be testing too, however as long as we can do it consistently in all environments and without introducing flakiness.

PragTob:
@seangeo thanks, sounds good!

With not running in parallel, I just meant the specific pg-changes tests - giving them a `parallel: false`/equivalent of `async: false` - as it's not that many I think/hope it wouldn't increase the test run time 10-15x. Just feels more generally applicable than the more specific approach right now, which is more akin to the async groups in ExUnit (which technically is the better solution though).

For `run_id` column - it just feels off to me to have database structure different from what it should be on prod in test. Maybe that's what we do here and it's perfectly good - I can't think of a scenario where it'd break, just a feeling. Ideally I think I'd rather run them against separate databases than adding a column - but that's more work than we should do in this PR as I also really want this to land as it's all closing an important gap (running our e2e tests in our CI). If we want to do that, we can easily patch it later.

and yeah again, I'm good with this going in as is :)

PR #2195, opened 2026-09-10
chore(tests): Wait for muster to be ready + increase timeout

PragTob:
After test processes killed, distributed_realtime_channel_test.exs was the most prominent failure both yesterday and today https://supabase.slack.com/archives/C070A8D2DKP/p1789027852388999

This should help by first waiting that Muster is ready even if the environment is busy and then to also wait for the message a little longer - which only has a negative effect in the negative case so should be an ok tradeoff.

If this is deemed good, might make the `wait_for_muster` into a more general helper that we might apply elsewhere as well.

### DIFF
--- a/test/integration/distributed_realtime_channel_test.exs
+++ b/test/integration/distributed_realtime_channel_test.exs
@@ -4,6 +4,7 @@ defmodule Realtime.Integration.DistributedRealtimeChannelTest do
     async: false,
     parameterize: [%{serializer: Phoenix.Socket.V1.JSONSerializer}, %{serializer: RealtimeWeb.Socket.V2Serializer}]
 
+  alias Forum.Muster
   alias Phoenix.Socket.Message
 
   alias Realtime.Tenants.Connect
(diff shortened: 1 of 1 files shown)


edgurgel (approved):
LGTM

test/integration/distributed_realtime_channel_test.exs
```diff
@@ -44,7 +48,24 @@ defmodule Realtime.Integration.DistributedRealtimeChannelTest do
...
 
-      assert_receive %Message{event: "broadcast", payload: ^payload, topic: ^topic}, 2000
+      assert_receive %Message{event: "broadcast", payload: ^payload, topic: ^topic}, 5000
     end
   end
+
+  # Broadcasts route through Muster's region ring, so wait for the local and
+  # peer node to both consider it :ready and agree on the same ring view before
+  # sending anything cross-node.
+  defp wait_for_muster_ready(node, region) do
```
edgurgel, line 58: Yeah maybe later on we want to have some form of this done by `Clustered` as a wait to sync the nodes before tests start?

PragTob: Didn't think of clustered but that may be a great place. I have a couple of other tests in mind where this may be helpful but first wanted to get your input 😁

PR #2197, opened 2026-09-10
fix: sync presence state when a track enables presence

### LINKED ISSUE #2196: Enabling presence with a track message never sends presence_state, so already-tracked members stay invisible
## Bug report

A client that enables presence by sending a `track` message — rather than by setting `config.presence.enabled` at join — never receives `presence_state`. It only ever sees `presence_diff` events from that point on, so every member that was already tracked on the topic stays invisible to it, permanently.

`RealtimeChannel.join/3` sends `:sync_presence` only when presence was already enabled at join time:

```elixir
# Start presence and add user if presence is enabled
if presence_enabled?, do: send(self(), :sync_presence)
```

Nothing sends `:sync_presence` when `PresenceHandler` later flips `presence_enabled?` to `true` on a track, even though that is a supported path (added in #1527, "this will enable users to enable presence later in the flow by sending a track message").

This is the default path rather than an edge case: `tenants.presence_enabled` defaults to `false` (`lib/realtime/api/tenant.ex:35`, `priv/repo/migrations/20260304000000_add_presence_enabled_to_tenants.exs`), so any client that calls `track()` without also passing `config.presence.enabled: true` lands here.

### To Reproduce

1. Client A joins `realtime:room` with `config.presence.enabled: true` and tracks itself under key `early`. A receives `presence_state`, then a `presence_diff` for its own join.
2. Client B joins the same topic with `config.presence.enabled: false`. B correctly receives no `presence_state`.
3. B sends a `presence` `track` message. Presence is now enabled for B.
4. B receives a `presence_diff` containing only its own join. B never receives `presence_state`, and never learns that `early` is present.

Added as a regression test in `test/integration/rt_channel/presence_test.exs`; on `main` at `ee315cae` it fails with:

```
Assertion failed, no matching message after 500ms
code: assert_receive {:late, %Message{event: "presence_state", payload: state, topic: ^topic}}
```

### Expected behavior

When a track enables presence for a socket, that socket receives the current `presence_state`, exactly as it would have at join had presence been enabled there. Enabling presence at either point should converge on the same view of the topic.

### Actual behavior

No `presence_state` is sent. B's presence view contains only the members that happen to join or change *after* its track, and it has no way to discover the ones already there — a fresh join with `presence.enabled: true` is the only recovery.

### Root cause

`lib/realtime_web/channels/realtime_channel/presence_handler.ex`, `track/2`. On a successful `Presence.track/4` it assigns `presence_enabled?: true` but does not sync:

```elixir
     {:ok, _} <- Presence.track(self(), tenant_topic, presence_key, payload) do
  socket =
    socket
    |> assign(:presence_enabled?, true)
    |> assign(:presence_track_payload, payload)

  {:ok, socket}
```

The one-shot `send(self(), :sync_presence)` lives only in `join/3` behind `if presence_enabled?`, so the auto-enable path skips the initial snapshot entirely.

### Impact

Any presence UI built on `track()` without explicit join-time presence config shows an incomplete roster: the first client into a topic sees everyone who arrives later, while every later client sees only itself plus subsequent arrivals. It is order-dependent and silent — no error, no log line — and it does not self-heal, because `presence_state` is only ever sent at join.

### System information

- Version of realtime: `main` @ `ee315cae`
- Reproduced locally with `mix test test/integration/rt_channel/presence_test.exs` on both serializers (V1 JSON and V2), Elixir 1.20.4 / OTP 29, supabase/postgres:15.14.1.167

I have a fix and a regression test ready and will open a PR referencing this issue.

hsusul:
Fixes #2196

## What

`RealtimeChannel.join/3` sends `:sync_presence` only when presence is already enabled at join:

```elixir
# Start presence and add user if presence is enabled
if presence_enabled?, do: send(self(), :sync_presence)
```

Nothing sends it when `PresenceHandler.track/2` later flips `presence_enabled?` to `true`, even though enabling presence with a track message is a supported path (#1527). Such a client never receives `presence_state` — only the diffs that follow — so every member already tracked on the topic stays invisible to it, with no way to recover short of re-joining with `config.presence.enabled: true`.

It is the default path rather than an edge case: `tenants.presence_enabled` defaults to `false`, so any client that calls `track()` without explicit join-time presence config lands here. The result is order-dependent and silent — the first client into a topic sees everyone, every later client sees only itself plus subsequent arrivals.

## Fix

Sync when the track is what enables presence:

```elixir
if !socket.assigns.presence_enabled?, do: send(self(), :sync_presence)
```

Reusing the existing `:sync_presence` handler rather than pushing directly keeps both existing gates: `PresenceHandler.sync/1` still drops the state for a private socket without `presence.read`, and still goes through `limit_presence_event/1`. Because the message is handled after the current `handle_in` returns, the snapshot is taken after `Presence.track/4`, so it includes the tracking member itself and cannot race ahead of its own join diff.

## Tests

`test/integration/rt_channel/presence_test.exs`:

- **New** — "presence automatically enabled by track receives the members already tracked": an early member joins with presence enabled and tracks; a late member joins with presence disabled, gets no `presence_state`, then tracks — and must now receive a `presence_state` containing the early member.
- **Updated** — "presence_diff gate holds when presence is auto-enabled via track" ends in a catch-all `refute_receive _any`, which the new `presence_state` now trips for the authorized member. I made the new behavior explicit rather than widening the refute, and added the security-relevant half that was implicit before: the member **denied** `presence.read` must receive neither the diff nor the state its own track syncs.

Both fail on `main` @ `ee315cae` (4 failures across the two serializer parameterizations), for the intended reason:

```
Assertion failed, no matching message after 500ms
code: assert_receive {:late, %Message{event: "presence_state", payload: state, topic: ^topic}}
```

Green with the fix.

## Validation

Run locally on Elixir 1.20.4 / OTP 29 against `supabase/postgres:15.14.1.167` (the repo pins 1.19.5 / OTP 28.5.0.4 — CI will confirm on the pinned toolchain and the full postgres matrix):

- `mix test test/integration/rt_channel/presence_test.exs test/realtime_web/channels/realtime_channel/presence_handler_test.exs` — 86 passed (both serializers)
- `mix test test/integration/rt_channel test/realtime_web/channels` — 595 passed
- `mix compile --force --warnings-as-errors` — no warnings from the changed files (the checkout emits a few pre-existing warnings under 1.20 that do not appear on the pinned 1.19)
- `mix format --check-formatted`, `mix credo` (no issues), `mix deps.unlock --check-unused`, `mix sobelow`, `mix dialyzer` (passed successfully), `git diff --check` — all clean

Not run locally: the full 4-partition × 5-postgres matrix — left to CI.

### DIFF
--- a/lib/realtime_web/channels/realtime_channel/presence_handler.ex
+++ b/lib/realtime_web/channels/realtime_channel/presence_handler.ex
@@ -153,6 +153,11 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
          _ <- RealtimeWeb.TenantBroadcaster.collect_payload_size(socket.assigns.tenant, payload, :presence),
          :ok <- limit_presence_event(socket),
          {:ok, _} <- Presence.track(self(), tenant_topic, presence_key, payload) do
+      # This track is what enables presence for the socket, so it never got the join-time
+      # presence_state and only sees diffs from here on. Sync now, or the members tracked
+      # before this point stay invisible to this client.
+      if !socket.assigns.presence_enabled?, do: send(self(), :sync_presence)
+
       socket =
         socket
         |> assign(:presence_enabled?, true)
--- a/test/integration/rt_channel/presence_test.exs
+++ b/test/integration/rt_channel/presence_test.exs
@@ -98,6 +98,44 @@ defmodule Realtime.Integration.RtChannel.PresenceTest do
       assert get_in(join_payload, ["name"]) == payload.payload.name
       assert get_in(join_payload, ["t"]) == payload.payload.t
     end
+
+    test "presence automatically enabled by track receives the members already tracked", %{
+      tenant: tenant,
+      topic: topic,
+      serializer: serializer
+    } do
+      parent = self()
+      # Forward the late joiner's frames tagged so they don't collide with the main mailbox.
+      late_inbox = spawn_link(fn -> forward_frames(parent, :late) end)
+
+      topic = "realtime:#{topic}"
+
+      # An early member joins with presence enabled and tracks itself.
+      {early, _} = get_connection(tenant, serializer)
+      WebsocketClient.join(early, topic, %{config: %{presence: %{key: "early", enabled: true}, private: false}})
+      assert_receive %Message{event: "phx_reply", payload: %{"status" => "ok"}, topic: ^topic}, 500
+      assert_receive %Message{event: "presence_state", topic: ^topic}, 500
+
+      early_payload = %{type: "presence", event: "TRACK", payload: %{name: "early"}}
+      WebsocketClient.send_event(early, topic, "presence", early_payload)
+      assert_receive %Message{event: "presence_diff", payload: %{"joins" => %{"early" => _}}, topic: ^topic}, 500
+
+      # A late member joins with presence disabled, so it gets no presence_state at join.
+      {:ok, late_token} = token_valid(tenant, "anon", %{})
+
+      {:ok, late} =
+        WebsocketClient.connect(late_inbox, uri(tenant, serializer), serializer, [{"x-api-key", late_token}])
+
+      WebsocketClient.join(late, topic, %{config: %{presence: %{key: "late", enabled: false}, private: false}})
+      assert_receive {:late, %Message{event: "phx_reply", payload: %{"status" => "ok"}, topic: ^topic}}, 500
+      refute_receive {:late, %Message{event: "presence_state", topic: ^topic}}, 500
+
+      # Tracking enables presence for the late member, which must now learn who is already present.
+      WebsocketClient.send_event(late, topic, "presence", %{type: "presence", event: "TRACK", payload: %{name: "late"}})
+
+      assert_receive {:late, %Message{event: "presence_state", payload: state, topic: ^topic}}, 500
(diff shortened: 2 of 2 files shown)


lib/realtime_web/channels/realtime_channel/presence_handler.ex
```diff
@@ -153,6 +153,11 @@ defmodule RealtimeWeb.RealtimeChannel.PresenceHandler do
          _ <- RealtimeWeb.TenantBroadcaster.collect_payload_size(socket.assigns.tenant, payload, :presence),
          :ok <- limit_presence_event(socket),
          {:ok, _} <- Presence.track(self(), tenant_topic, presence_key, payload) do
+      # This track is what enables presence for the socket, so it never got the join-time
+      # presence_state and only sees diffs from here on. Sync now, or the members tracked
+      # before this point stay invisible to this client.
+      if !socket.assigns.presence_enabled?, do: send(self(), :sync_presence)
```
edgurgel, line 159: Can we do this from `RealtimeChannel` instead? Because this is a message that RealtimeChannel processes and `PresenceHandler` feels like the wrong place to assume what process is running this.

edgurgel (changes requested):

edgurgel (approved):
LGTM! Thanks once again ❤️

PR #2198, opened 2026-09-11
chore: use docker for schema & dump generation

edgurgel:

This way we don't need to have pg-dump for a specific major version etc. We just use the tools from the image that we are extracting the dump from.

It's also easier to run it locally. We won't __need__ to push to GitHub to get the new dumps. It's now optional.

### DIFF
--- a/.github/workflows/update-tenant-db-snapshots.yml
+++ b/.github/workflows/update-tenant-db-snapshots.yml
@@ -24,24 +24,26 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  export-snapshots:
-    name: Export schema & dump (pg${{ matrix.pg_major }})
+  update-snapshots:
+    name: Export and commit schema & dumps
     runs-on: blacksmith-2vcpu-ubuntu-2404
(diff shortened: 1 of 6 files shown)


seangeo (approved):
Looks good to me.

I was wondering if we could simplify this even more, I'm assuming the container it's connecting to is something we control in the dev and ci environments? And thinking this could almost be a script on the container that you run, which does all this. But probably not worth it.

This change makes sense. Should I try running it locally or is CI already running it?

edgurgel:
> Looks good to me.
> 
> I was wondering if we could simplify this even more, I'm assuming the container it's connecting to is something we control in the dev and ci environments? And thinking this could almost be a script on the container that you run, which does all this. But probably not worth it.
> 

Hmm fair point. This mix task could be a simple bash script really 🤔 . Might play with this.

> This change makes sense. Should I try running it locally or is CI already running it?

CI runs it on `.github/workflows/update-tenant-db-snapshots.yml`

edgurgel:
@seangeo updated to use a bash script now!

Edit:

The CI check seems to be working as expected: 

chore__use_docker_for_schema___dump_generation_by_edgurgel_·_Pull_Request__2198_·_supabase_realtime

seangeo (approved):
LGTM!

PR #2199, opened 2026-09-11
chore(tests): Attempt to fix flaky Node regions test

PragTob:
https://github.com/supabase/realtime/actions/runs/34439478592/job/102751321026

```
  5) test all_node_regions/0 with no other nodes, returns my region only (Realtime.NodesTest)
     test/realtime/nodes_test.exs:36
     Assertion with == failed
     code:  assert Nodes.all_node_regions() == ["us-east-1"]
     left:  []
     right: ["us-east-1"]
     stacktrace:
       test/realtime/nodes_test.exs:37: (test)
```

I'm honestly unsure about the mechanism at play here, but deemed an `eventually` a worthy try and if not, we at least learn that under some circumstances it may be "more" broken and have that aid our investigation.

### DIFF
--- a/test/realtime/nodes_test.exs
+++ b/test/realtime/nodes_test.exs
@@ -34,7 +34,9 @@ defmodule Realtime.NodesTest do
     end
 
     test "with no other nodes, returns my region only" do
-      assert Nodes.all_node_regions() == ["us-east-1"]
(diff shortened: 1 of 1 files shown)


edgurgel (approved):
Yeah it could be some timing issue but I'm not 100% sure how as the running node should be good to go before tests start 🤔

PR #2203, opened 2026-09-14
chore(devtools): Improve local Realtime Inspector functionality and DX

seangeo:

Gets the Realtime Inspector working locally and improves the DX.

The Realtime Inspector wasn't working locally because it was expecting the socket to be running on a different path, `/realtime/v1/websocket` not `/socket/websocket`, since this is the path that Supabase-js expects.

## What is the new behavior?

Rather than getting into the guts of the supabase-js client library and trying to override the path it uses, or making changes to the inspector to use the RealtimeClient directly, I opted for mounting the socket on the same path as the gateway mounts it when in the dev environment. I'm open to other opinions here...

I've also added a mise task to get a token for local dev, and one to open the inspector with everything already filled out, using `mise run dev-inspector`.

## Additional context

Closes REAL-1070.


config/dev.exs
```diff
@@ -44,6 +44,12 @@ config :realtime, RealtimeWeb.Endpoint,
   debug_errors: true,
   code_reloader: true,
   check_origin: false,
+  # supabase-js's createClient() always builds the realtime websocket URL as
+  # `${url}/realtime/v1/websocket`, the path a full Supabase stack or hosted project serves
+  # behind a gateway. The bare dev server only serves `/socket`; this mounts the same socket at
+  # that path too, so local tooling built on supabase-js (like the Inspector) can connect
+  # straight to `mise run dev` without a gateway in front.
+  local_gateway_socket: true,
```
leandrocp: Not a big concern but since `:local_gateway_socket` is not a built-in Phoenix Endpoint option I'd consider moving it out of `config :realtime, RealtimeWeb.Endpoint` to be more clear and easier to update Phoenix versions.

lib/mix/tasks/realtime.gen_token.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Mix.Tasks.Realtime.GenToken do
```
leandrocp, line 1: This is a time saver!

mise.toml
```diff
@@ -41,6 +41,26 @@ description = "Start another dev server in a second region (orange)"
...
 
+[tasks.dev-inspector]
+description = "Open the Realtime Inspector prefilled for TENANT"
+dir = "{{cwd}}"
+run = '''
+set -eu
+tenant=${TENANT:-realtime-dev}
+port=${PORT:-4000}
+host="http://${tenant}.localhost:${port}"
+token=$(mix realtime.gen_token)
```
leandrocp: I think this might fail if the project needs recompilation, for example:

```shell
~/code/supabase/realtime/trees/main seangeoghegan/real-1070-support-local-realtime-inspector-connections-and-publishable
❯ mix realtime.gen_token
==> forum
Generated forum app
==> realtime
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE3ODk0OTQwNzAsInJvbGUiOiJhdXRoZW50aWNhdGVkIn0.MBYyiES_jLo4VrAjOPmB6CwftnCXl26_HbSarnmhjEk
```

seangeo: Great catch!

config/dev.exs
seangeo: Fixed!

mise.toml
seangeo: Fixed this by just grabbing the last line.

leandrocp (approved):
🎉

edgurgel:
I think this solution is fine but I would love to know if we couldn't just use `RealtimeClient` directly with whatever URL we want (which won't force us to have the same supabase path)

seangeo:
@edgurgel Tried that in [55b186c](https://github.com/supabase/realtime/pull/2203/commits/55b186ca548b68e444ccf171e9ff1a4d4836a91d), works well both locally and against a hosted instance.

lib/realtime_web/endpoint.ex
```diff
@@ -13,33 +13,37 @@ defmodule RealtimeWeb.Endpoint do
...
-        {RealtimeWeb.Socket.V2Serializer, "~> 2.0.0"}
-      ]
-    ],
-    longpoll: [
-      connect_info: [:peer_data, :uri, :x_headers],
-      serializer: [
-        {Phoenix.Socket.V1.JSONSerializer, "~> 1.0.0"},
-        {Phoenix.Socket.V2.JSONSerializer, "~> 2.0.0"}
-      ]
+  @user_socket_websocket_opts [
```
edgurgel: Can we then revert all of this now?

edgurgel (approved):
Tested with both localhost and remote on staging both worked. Nice one

lib/realtime_web/endpoint.ex
seangeo: Done!

PR #2210, opened 2026-09-14
fix: `postgres` can grant on `realtime.messages`

leandrocp:
Closes REAL-1074

Allow users to grant select and insert on realtime.messages on the dashboard (postgres).


PragTob (approved):
Again, best I know this sounds good :) 

Image

PR #2212, opened 2026-09-14
fix: keep Connect alive when a user rejoins during idle shutdown

### LINKED ISSUE #2211: Connect tears down the tenant database pool when a user reconnects during the idle grace interval
## Bug report

After a tenant has been idle long enough for Connect's user-count bucket to fill with zeros, Connect schedules `:shutdown_no_connected_users` one interval later and **stops sampling** `UsersCounter`. A client that reconnects during that last interval is counted, but Connect still shuts down.

That kills the tenant database pool and the `realtime.messages` replication connection under a live websocket. Database broadcasts (`realtime.send` / trigger-backed Broadcast) then stay down until something starts Connect again — the next private join/broadcast, or `Realtime.Tenants.Reconnector` which only runs every 5 minutes (#2014).

### To Reproduce

1. Start Connect for a tenant with no connected users.
2. Let the idle checker collect 11 consecutive zero samples (`@connected_users_bucket_shutdown`). At the default 60s interval that is ~11 minutes of idle.
3. Connect schedules `:shutdown_no_connected_users` one interval later and does not schedule another `:check_connected_users`.
4. A client joins a channel (or `UsersCounter.add/2` runs). `UsersCounter.tenant_users/1` is now `> 0`.
5. `:shutdown_no_connected_users` fires.

A regression in `test/realtime/tenants/connect_test.exs` fails on `main` at `7e76185c` with:

```
code: refute_process_down(pid, 300)
Unexpectedly received message {:DOWN, #Reference<...>, :process, #PID<...>, :shutdown}
```

The log at that moment is `Tenant has no connected users, database connection will be terminated` even though a user was added before the shutdown message was delivered.

### Expected behavior

Connect should tear down only when the tenant still has no connected users. A user who joined during the grace interval should keep the pool and messages replication running, and idle sampling should resume.

This matches the contract from #701 ("if that tenant has any users and if not, we kill the database connection") and from Reconnector (#2014): Connect dying while websockets are still open is a failure mode the server already tries to heal.

### Actual behavior

Shutdown does not look at `UsersCounter`. The process stops with `:shutdown`, the pool dies, and the log claims there are no connected users.

Existing tests only cover "always idle → stop" and "always has users → stay". They do not cover a rejoin during the final grace interval.

### Root cause

[`lib/realtime/tenants/connect.ex`](https://github.com/supabase/realtime/blob/7e76185c1e3142ace9e2bbc6f980ea6a02ec330a/lib/realtime/tenants/connect.ex#L558-L587)

Once the 11-sample all-zero bucket matches, `send_connected_user_check_message/2` schedules a blind `:shutdown_no_connected_users` and does not reschedule `:check_connected_users`:

```elixir
defp send_connected_user_check_message(
       @connected_users_bucket_shutdown,
       check_connected_user_interval
     ) do
  Process.send_after(self(), :shutdown_no_connected_users, check_connected_user_interval)
end
```

```elixir
def handle_info(:shutdown_no_connected_users, state) do
  Logger.info("Tenant has no connected users, database connection will be terminated")
  {:stop, :shutdown, state}
end
```

The extra interval looks like a confirmation delay, but it never samples again. Channel join already calls `UsersCounter.add(transport_pid, tenant_id)` before the socket is fully up, so this is a normal reconnect path, not a synthetic race.

### Impact

A client that comes back just after a long idle (mobile backgrounding, laptop sleep, tab restore) can sit on a healthy-looking websocket whose database broadcasts no longer flow, for up to 5 minutes until Reconnector, or until some other operation restarts Connect. Presence and client-to-client Broadcast still work because they do not need the tenant pool; `realtime.send` / private Broadcast-from-DB does.

### System information

- Version of realtime: `main` @ `7e76185c`
- Reproduced locally with `mix test test/realtime/tenants/connect_test.exs:413` (Elixir 1.20.4 / OTP 29, supabase/postgres:17.6.1.166)

I have a fix and a regression test ready and will open a PR referencing this issue.

hsusul:
Fixes #2211

## What

Connect's idle checker fills an 11-sample zero bucket, then schedules `:shutdown_no_connected_users` one interval later **and stops sampling** `UsersCounter`. A client that joins during that last interval is counted (`UsersCounter.add/2` on channel join), but the shutdown handler still stops Connect with `:shutdown`.

That tears down the tenant database pool and the `realtime.messages` replication connection under a live websocket. Database broadcasts stay down until something starts Connect again — another DB-touching operation, or `Realtime.Tenants.Reconnector` which only runs every 5 minutes (#2014). The log at that moment is `Tenant has no connected users, database connection will be terminated` even though a user is present.

Idle teardown of a still-empty tenant is unchanged.

## Fix

Recount `UsersCounter.tenant_users/1` when `:shutdown_no_connected_users` fires. If it is still 0, stop as today. If a user joined during the grace interval, resume the check loop with bucket `[count]` instead of stopping.

`connected_users_bucket` must be reset to a list: after the shutdown was scheduled, that field holds the `send_after` timer ref, and resuming checks without replacing it would crash (`ref ++ [count]`).

## Test

`test/realtime/tenants/connect_test.exs` — "does not shut down if a user reconnects during the idle grace interval": drives 11 `:check_connected_users` messages so the all-zero bucket schedules the delayed stop, then `UsersCounter.add/2` and `:shutdown_no_connected_users`, and asserts Connect, the pool, and `:syn` stay up.

It sits next to the existing tests that already pin "always idle → stop" and "always has users → stay".

Red on `main` @ `7e76185c`:

```
1) test Database connection stopping does not shut down if a user reconnects during the idle grace interval
   Unexpectedly received message {:DOWN, #Reference<...>, :process, #PID<...>, :shutdown}
   code: refute_process_down(pid, 300)
```

The log at that moment was `Tenant has no connected users, database connection will be terminated`.

Green with the fix.

## Validation

Run locally on Elixir 1.20.4 / OTP 29 against `supabase/postgres:17.6.1.166` (the repo pins 1.19.5 / OTP 28.5.0.4 — CI will confirm on the pinned toolchain and the full postgres matrix):

- `mix test test/realtime/tenants/connect_test.exs:413` — red on `main`, green after the fix
- Idle-family tests in `connect_test.exs` (always idle / always connected / rejoin during grace / user leaving) — 4 passed; repeated 3x, stable
- `mix test test/realtime/tenants/reconnector_test.exs` — 2 passed
- `mix compile --force --warnings-as-errors` — no warnings from the changed files (the checkout emits a few pre-existing warnings under 1.20 that do not appear on the pinned 1.19)
- `mix format --check-formatted`, `mix credo` (no issues), `mix deps.unlock --check-unused`, `mix sobelow --config .sobelow-conf`, `mix dialyzer` (passed successfully), `git diff --check` — all clean

Not run locally: the full 4-partition × 5-postgres test matrix — left to CI.

Two other `connect_test.exs` cases were unrelated to this change: the replication-pid-stop test passed on `--failed` retry (flake); `handles max_wal_senders by logging the correct operational code` still failed in isolation (`public.test` missing / all replication slots in use) — environmental, present on this machine independently of the idle-shutdown handler.


lib/realtime/tenants/connect.ex
```diff
@@ -381,9 +381,18 @@ defmodule Realtime.Tenants.Connect do
...
+      0 ->
+        Logger.info("Tenant has no connected users, database connection will be terminated")
+        {:stop, :shutdown, state}
+
+      count ->
+        # Sampling stopped when this shutdown was scheduled; resume it for the live user.
+        bucket = [count]
+        send_connected_user_check_message(bucket, state.check_connected_user_interval)
+        {:noreply, %{state | connected_users_bucket: bucket}}
+    end
```
edgurgel, line 395: I wonder if we could call `update_connected_users_bucket`, then check the buckets once again and if still just `0`s we shutdown otherwise call send_connected_user_check_message ?

This way we use the same functions that the recurring check does. WDYT?

edgurgel (changes requested):
Added a comment!

PR #2213, opened 2026-09-14
chore: wait for all events on billable event tests

edgurgel:
We are seeing this sometimes:

```
  1. test presence events presence events (Realtime.Integration.RtChannel.BillableEventsTest)
      Parameters: %{serializer: Phoenix.Socket.V1.JSONSerializer}
      test/integration/rt_channel/billable_events_test.exs:141
      match (=) failed
      code:  assert 7 = get_count([:realtime, :rate_counter, :channel, :presence_events], external_id)
      left:  7
      right: 6
      stacktrace:
        test/integration/rt_channel/billable_events_test.exs:186: (test)
```

Counters are incremented after the message has already been fastlaned to the transport, so the `assert_receive` above can win the race against the increment.  Re-tick until the counters settle rather than sampling them once.


seangeo (approved):
LGTM

PR #2214, opened 2026-09-15
chore: run sequential migrations on dev/test

leandrocp:
Small quality of life improvement so we don't need to generate dumps when changing migrations locally and also safer to test migrations.


lib/realtime/tenants/migrations.ex
```diff
@@ -222,7 +222,7 @@ defmodule Realtime.Tenants.Migrations do
 
         try do
           {applied_count, migrations_executed, source} =
-            if load_db_dump?(migrations_ran, repo) do
+            if __MODULE__.load_db_dump?(migrations_ran, repo) do
```
edgurgel, line 225: Why do you need this to be public? 🤔

edgurgel (approved):

lib/realtime/tenants/migrations.ex
leandrocp: To mimic instead of `Application.put_env` in tests, which I find more explicit and ergonomic because I don't need to revert env values etc, I just set the expectation I need.

PR #2216, opened 2026-09-16
fix: count tenant users only after channel joins succeed

### LINKED ISSUE #2215: Rejected Postgres Changes joins retain a counted tenant user
A rejected Postgres Changes join can leave its websocket transport counted as a connected tenant user until the websocket closes.

Reproduced on upstream `217744c20389fce5c91636ac8273878f1b4f7c73`; the affected code is unchanged on `5a258870f3c7ce0ceccdfbbe64eda56a096eff61`.

### Reproduction

1. Connect a websocket with a valid tenant JWT, without another successfully joined channel.
2. Join a channel using this config while the CDC connection cannot become ready before the timeout:

```json
{"config":{"postgres_changes":[{"event":"INSERT","schema":"public"}],"postgres_changes_options":{"wait":true,"timeout":100}}}
```

3. Observe the join return `PostgresChangesSubscribeTimeout` and keep the websocket open.
4. Check `Realtime.UsersCounter.already_counted?(transport_pid, tenant_id)`: it returns `true` despite the rejected join.

For a deterministic local regression, the existing channel test harness can stub `Extensions.PostgresCdcRls.handle_connect/1` to return `nil`, connect through `UserSocket`, and call `subscribe_and_join/3` with the config above. Assert the timeout error, assert the transport remains alive, then refute `UsersCounter.already_counted?/2`. The final assertion fails on upstream with `Expected false or nil, got true`.

### Expected / actual

Expected: a transport with no successful channel join is not added to the tenant user census by a rejected join.

Actual: the rejected join adds the transport to the census, and the entry remains while the transport stays alive.

### Root cause and impact

`RealtimeChannel.join/3` calls `UsersCounter.add/2` before `await_muster_join/2` and `start_postgres_subscribe/4` succeed. The subscription wait added in #2109 can reject the join after this registration. The census monitors the transport, so rejecting the channel does not remove that registration.

This consumes tenant concurrent-user capacity without a successful channel and can keep the tenant's idle-connection logic seeing an active user. A slow or unavailable CDC connection is enough to trigger it with `wait: true`.

Moving registration into the successful join branch makes the deterministic regression pass; existing successful-join and counter tests also pass.

hsusul:
Closes #2215.

A Postgres Changes join using `wait: true` can time out after its websocket has already been registered in the tenant user census. The rejected channel leaves that transport counted until the websocket closes, consuming concurrent-user capacity and affecting idle detection.

Move `UsersCounter.add/2` after the join gates succeed. The regression keeps the transport alive after a deterministic subscription timeout and verifies that it was not counted.

Validation:

- Regression failed on clean upstream with `Expected false or nil, got true`, then passed with the fix.
- Channel, tracker, and tenant-connection suites: 133 passed.
- Postgres Changes and connection-lifecycle integration suites: 74 passed against local Docker/Postgres.
- Final focused regression and user-counter suite: 8 passed.
- Elixir 1.19.5 / OTP 28 Docker: `MIX_ENV=test mix compile --force --warnings-as-errors` and full `mix format --check-formatted` passed on latest main.
- Credo strict, Sobelow, unused-dependency check, and `git diff --check` passed.

The host's newer Elixir 1.20.4 / OTP 29 reports existing application warnings; the required compile passes on the Elixir 1.19.5 / OTP 28 toolchain. The full CI Postgres matrix, Dialyzer, and end-to-end suite were not run locally.


test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -320,6 +320,25 @@ defmodule RealtimeWeb.RealtimeChannelTest do
               }} = join_waiting_for_postgres_changes(tenant, %{"wait" => true, "timeout" => 100})
     end
 
+    test "does not count a channel whose postgres_changes join is rejected", %{tenant: tenant} do
+      expect(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
+      stub(Extensions.PostgresCdcRls, :handle_connect, fn _ -> nil end)
```
edgurgel, line 325: BTW no point on using expect and stub like this (if you don't mind fixing the above test as well).

Let's just use expect if possible.

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -163,10 +163,12 @@ defmodule RealtimeWeb.RealtimeChannel do
       # Start presence and add user if presence is enabled
       if presence_enabled?, do: send(self(), :sync_presence)
 
-      UsersCounter.add(transport_pid, tenant_id)
-
       with :ok <- await_muster_join(muster_join_task, socket),
            :ok <- start_postgres_subscribe(socket, join, tenant, pg_change_params) do
+        # Count the client only after all join gates have succeeded. A rejected
+        # join leaves the websocket transport available for another channel, so
+        # counting before this point would retain a phantom tenant user.
```
edgurgel, line 170: Can we drop this whole comment, please

edgurgel (changes requested):
Added a couple of comments

edgurgel (approved):

edgurgel:
Thanks ❤️

PR #2217, opened 2026-09-16
fix: remove grant option from tenant schema

leandrocp:
Neither supabase/postgre nor migrations add that grant option and having it on the tenant schema creates unnecessary schema drifts on Tenant Migrations plan.


PragTob (approved):
Looking good from what I understand!

Image

PR #2221, opened 2026-09-17
fix: throttle access token refreshing

edgurgel:

Avoid refreshing tokens so aggressively


config/runtime.exs
```diff
@@ -12,6 +12,7 @@ broadcast_pool_size = Env.get_integer("BROADCAST_POOL_SIZE", 10)
 channel_error_backoff_ms = Env.get_integer("CHANNEL_ERROR_BACKOFF_MS", :timer.seconds(5))
 client_presence_max_calls = Env.get_integer("CLIENT_PRESENCE_MAX_CALLS", 5)
 client_presence_window_ms = Env.get_integer("CLIENT_PRESENCE_WINDOW_MS", 30_000)
+access_token_throttle_ms = Env.get_integer("ACCESS_TOKEN_THROTTLE_MS", :timer.seconds(10))
```
PragTob, line 15: nittiest of nits: I've been converted to [`to_timeout/1`](https://elixir.hexdocs.pm/Kernel.html#to_timeout/1) - ie. `to_timeout(second: 10)`

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -34,7 +34,8 @@ defmodule RealtimeWeb.RealtimeChannel do
   alias RealtimeWeb.RealtimeChannel.PresenceHandler
   alias RealtimeWeb.RealtimeChannel.Tracker
 
-  @confirm_token_ms_interval :timer.minutes(5)
+  # The cap exists because Process.send_after can't take an arbitrary delay
+  @confirm_token_ms_interval :timer.hours(1)
```
PragTob, line 38: I don't get the comment/I'm missing something. `Process.send_after/3` can take any arbitrary delay - what am I missing? 🤔

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -512,21 +522,65 @@ defmodule RealtimeWeb.RealtimeChannel do
...
-      when is_nil(refresh_token) do
+  def handle_in("access_token", %{"access_token" => nil}, socket), do: {:noreply, socket}
+
+  # Already held for the current window, nothing new to verify
+  def handle_in(
+        "access_token",
+        %{"access_token" => refresh_token},
+        %{assigns: %{pending_access_token: pending}} = socket
+      )
+      when refresh_token == pending do
```
PragTob, line 540: Q (here and similar places): Is there any case where someone might want to guess the `pending` token? I.e. any opportunity/impact of a timing attack?

I think not from my limited understanding, but better safe than sorry (hence asking) :)

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -767,6 +815,8 @@ defmodule RealtimeWeb.RealtimeChannel do
     assign(socket, :presence_client_rate_limit, client_rate_limit)
   end
 
+  defp access_token_throttle_ms, do: Application.get_env(:realtime, :access_token_throttle_ms, 10_000)
```
PragTob, line 818: maybe we could also config it for `dev` and then make this a `fetch_env!` to avoid having another default?

lib/realtime_web/channels/realtime_channel.ex
```diff
@@ -512,21 +522,65 @@ defmodule RealtimeWeb.RealtimeChannel do
...
+    Process.send_after(self(), :apply_pending_access_token, remaining_ms)
+    assign(socket, :pending_access_token, refresh_token)
+  end
+
+  defp hold_access_token(socket, refresh_token, _remaining_ms), do: assign(socket, :pending_access_token, refresh_token)
+
+  defp apply_pending_access_token(%{assigns: %{pending_access_token: nil}} = socket), do: {:noreply, socket}
+
+  defp apply_pending_access_token(%{assigns: %{pending_access_token: token}} = socket) do
+    apply_access_token(socket, token, System.monotonic_time(:millisecond))
```
PragTob, line 580: nittiest of nits: I like to define a `now()` helper function or similar, to both have it a bit shorter and make sure everyone is using the same definition of `now()` while both hiding the precision detail and making sure it's consistent :)

ERROR_CODES.md
```diff
@@ -104,4 +104,5 @@ This is the list of operational codes that can help you understand your deployme
 | MetricsPusherTaskExited            | A metrics push task exited unexpectedly before completing                                                                                                                                             |
 | MetricsPusherFailed                | Realtime failed to push metrics to the configured endpoint                                                                                                                                            |
 | MetricsPusherException             | An exception was raised while pushing metrics to the configured endpoint                                                                                                                              |
+| AccessTokenRefreshThrottled        | An `access_token` refresh arrived before `ACCESS_TOKEN_THROTTLE_MS` elapsed since the last verification. The newest token is held and applied when the window closes                                   |
```
PragTob, line 107: great that we have this list and thanks for applying it! 👌

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -1455,6 +1455,115 @@ defmodule RealtimeWeb.RealtimeChannelTest do
     end
   end
 
+  describe "access_token throttling" do
+    # The window is configured for the test env and only ever read, never mutated, so these stay async
```
PragTob, line 1459: 💚

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -1455,6 +1455,115 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+
+      log =
+        capture_log(fn ->
+          push(socket, "access_token", %{"access_token" => second})
+          push(socket, "access_token", %{"access_token" => third})
+
+          # Both were coalesced into the pending slot, newest wins, and neither was verified
+          assigns = Server.socket(socket.channel_pid).assigns
+          assert assigns.access_token == first
+          assert assigns.pending_access_token == third
```
PragTob, line 1490: edit: I am indeed to paranoid. I measured the one below that does 50 pushes and it came in at 500microseconds. We should be fine.

I'm mildly afraid the test _may_ become flakey (if under heavy load the 500ms window passed, before the `third` push reaches the server) - somewhat unrealistic but stranger things have happend 😅  I'd feel safer with a second, but maybe I'm also just paranoid.

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -1455,6 +1455,115 @@ defmodule RealtimeWeb.RealtimeChannelTest do
...
+          exp: System.system_time(:second) - 1,
+          role: "authenticated",
+          sub: "expired"
+        })
+
+      # The state the throttle produces near expiry: current token at its exp, newest refresh still
+      # held. Injected rather than timed, so the window timer plays no part in the outcome.
+      :sys.replace_state(channel_pid, fn socket ->
+        %{socket | assigns: %{socket.assigns | access_token: expired, pending_access_token: fresh}}
+      end)
```
PragTob, line 1554: crafty! 🚀

test/realtime_web/channels/realtime_channel_test.exs
```diff
@@ -1860,6 +1969,30 @@ defmodule RealtimeWeb.RealtimeChannelTest do
     assert_receive {:DOWN, ^ref, :process, ^pid, _reason}
   end
 
+  defp join_public_channel(tenant) do
+    jwt = Generators.generate_jwt_token(tenant)
+    {:ok, %Socket{} = socket} = connect(UserSocket, %{"log_level" => "warning"}, conn_opts(tenant, jwt))
+    subscribe_and_join!(socket, "realtime:test", %{})
+  end
```
PragTob, line 1976: love helpers like this!

PragTob (approved):
Looking great 👌 💚 

Sorry for the many comments - they're mostly praise or the tiniest of nits/ideas that can be ignored safely (and me being too paranoid).

I think the one thing I'd love to have clarified is the "`Process.send_after` can't take an arbitrary delay" comment as I don't get it 😅 

Image

lib/realtime_web/channels/realtime_channel.ex
edgurgel: this:

```
iex(2)> Process.send_after(self(), :hi, 1_00000000000000000000000000000000)
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: exceeds the maximum supported time value

    :erlang.send_after(100000000000000000000000000000000, #PID<0.108.0>, :hi)
    iex:2: (file)
iex(3)> Process.send_after(self(), :hi, to_timeout(week: 52000))
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: exceeds the maximum supported time value

    :erlang.send_after(31449600000000, #PID<0.108.0>, :hi)
    iex:4: (file)
```

If we simply grab the `exp` it can be a very long expiration for a JWT. So using 1 hour is a cheap way to just not have a timer that is set for 100 years in the future or something. 

I guess I could grab the maximum supported time value and use that as max interval?

Or simply not schedule this message if `exp` is longer than 1 year as we can assume no node will be running with a connected websocket for this long? 😆

lib/realtime_web/channels/realtime_channel.ex
edgurgel: I don't think so? Or at least I can't think of a case

lib/realtime_web/channels/realtime_channel.ex
edgurgel: Makes sense!

lib/realtime_web/channels/realtime_channel.ex
edgurgel: yeah good point

test/realtime_web/channels/realtime_channel_test.exs
edgurgel: Yeah I think we can use more of these for some of the realtime channel tests (including integration ones)

ERROR_CODES.md
edgurgel: I usually forget but not this time. @leandrocp is the one that reminds me every single time 🙈

lib/realtime_web/channels/realtime_channel.ex
PragTob: Ah yes, thanks didn't know that! And it has the potential for another docs PR 🤔 

I'm guessing we could call it `max_interval` and take the `min` between that and the configured value, right?

edgurgel: Yeah we already do the `min`! Will call it max_interval then

PR #2226, opened 2026-09-17
fix: look up the tenant on the primary when deleting it

### LINKED ISSUE #2225: DELETE /api/tenants reads the replica and returns 204 without deleting a tenant it hasn't replicated yet
## Bug report

`DELETE /api/tenants/:tenant_id` looks the tenant up on the **read replica**, so deleting a tenant the replica has not caught up with yet answers `204` and deletes nothing.

`TenantController.delete/2` passes the option as `use_replica: false`:

```elixir
with %Tenant{} = tenant <- Api.get_tenant_by_external_id(tenant_id, use_replica: false),
```

but `Api.get_tenant_by_external_id/2` reads `use_replica?` (with a `?`) and defaults it to `true`:

```elixir
use_replica? = Keyword.get(opts, :use_replica?, true)
```

The misspelled key is ignored and the lookup goes to `Replica.replica()`. When the replica returns `nil` (a tenant created moments earlier, or any replication lag), the `with` falls to its `nil -> send_resp(conn, 204, "")` clause: no disconnect, no row deletion, no cache invalidation, no `Connect` shutdown, no replication slot teardown — and the caller is told the delete succeeded.

Every other management call site (`update`, `reload`, `shutdown`, `FeatureFlags`, `Api`) spells it `use_replica?: false`. The misspelling dates to #1633, which introduced the option.

### To Reproduce

Deterministic locally by simulating a replica that hasn't seen the tenant, on `main` @ `a85489e0`:

```elixir
defmodule LaggingReplica do
  def get_by(_, _), do: nil
  def preload(nil, _), do: nil
end

Mimic.stub(Realtime.Repo.Replica, :replica, fn -> LaggingReplica end)
conn = delete(conn, ~p"/api/tenants/#{tenant.external_id}")

conn.status                                                            #=> 204
Realtime.Api.get_tenant_by_external_id(tenant.external_id, use_replica?: false)  #=> %Tenant{} — still there
```

Added as a regression test in `test/realtime_web/controllers/tenant_controller_test.exs`, following the existing "does not read the replica" test for `Api.list_existing_external_ids/1`. On `main` it fails because the request itself reads the replica:

```
** (Mimic.UnexpectedCallError) expected Realtime.Repo.Replica.replica/0 to be called 0 time(s) but it has been called 1 time(s)
```

### Expected behavior

DELETE reads the primary, like the other tenant management endpoints, so a tenant that exists is always deleted and torn down.

### Actual behavior

DELETE reads the replica. Under replication lag it returns `204` while the tenant row, its cached entry, its `Connect` process and its replication slot all survive.

### Impact

A tenant created and then deleted in quick succession — provisioning rollback, test/preview environments, retries — can be reported as deleted while it stays live: clients keep connecting, the tenant database keeps a Realtime connection and replication slot, and the delete isn't retried because it "succeeded". It is silent; nothing is logged on the `nil` path.

### System information

- Version of realtime: `main` @ `a85489e0`
- Reproduced locally with `mix test test/realtime_web/controllers/tenant_controller_test.exs`, Elixir 1.20.4 / OTP 29, supabase/postgres:15.14.1.167

I have a one-line fix and the regression test ready and will open a PR referencing this issue.

hsusul:
Fixes #2225

## What

`TenantController.delete/2` looks the tenant up with `use_replica: false`, but `Api.get_tenant_by_external_id/2` reads `use_replica?` and defaults it to `true`:

```elixir
use_replica? = Keyword.get(opts, :use_replica?, true)
```

The misspelled key is ignored, so DELETE reads the **replica**. For a tenant the replica hasn't caught up with, the lookup returns `nil` and the `with` falls to `nil -> send_resp(conn, 204, "")`: nothing is disconnected, deleted, invalidated, shut down or torn down, and the caller is told it succeeded.

Every other call site (`update`, `reload`, `shutdown`, `FeatureFlags`, `Api`) spells it `use_replica?: false`. The misspelling dates to #1633, which introduced the option.

## Fix

```diff
-    with %Tenant{} = tenant <- Api.get_tenant_by_external_id(tenant_id, use_replica: false),
+    with %Tenant{} = tenant <- Api.get_tenant_by_external_id(tenant_id, use_replica?: false),
```

## Test

`test/realtime_web/controllers/tenant_controller_test.exs`, "delete tenant" → "does not read the replica". It mirrors the existing test of the same name for `Api.list_existing_external_ids/1`: `Mimic.reject(&Realtime.Repo.Replica.replica/0)`, DELETE, then assert on the primary that the tenant is gone.

Red on `main` @ `a85489e0`, because the request itself reads the replica:

```
** (Mimic.UnexpectedCallError) expected Realtime.Repo.Replica.replica/0 to be called 0 time(s) but it has been called 1 time(s)
```

Green with the fix.

I also checked the user-facing outcome with a throwaway probe that stubs `Replica.replica/0` with a replica that hasn't seen the tenant (`get_by` → `nil`). On `main`: `204` and the tenant still exists. With the fix: `204` and the tenant is deleted.

## Validation

Local, Elixir 1.20.4 / OTP 29 (repo pins 1.19.5 / OTP 28.5.0.4), `supabase/postgres:15.14.1.167`:

- `mix test test/realtime_web/controllers test/realtime/api_test.exs test/realtime/tenants` — 509/510; the one failure is the pg-delta `SchemaTest` reconciliation test, which fails with `pgdelta not found on PATH` because the tool isn't installed in my environment. It doesn't exercise this code path.
- `mix compile --force --warnings-as-errors` — no warnings from the changed files; the warnings it reports are pre-existing, in `forum/muster/scope.ex`, `api.ex`, `application.ex` and `migrations.ex`, under 1.20
- `mix format --check-formatted`, `mix credo` (strict config, no issues), `mix sobelow`, `mix dialyzer` (passed), `git diff --check` — clean

Not run locally: the full 4-partition × 5-postgres matrix — left to CI.


lib/realtime_web/controllers/tenant_controller.ex
```diff
@@ -192,7 +192,7 @@ defmodule RealtimeWeb.TenantController do
   def delete(conn, %{"tenant_id" => tenant_id}) do
     stop_all_timeout = Enum.count(PostgresCdc.available_drivers()) * 1_000
 
-    with %Tenant{} = tenant <- Api.get_tenant_by_external_id(tenant_id, use_replica: false),
+    with %Tenant{} = tenant <- Api.get_tenant_by_external_id(tenant_id, use_replica?: false),
```
leandrocp, line 195: Since we're here let's add `Keyword.validate!` into `get_tenant_by_external_id/2` to avoid future mistakes:

```elixir
def get_tenant_by_external_id(external_id, opts \\ []) do
  opts = Keyword.validate!(opts, use_replica?: true)
  ...
```

leandrocp (changes requested):
Nice finding. Just one suggestion to make it safer.

leandrocp (approved):
Thanks @hsusul !

PR #2228, opened 2026-09-17
Consolidate extension tests into one canonical location mirroring lib/

mandar1045:
Resolves #2223

###What changed:

Consolidated all CDC extension tests into test/extensions/postgres_cdc_rls/ so the test tree correctly mirrors the lib/ tree.
Moved the 5 unique test files and updated their defmodule names to match the new path.
Safely merged the duplicate replications_test.exs files (brought over the unique drop_replication_slot/2 and special characters tests without losing the target file's setup).
Removed the old test/realtime/extensions/cdc_rls/ directory.

PR #2230, opened 2026-09-17
test: consolidate cdc_rls extension tests into test/extensions/postgres_cdc_rls

### LINKED ISSUE #2223: Consolidate extension tests into one canonical location mirroring lib/
## Problem

Extension tests live in two trees:

```
test/extensions/postgres_cdc_rls/
test/realtime/extensions/cdc_rls/
```

The code lives in `lib/extensions/postgres_cdc_rls/`. The second tree matches neither path. It adds a `realtime/` prefix that `lib/` does not have, and it shortens `postgres_cdc_rls` to `cdc_rls`.

Because the paths do not line up, you can break a function, run what looks like its test file, and still get a green run.

## Fix

Move the six files in `test/realtime/extensions/cdc_rls/` into `test/extensions/postgres_cdc_rls/`, so the test tree mirrors `lib/`.

Two things to watch:

- `replications_test.exs` exists in both trees and covers the same module. Merge it, do not overwrite.
- `test/realtime/api/extensions_test.exs` covers `lib/realtime/api/extensions.ex`. Leave it where it is.

No behavior changes. The same tests should run before and after.

Ref REAL-1014

akshar27:
## Summary

Closes #2223.

The `postgres_cdc_rls` extension code lives in `lib/extensions/postgres_cdc_rls/`, but its tests were split across two trees, neither of which matched that path:

- `test/extensions/postgres_cdc_rls/` — partial, matches `lib/`
- `test/realtime/extensions/cdc_rls/` — the rest, adds a `realtime/` prefix `lib/` doesn't have and shortens `postgres_cdc_rls` to `cdc_rls`

Because the paths didn't line up, a broken function could still show a green run against the wrong test file.

## What changed

Moved the five non-overlapping files from `test/realtime/extensions/cdc_rls/` into `test/extensions/postgres_cdc_rls/` as straight renames (no content changes):

- `cdc_rls_test.exs`
- `replication_poller_test.exs`
- `subscription_manager_distributed_test.exs`
- `subscription_manager_test.exs`
- `subscriptions_test.exs`

`replications_test.exs` existed in both trees, testing the same module from different angles, so per the issue this one was merged rather than overwritten:

- Kept the canonical version's describe blocks: `prepare_replication/2`, `terminate_backend/2`, `get_pg_stat_activity_diff/2`, and the five `list_changes/5` slot_changes_count scenarios.
- Added the two describe blocks the other copy had that the canonical one didn't: `drop_replication_slot/2` (not covered at all in the canonical file) and list_changes for schemas/tables with special characters (spaces, commas, quotes, emoji, etc. — also not covered in the canonical file).
- Tests where both files covered the same behavior (just with different setup styles) were not duplicated — canonical's versions already exercise that behavior.

`test/realtime/api/extensions_test.exs` is left where it is, per the issue — it covers `lib/realtime/api/extensions.ex`, a different module.

No production code changed.

## Test plan

`mix test test/extensions/postgres_cdc_rls/` against a local realtime + tenant Postgres (via `mix setup && mise run db-start`): **12 doctests, 205 tests, 0 failures**. Same test count/coverage as the sum of both trees before the merge, minus the genuinely duplicated cases.

`mix format --check-formatted` and `mix credo --strict` both clean.


PragTob:
Heyo, sorry but I'm afriad this was already solved in #2228 - sorry for the race condition! Thank you for the effort still! 💚

PR #2231, opened 2026-09-18
test: fix max_wal_senders test flakes caused by missing public.test table

mandar1045:
As discussed in #2228, taking a look at some of the flaky tests.

This fixes the `max_wal_senders` flakes in `connect_test.exs` and `replication_connection_test.exs`.

**The issue:** The test workers were crashing because they assumed a `public.test` table existed. This was only true if the pooled tenant database happened to be dirtied by a previous CDC test. On a fresh database, the workers crashed instantly instead of cleanly hitting the WAL sender limit.

**The fix:** Explicitly run `CREATE TABLE IF NOT EXISTS public.test (id INT)` right before spawning the replication connections to guarantee the table exists and eliminate the race condition.


PragTob:
Thank you! 💚 

Confirmed `mix test test/realtime/tenants/connect_test.exs:620` fails on main in isolation.

Checked `mix test test/realtime/tenants/replication_connection_test.exs:727` - that one doesn't fail so I'm not confident it relies on "dirty" state.

Looking into it a bit!

mandar1045:
> Thank you! 💚
> 
> Confirmed `mix test test/realtime/tenants/connect_test.exs:620` fails on main in isolation.
> 
> Checked `mix test test/realtime/tenants/replication_connection_test.exs:727` - that one doesn't fail so I'm not confident it relies on "dirty" state.
> 
> Looking into it a bit!

Yeah,  replication_connection_test.exs does pass on its own, but the missing public.test table is still what’s causing the flake in connect_test.exs.

The other test creates that table through Integrations.setup_postgres_changes, while connect_test.exs doesn’t. Since the test DB can be reused, connect_test.exs was sometimes just getting lucky and using a DB where the table already existed. On a fresh DB, it fails with public.test not found.

Since both tests are covering the same max_wal_senders behavior, I figured it made sense to explicitly set up the table in both.

test/realtime/tenants/replication_connection_test.exs
```diff
@@ -729,6 +729,10 @@ defmodule Realtime.Tenants.ReplicationConnectionTest do
       opts = Database.opts(settings)
       parent = self()
 
+      {:ok, db_conn} = Database.connect(tenant, "realtime_test", :stop)
+      Postgrex.query!(db_conn, "CREATE TABLE IF NOT EXISTS public.test (id INT)", [])
+      GenServer.stop(db_conn)
```
PragTob: I don't think this test needs it, the setup should always set it up via `Integrations.setup_postgres_changes(db_conn)`

test/realtime/tenants/connect_test.exs
```diff
@@ -622,6 +622,10 @@ defmodule Realtime.Tenants.ConnectTest do
       opts = Database.opts(settings)
       parent = self()
 
+      {:ok, db_conn} = Database.connect(tenant, "realtime_test", :stop)
+      Postgrex.query!(db_conn, "CREATE TABLE IF NOT EXISTS public.test (id INT)", [])
+      GenServer.stop(db_conn)
```
PragTob: This looks to be the only test here that needs the `"test"` database - do we really need it? Maybe just setting the table to `:all` would do this trick?

PragTob (commented):
Thank you! 💚 

I believe one of the tests doesn't need this change though and the other one might have another solution (see inline comments)

Image

PragTob:
@mandar1045 sorry didn't see the comment before posting, but in this case the other test already has the "test" table created via its setup so I don't think we need to explicitly create it a second time there (it doesn't do harm, but it also doesn't really help imo)

mandar1045:
ohh, yeah thats right, 

I totally missed that `Integrations.setup_postgres_changes` was creating `public.test` in the setup for `replication_connection_test.exs`. explains why it was passing on its own but failing in `connect_test.exs`.

`table: :all` makes a lot more sense here anyway. It works with an empty DB, so we don't need to create the table just for the test.

I'll update it  and switch it to `table: :all`. Thanks for catching that.

PragTob (approved):
Thanks!

mandar1045:
Also, are there any other issues, features, or areas of the codebase you'd like some help with? I'd love to pick up another task and keep contributing, just point me in the right direction, thanks ❤️

PR #2234, opened 2026-09-18
fix: prevent max_concurrent_users leak when clients leave channels

### LINKED ISSUE #2233: max_concurrent_users  is leaking  when leaving channels
## Issue

When a client unsubscribes from a channel, `RealtimeChannel.terminate/2` doesn't call `UsersCounter.remove`. This means the user stays in the census until the idle socket sweeper (`Tracker`) eventually kills the connection, which can take up to 10 minutes.

I found this while debugging an SPA where users frequently switch between channels. The connected user count kept increasing because users weren't being removed when they left a room, eventually causing `"Too many connected users"` errors.

### Consequences

1) Unnecessary Postgres connections: Ghost users can keep tenant Postgres pools alive for up to 10 minutes, potentially wasting connections.
2) Customer friction: Stale users can consume `max_concurrent_users` slots, causing customers to hit their limits sooner and trigger rate limiting.

### To Reproduce

1) Open a WebSocket without joining a channel — uses 0 user slots.
2) Join a channel — `UsersCounter.add/2` is called, using 1 slot.
3) Call `channel.unsubscribe()` — `RealtimeChannel.terminate/2` runs.
4) **Bug:** `UsersCounter.remove` is never called. The user has no active channels but still counts as 1 user for up to 10 minutes.

### Proposed Fix

Remove the user from `UsersCounter` when their active channel count drops to 0 in `terminate/2`:

`
if Tracker.untrack(transport_pid) <= 0 do
  Realtime.UsersCounter.remove(transport_pid, tenant_id)
end
`

mandar1045:
Fixes #2233

### TL;DR

When a client unsubscribed from a channel, they stayed in `UsersCounter` until the idle socket cleanup ran, which could take up to 10 minutes.

This PR removes the user from `UsersCounter` as soon as their active channel count reaches `0`, freeing the concurrent user slot immediately.


edgurgel (changes requested):
Thanks for the contribution!

We really need some proper tests for this including integration tests with multiple channels.

There is also a missing thing here which is there is no guarantee that `terminate` will be called as a process can be killed, crash and so on. We don't need to solve this but it's something that won't be covered by `terminate`. Tracker will eventually stop sockets with no channels anyway so not a huge deal.

I think ultimately this might need to live in a custom Socket code that we might need to do instead of using Phoenix.Socket as of today. Future work though

mandar1045:
> Thanks for the contribution!
> 
> We really need some proper tests for this including integration tests with multiple channels.
> 
> There is also a missing thing here which is there is no guarantee that `terminate` will be called as a process can be killed, crash and so on. We don't need to solve this but it's something that won't be covered by `terminate`. Tracker will eventually stop sockets with no channels anyway so not a huge deal.
> 
> I think ultimately this might need to live in a custom Socket code that we might need to do instead of using Phoenix.Socket as of today. Future work though

Sure , thanks for letting me know, I'll work on this and push the changes soon

mandar1045:
> Thanks for the contribution!
> 
> We really need some proper tests for this including integration tests with multiple channels.
> 
> There is also a missing thing here which is there is no guarantee that `terminate` will be called as a process can be killed, crash and so on. We don't need to solve this but it's something that won't be covered by `terminate`. Tracker will eventually stop sockets with no channels anyway so not a huge deal.
> 
> I think ultimately this might need to live in a custom Socket code that we might need to do instead of using Phoenix.Socket as of today. Future work though

Thanks for the review, @edgurgel! Agreed on all points. I’ll add integration tests for multiple channels per socket and clarify in the PR that `terminate` only handles graceful unsubscribes, with Tracker still acting as the safety net.

Custom socket definitely sounds like the right long-term fix. Happy to help with that too — I can take a first pass or open a tracking issue if that helps.

mandar1045:
Pushed the multi-channel integration tests covering single/multi-channel leave, rejoin, socket isolation, and `max_concurrent_users` enforcement.

Also added a comment in `terminate/2` noting that it isn't guaranteed to run on kill/crash, with Tracker's sweep as the fallback.

One thing I noticed while testing: if a join and the last channel's `terminate` interleave on the same transport, `remove` can land after `add`, leaving a live user uncounted until they join another channel. I’ve left that for the custom socket work, and I’d be happy to he

PR #2236, opened 2026-09-18
chore(tests): only run coverage on pg17

leandrocp:


.github/workflows/tests.yml
```diff
@@ -130,6 +130,7 @@ jobs:
           MIX_TEST_PARTITION: ${{ matrix.partition }}
           MATRIX_POSTGRES: ${{ matrix.postgres }}
           JOB_URL: ${{ steps.job_url.outputs.url }}
+          EXPORT_COVERAGE: ${{ matrix.postgres == 'pg17' }}
```
PragTob: highly optional: these days I prefer to give the matrix an additional attribute like (coverage: true) - means if I change the matrix I don't need to go hunting for references elsewhere (as has many times broken my build 😭 )

PragTob (approved):
Looks good to me, very optional and incremental comment!

Image

.github/workflows/tests.yml
leandrocp: Sounds good, done!

PR #2237, opened 2026-09-18
fix(auth): support tenants that only have a jwt_jwks

Snehil-Shah:
Following up on #1646, tenants with a null jwt secret using only jwks crash on every websocket upgrade and tenant API request, as the code unconditionally tries to decrypt a null jwt secret. This PR adds necessary support for tenants with only jwks auth.


lib/realtime_web/channels/auth/channels_authorization.ex
```diff
@@ -3,10 +3,19 @@ defmodule RealtimeWeb.ChannelsAuthorization do
...
   """
 
+  alias Realtime.Crypto
+
+  @doc """
+  Decrypts a tenant's `jwt_secret`.
+  """
+  @spec decrypt_jwt_secret(binary() | nil) :: binary() | nil
+  def decrypt_jwt_secret(nil), do: nil
+  def decrypt_jwt_secret(jwt_secret), do: Crypto.decrypt!(jwt_secret)
```
Snehil-Shah, line 13: Also, this is an opinionated refactor, as the three call sites needed nil handling. Let me know if we would prefer to add a nil clause to `Crypto.decrypt!/1` itself (I didn't do it as it has other callers too) or move this helper elsewhere.

lib/realtime_web/channels/auth/jwt_verification.ex
```diff
@@ -145,10 +145,16 @@ defmodule RealtimeWeb.JwtVerification do
     else
       # If there's no JWK, and HS* is being used, instead of erroring, try
       # the jwt_secret instead.
-      {:ok, Joken.Signer.create(alg, jwt_secret)}
+      if jwt_secret,
```
leandrocp: An empty `""` jwt secret would match here so we need to verify the content as well:

```suggestion
      if is_binary(jwt_secret) and jwt_secret != "",
```

And also add a test to make sure.

lib/realtime_web/channels/auth/jwt_verification.ex
```diff
@@ -47,7 +47,7 @@ defmodule RealtimeWeb.JwtVerification do
   @doc """
   Verify JWT token and validate claims
   """
-  @spec verify(binary(), binary(), binary() | nil) ::
+  @spec verify(binary(), binary() | nil, binary() | nil) ::
           {:ok, map()}
           | {:error, Joken.error_reason()}
           | {:error, {:error_generating_signer, binary()}}
```
leandrocp, line 53:
```suggestion
          | {:error, {:error_generating_signer, binary()}}
          | {:error, :error_generating_signer}
```

lib/realtime_web/channels/auth/channels_authorization.ex
```diff
@@ -3,10 +3,19 @@ defmodule RealtimeWeb.ChannelsAuthorization do
   Check connection is authorized to access channel
   """
 
+  alias Realtime.Crypto
+
+  @doc """
+  Decrypts a tenant's `jwt_secret`.
+  """
+  @spec decrypt_jwt_secret(binary() | nil) :: binary() | nil
+  def decrypt_jwt_secret(nil), do: nil
```
leandrocp: I think we can move `decrypt_jwt_secret/1` to `Realtime.Api.Tenant` next to `encrypt_jwt_secret/1` since it's related to tenants.

leandrocp (changes requested):
Hi @Snehil-Shah LGTM just two small suggestions.

lib/realtime_web/channels/auth/channels_authorization.ex
Snehil-Shah: @leandrocp Thanks for finding it the right home😅

lib/realtime_web/channels/auth/jwt_verification.ex
Snehil-Shah: Good catch.

PR #2238, opened 2026-09-18
Use wait_for_it library in tests

jvoegele:

Use @jvoegele's [wait_for_it](https://hex.pm/packages/wait_for_it) in tests.

Tests used a variety of hand-rolled poll/sleep waiting mechanisms, and bare `assert`/`refute` against truthy values resolved by the custom waiting mechanisms.

## What is the new behavior?

- Use `wait_for_it`'s `assert_eventually`, `refute_eventually`, and `assert_always` for "patient" assertions. This results in, in many cases, better assertion failure messages vs "expected truthy, got falsy" kind of messages.
- Replace hand-rolled poll/sleep loops with `WaitForIt` waiting constructs, such as `case_wait`, `match_wait`, etc.
- Retain the default wait budget so tests would not suddenly start failing due to silent differences in defaults.

## Additional context

I hand wrote the first two commits, then let Claude Code do a big, sweeping pass to find other places where `wait_for_it` might provide some benefit. I'll let Claude speak for itself about those changes:

## Verbose Claude Code Changes

ddcd9cc7 swapped `assert eventually(...)` for `assert_eventually`, i.e. the
waiting that already looked like waiting. This goes after the rest: 14
hand-rolled polling helpers across 58 call sites, and 100+ of the 143 blocking
sleeps that sat between a trigger and an assertion.

The strategy, roughly in the order the decisions were made:

1. Classify each sleep by what FOLLOWS it, not by how long it is.

   - "sleep, then assert something happened" -> `assert_eventually`. The sleep
     was a guess at how long an async effect takes; the wait returns as soon as
     the effect lands and reports the last observed value on failure.
   - "sleep, then assert something did NOT happen" -> `assert_always`. This is
     the larger category, and the one the suite had no tool for: sampling an
     invariant once at the end of a window cannot catch a regression in the
     middle of it. Same wall clock, strictly stronger assertion. 18 sites.
   - "sleep, then read state that a message already ordered" -> delete the
     sleep. `:sys.get_state/1` queues behind the message under test, so it
     already observes the state after that clause has run.

2. Replace hand-rolled loops with the construct matching their shape, so the
   failure message comes for free.

   The `Enum.reduce_while` and recursive retry helpers returned nil on
   exhaustion, which surfaced later as a FunctionClauseError or MatchError
   naming neither the condition nor the last value seen. `case_wait`'s `else`
   clause matches the last value that was evaluated, so the flunk/raise can say
   what it actually got. `wait!`/`match_wait!` cover the ones that only needed
   a value out.

3. Preserve every wait budget; never shorten one.

   Each converted helper kept its original timeout (attempts x interval). The
   two probes against an external dependency - docker container readiness and a
   gen_rpc TCP port - moved to `Backoff.exponential`, so the common fast case
   stops paying a flat poll interval across a 25s budget.

4. Leave alone anything that is not a wait for a condition.

   Sleeps inside Mimic stubs simulate slowness; `for _ <- 1..10` loops with a
   sleep are load generation; a sleep past a rate-limit window or a token expiry
   is genuine time passage. assert_receive/refute_receive already are the right
   tool where a message exists and were not converted to polling. Retry loops
   that retry on a raised exception (TestTenantDb.reset_realtime_schema!) are
   not expressible as a wait at all, since an exception propagates out of one.

5. Keep side effects out of waited-on expressions unless they are idempotent.

   A waited expression is re-evaluated an indeterminate number of times.
   seed_remote_cache/3 re-pushes a cache entry on each attempt, which is how it
   wins against a concurrent invalidation - idempotent, so it stays inside the
   wait. Anything that inserts or increments does not; in janitor_test the DB
   connections are opened once and only the query is polled.

Fixes a real bug along the way: RateCounterHelper.await_initial_tick/1 was an
unbounded hot spin on :sys.get_state with no sleep and no timeout, so a counter
that never ticked would pin a core and hang the suite with no diagnostic.

Adds a WAIT_MARGIN_REPORT=true diagnostic to test_helper.exs. wait_for_it's
[:wait_for_it, :wait, :stop] telemetry reports how many evaluations a wait
actually used, so a wait that succeeded on its 48th of 50 attempts - a flake
that has not happened yet - becomes visible on a green run, which the existing
rerun-based flaky digest cannot see. Off by default.

Sequential blocking sleeps: 60.1s across 143 sites -> 20.6s across 41.


test/extensions/postgres_cdc_rls/cdc_rls_test.exs
```diff
@@ -534,8 +516,8 @@ defmodule Extensions.PostgresCdcRlsTest do
 
       RealtimeWeb.Endpoint.subscribe(Realtime.Syn.PostgresCdc.down_topic(external_id))
 
-      assert eventually(fn -> match?({:ok, _, _}, PostgresCdcRls.get_manager_conn(external_id)) end)
-      assert {:ok, manager_pid, _conn} = PostgresCdcRls.get_manager_conn(external_id)
+      assert_eventually({:ok, manager_pid, _conn} = PostgresCdcRls.get_manager_conn(external_id))
```
leandrocp, line 519: What if wait_for_it export a formatter config to remove parens for this and the other macros? It's a personal preference but I think it would read better as `assert_eventually expr` (similar to `assert`, `refute`, etc).

jvoegele: `wait_for_it` already had this rule in its own formatter config, but it was not being included in the release tarball so the rule was not being applied. I released a new version that includes the exported formatter rules, and reformatted so that `assert_eventually` does not use parens, except for some cases of multi-line expressions that format better with the parens.

seangeo (commented):
This looks good to me, caveat emptor on my review though. ;) But definitely some readability wins there, especially removing all those Process.sleeps.

I wonder if some of those comments can be deleted though, quite a lot look like Claude describing it's "reasoning" rather than things that make sense after the change is done.

test/realtime/monitoring/prom_ex/plugins/tenant_test.exs
```diff
@@ -362,8 +345,7 @@ defmodule Realtime.PromEx.Plugins.TenantTest do
         hit: true
       })
 
-      Process.sleep(100)
-      assert (metric_value(metric, tenant: external_id, hit: true) || 0) == metric_value
```
PragTob, line 366: I believe the swap is faulty here.

We're testing that a counter increase is _not_ counted. I believe the sleep was there to give it time to potentially increase the counter. 

Not sure, how to best fix this though (i.e. we'd need to make sure the event that could increase the counter has already passed/executed). Probably swap to `assert_always` as in another place should do the job.

test/support/clustered.ex
```diff
@@ -183,33 +193,30 @@ defmodule Clustered do
...
-      TestHelpers.eventually(fn -> Env.port_available?(port) end,
-        retries: @port_wait_attempts,
-        sleep: @port_wait_delay_ms
-      )
-
-    if !available? do
+    unless wait(Env.port_available?(port),
+             timeout: @port_wait_timeout_ms,
+             interval: @port_wait_interval_ms
+           ) do
```
PragTob: `unless` is soft-deprecated, will start emitting warnings in a couple of versions - we should stop using it imo

PragTob: I also slightly prefer the previous version with an extracted varaible (multi line conditionals are a bit harder to read imo), but that's style and very optional

test/test_helper.exs
```diff
@@ -81,6 +81,71 @@ ExUnit.after_suite(&TestTenantDb.shutdown/1)
 # the rate has to be reported explicitly or it disappears from CI entirely.
 ExUnit.after_suite(&TestTenantDb.report_unhealthy_checkouts/1)
 
+# A wait that succeeded on its 48th of 50 attempts is a flake that has not happened yet, and a
+# green run says nothing about it. `[:wait_for_it, :wait, :stop]` reports how many evaluations a
+# wait actually took, so with WAIT_MARGIN_REPORT=true every wait that timed out or burned more
+# than half its budget is written to wait-margin.jsonl for the flaky digest to pick up.
```
PragTob: This is a really good idea, however this is not picked up (or triggered) right now as best as I can tell and since this is an entirely separate feature I think we could move it out to its own PR.

test/integration/rt_channel/billable_events_test.exs
```diff
@@ -239,13 +239,13 @@ defmodule Realtime.Integration.RtChannel.BillableEventsTest do
...
-    eventually(
+    until(
       fn ->
         RateCounterHelper.tick_tenant_rate_counters!(tenant)
         Enum.all?(expected, fn {key, value} -> get_count(key, tenant) == value end)
       end,
-      retries: 10,
-      sleep: 50
+      timeout: 500,
+      interval: 50
```
PragTob: `eventually` should work here, right? Would be nice for consistency (suddenly having `until` made me stumble a bit)

test/integration/distributed_realtime_channel_test.exs
```diff
@@ -66,14 +66,11 @@ defmodule Realtime.Integration.DistributedRealtimeChannelTest do
...
-                 Muster.view_hash(scope) == :erpc.call(node, Muster, :view_hash, [scope])
-             end,
-             retries: 150,
-             sleep: 100
-           )
+    assert_eventually(
+      Muster.status(scope) == :ready and
+        :erpc.call(node, Muster, :status, [scope]) == :ready and
+        Muster.view_hash(scope) == :erpc.call(node, Muster, :view_hash, [scope]),
+      timeout: :timer.seconds(15)
```
PragTob: nit/mega optional: imo [to_timeout/1](https://elixir.hexdocs.pm/Kernel.html#to_timeout/1) is nicer :) `to_timeout(second: 15)`

test/realtime/monitoring/prom_ex/plugins/muster_test.exs
```diff
@@ -49,19 +54,38 @@ defmodule Realtime.PromEx.Plugins.MusterTest do
...
+    end
+  end
 
-      assert value(metrics, state: "ready") == 0
-      assert value(metrics, state: "converging") == 0
-      assert value(metrics, state: "rebalancing") == 0
+  # PromEx polls every 100ms, so the stub a test just installed only reaches the gauge on the next
+  # poll. Waiting for the one-hot shape identifies a post-stub snapshot unambiguously, where a
+  # fixed sleep only guesses at how many cycles that takes.
+  defp poll_until(expected_state) do
```
PragTob, line 67: like test helpers like this one 👌

test/realtime/syn_handler_test.exs
```diff
@@ -216,7 +220,7 @@ defmodule Realtime.SynHandlerTest do
 
           assert :peer.call(peer_pid, Process, :alive?, [remote_pid])
 
-          refute Process.alive?(local_pid)
+          assert_eventually not Process.alive?(local_pid)
```
PragTob: I believe we can go back to the old version here or remove the assert in total - since we already called `assert_process_down` on the same pid.

PragTob (commented):
👋 

Hello, thank you for your PR and welcome to your first Tobi review! I tend to leave a lot of comments, that doesn't mean I think something is bad - quite the opposite often. I do like to discuss code and patterns and learn from each other. Hence, I'll ask questions or often times also just praise code I particularly like.
There's no need to address all of my comments - I try to make it clear which ones are optionals/just nits and which ones matter more - both in the comments themselves and in my review summary. Rather than keep going back and forth I'm also always available for a quick call to hash things out - as that can often be more efficient (after a round or 2).

Aight, on with the review!

---------------------------------

Thanks for the work, I'm glad to see all those sleeps gone 🔥 

That said, there are _a lot of changes_ here which makes the review quite hard. I'd have preferred the changes over a couple of smaller PRs that would be easier to review & ship independently.

* introduce `wait_for_it` and use it instead of `eventually`
* Replace `Process.sleep` with new `assert_eventually` etc. calls - personally I think I'd also have doen this a couple of folders/files at a time to keep scope and merge conflict risk down
* The structural changes to our docker helper (i.e. it's not a pure swap here, but new mechanisms were added) - I think they look good and are helpful, but I think they're a separate concern from the remainder of the PR, that change is easy to miss alongside all the other changes here
* the `WAIT_MARGIN_REPORT` (in `test_helper.exs`) is a very cool idea, but similar to the docker changes is a separate thing I'd have preferred in a separate PR - as best as I can see it's also currently not activated on CI nor does the flaky digest report pick it up (as the comment implies) - which I think can be fixed in another PR

I think I'd at least remove the wait margin report for now/delegate it to another PR as I don't want it to hold up this PR from landing.

Other than that, some of the comments are overly verbose & repetitive as Sean mentioned which would be good to cut down imo.

Other than that:
* I think one of the swaps is faulty/should swap to `assert_always` probably
* not reintroducing `unless` would be really nice, but not super important (although we're likely to switch it back again)
* the others are more optional

Image

test/support/clustered.ex
leandrocp: Agree with @PragTob to avoid `unless` 👍🏻

test/integration/gcm_encryption_backfill_test.exs
```diff
@@ -44,7 +45,8 @@ defmodule Realtime.Integration.GcmEncryptionBackfillTest do
       assert {:ok, _conn} = Database.connect(tenant, "realtime_backfill_before", :stop)
 
       read_through_cache(external_id)
-      assert eventually(fn -> not is_nil(Api.get_tenant_by_external_id(external_id).gcm_migrated_at) end)
+
+      assert_eventually not is_nil(Api.get_tenant_by_external_id(external_id).gcm_migrated_at)
```
leandrocp: Isn't this the same as?

```suggestion
      assert_eventually Api.get_tenant_by_external_id(external_id)[:gcm_migrated_at]
```

test/integration/gcm_encryption_backfill_test.exs
```diff
@@ -93,11 +95,14 @@ defmodule Realtime.Integration.GcmEncryptionBackfillTest do
...
+    assert_always on_legacy_cipher?(external_id), timeout: @async_write_window, interval: @probe_interval
+  end
 
+  defp on_legacy_cipher?(external_id) do
     untouched = Api.get_tenant_by_external_id(external_id)
-    refute Crypto.gcm?(untouched.jwt_secret)
-    assert Enum.all?(untouched.extensions, &Crypto.legacy_settings?(&1.settings, encrypted_settings_keys()))
-    assert is_nil(untouched.gcm_migrated_at)
+
+    not Crypto.gcm?(untouched.jwt_secret) and
```
leandrocp: Nitpick but the previous code would tell use which one failed, right?

test/integration/measure_traffic_test.exs
```diff
@@ -94,15 +94,16 @@ defmodule Realtime.Integration.MeasureTrafficTest do
...
-      # Wait for RateCounter to run
-      RateCounterHelper.tick_tenant_rate_counters!(tenant.external_id)
-      Process.sleep(100)
-
-      output_bytes = get_count([:realtime, :channel, :output_bytes], tenant.external_id)
-      input_bytes = get_count([:realtime, :channel, :input_bytes], tenant.external_id)
-
-      assert output_bytes > 0
-      assert input_bytes > 0
+      assert_eventually(
```
leandrocp: Same as https://github.com/supabase/realtime/pull/2238/changes#r4063727547

By aggregating multiple asserts into a single one we might lose context when tests fails.

test/realtime/gen_rpc_pub_sub/region_rings_test.exs
```diff
@@ -125,9 +131,7 @@ defmodule Realtime.GenRpcPubSub.RegionRingsTest do
       Process.exit(ring_pid, :kill)
       assert_receive {:DOWN, ^ref, :process, ^ring_pid, :killed}
 
-      wait_until(fn ->
-        match?({:ok, _, _}, RegionRings.expected_router(@region, "tenant_0", table))
-      end)
+      assert_eventually match?({:ok, _, _}, RegionRings.expected_router(@region, "tenant_0", table)), @short_wait
```
leandrocp: I think `assert_eventually` doesn't need `match?` right? I see a bunch of `match?` assertions that seems like could be simplified 🤔

test/realtime/tenants/connect_test.exs
```diff
@@ -43,11 +47,10 @@ defmodule Realtime.Tenants.ConnectTest do
...
       Process.exit(pid, :kill)
       assert_process_down(pid)
-      # Wait to ensure that the process has not restarted
-      Process.sleep(1000)
 
-      # Temporary process should not be registered in syn
-      refute Connect.whereis(tenant.external_id)
+      # Temporary process must not be registered in syn at any point after it dies, so assert the
+      # absence holds.
+      assert_always is_nil(Connect.whereis(tenant.external_id)), timeout: 1_000, interval: 50
```
leandrocp, line 53: Should wait_for_it also have `refute_always`? 🤔 
Not a blocker obviously, just wondering because it might read better than `assert_always is_nil(...)`

test/realtime/monitoring/prom_ex/plugins/tenant_test.exs
jvoegele: Fixed in [8214cc1](https://github.com/supabase/realtime/pull/2238/commits/8214cc14e57453ce5675fada6a9e4d69e72a2bf5) using `assert_always` as you suggested.

test/support/clustered.ex
jvoegele: Resolved in [8214cc1](https://github.com/supabase/realtime/pull/2238/commits/8214cc14e57453ce5675fada6a9e4d69e72a2bf5)

test/test_helper.exs
jvoegele: I pulled this out and will put together a separate PR for the wait margin report.

test/integration/rt_channel/billable_events_test.exs
jvoegele: Fixed in [8214cc1](https://github.com/supabase/realtime/pull/2238/commits/8214cc14e57453ce5675fada6a9e4d69e72a2bf5)

test/integration/distributed_realtime_channel_test.exs
jvoegele: Agreed, changed to `to_timeout/1` in [8214cc1](https://github.com/supabase/realtime/pull/2238/commits/8214cc14e57453ce5675fada6a9e4d69e72a2bf5)

test/realtime/syn_handler_test.exs
jvoegele: Fixed in [8214cc1](https://github.com/supabase/realtime/pull/2238/commits/8214cc14e57453ce5675fada6a9e4d69e72a2bf5)

test/integration/gcm_encryption_backfill_test.exs
jvoegele: Fixed in. [2eecc0c](https://github.com/supabase/realtime/pull/2238/commits/2eecc0c29da434e9ce8c884738c867d71f7f98b5)

test/integration/gcm_encryption_backfill_test.exs
jvoegele: Fixed in [2eecc0c](https://github.com/supabase/realtime/pull/2238/commits/2eecc0c29da434e9ce8c884738c867d71f7f98b5) by breaking apart the big boolean expression into individual `assert`/`refute` statements (which is how it was originally).

test/integration/measure_traffic_test.exs
jvoegele: Fixed in [2eecc0c](https://github.com/supabase/realtime/pull/2238/commits/2eecc0c29da434e9ce8c884738c867d71f7f98b5) by refactoring into an `assert_counts` helper that provides the context if a test fails.

test/realtime/gen_rpc_pub_sub/region_rings_test.exs
jvoegele: Indeed, there were a few instances of this. Fixed in [2eecc0c](https://github.com/supabase/realtime/pull/2238/commits/2eecc0c29da434e9ce8c884738c867d71f7f98b5).

(NOTE: I left a couple instances of this in there, because they were compound matches.)

test/realtime/tenants/connect_test.exs
jvoegele: Yeah, it probably should have `refute_always`. I balked at adding this originally, but I can't recall the reason why.

jvoegele (commented):
Thanks for the detailed feedback, everyone.

I believe I've addressed all of the concerns that were raised, so if you'd like to have a look at the changes please feel free!

leandrocp (approved):
LGTM!

jvoegele:
> Thanks for the work, I'm glad to see all those sleeps gone 🔥
> 
> That said, there are _a lot of changes_ here which makes the review quite hard. I'd have preferred the changes over a couple of smaller PRs that would be easier to review & ship independently.

Thanks @PragTob, I promise future PRs will be smaller and easier to digest! 😬

PR #2240, opened 2026-09-19
fix(janitor): close connections when maintenance fails

### LINKED ISSUE #2239: Janitor leaves tenant database connections open when maintenance raises
### Describe the bug

`Realtime.Tenants.Janitor.MaintenanceTask.run/1` closes its tenant database connection only after both maintenance operations succeed. If deleting old messages or creating partitions raises, the cleanup line is skipped and the Janitor connection remains alive.

### To reproduce

A deterministic regression test can replace the tenant connection with a lightweight GenServer and make the first maintenance operation raise:

1. Return `{:ok, conn}` from `Realtime.Database.connect/2`.
2. Raise from `Realtime.Messages.delete_old_messages/1`.
3. Call `MaintenanceTask.run/1` and catch the expected exception.
4. Check `Process.alive?(conn)`.

On current `main`, step 4 returns `true`.

### Expected behavior

Once `Database.connect/2` succeeds, `MaintenanceTask.run/1` should always close that connection, including when maintenance raises. The original exception should still propagate.

### Actual behavior

`GenServer.stop(conn)` is inside the success-only `with` body, so exceptions and other non-success exits bypass connection cleanup.

### Impact

Janitor runs maintenance for tenant databases. A normal database error during message cleanup or partition maintenance can leave a short-lived Janitor connection orphaned, retaining a tenant database slot after the task fails. Repeated maintenance failures can accumulate unnecessary tenant connections.

### Root cause

The connection lifecycle is not protected by an `after` cleanup after a successful connect.

### Related work

#1206 added successful-path Janitor connection shutdown, #1294 improved tracking of long-lived Janitor connections, and #2161 removed failed task references from Janitor state. None closes the connection when `MaintenanceTask` raises after connecting.

hsusul:
## Summary

- close the Janitor tenant database connection in an `after` block once connect succeeds
- preserve the original maintenance exception
- add a regression test proving the connection is stopped when message cleanup raises

## Problem

`Realtime.Tenants.Janitor.MaintenanceTask.run/1` stopped its connection only after both maintenance operations succeeded. A database exception during message cleanup or partition creation bypassed that line and left the Janitor connection alive.

## Root cause

Connection cleanup lived in the success-only body of the `with` expression instead of covering the whole post-connect lifecycle.

## Solution

Wrap the maintenance operations in `try/after` after a successful `Database.connect/2`, so the connection is stopped on success and failure without changing the public return contract.

## Tests

The new regression test stubs a successfully opened connection, raises from `Messages.delete_old_messages/1`, verifies the exception still propagates, and asserts the connection process is no longer alive.

The test failed on clean `main` with:

```
Expected false or nil, got true
code: refute Process.alive?(conn)
```

It passes with this change.

## Validation

- `DB_PORT=56432 TEST_RUN=realtime_hunt_20260918 mix test test/realtime/tenants/janitor/maintenance_task_test.exs` — 4 passed
- `DB_PORT=56432 TEST_RUN=realtime_hunt_20260918 mix test test/realtime/tenants/janitor test/realtime/messages_test.exs` — 19 passed
- isolated rerun of the three load-sensitive failures from the full suite — 3 passed
- `mix credo --strict` — passed
- `mix deps.unlock --check-unused` — passed
- `mix sobelow --config` — passed
- `mix format --check-formatted` — passed
- Elixir 1.19.5 / OTP 28 `mix compile --force --warnings-as-errors` in Docker — passed
- `git diff --check` — passed

The broader local suite completed with 2,190/2,205 passing and 6 excluded. Twelve failures require the unavailable `pgdelta` binary. Three 100 ms rate/presence assertions failed under full-suite load and passed together on isolated rerun. The affected Janitor and message suites pass.

## Compatibility and risk

No public API or success-path return value changes. Cleanup is limited to the connection already owned by this maintenance task.

## Documentation

No documentation change is needed; this restores the intended internal connection lifecycle.

Closes #2239


edgurgel (changes requested):
If the task crashes, the connection closes as they are linked. We technically don't even need to call `GenServer.stop/1` there

edgurgel:
I guess we can check this out by checking if the connection process is closed after the task finishes? We can probably write a test that asserts this?

edgurgel (approved):

edgurgel:
Thank you!

PR #2242, opened 2026-09-19
fix(dashboard): close tenant version query connections

### LINKED ISSUE #2241: Tenant Info dashboard retains PostgreSQL connections after version lookup
### Reproduction

On current `main` (`e28e2ba48e170954b87c0e1cb926cef9c8cbd378`):

1. Open `/admin/dashboard/tenant_info?external_id=<tenant external_id>`.
2. Keep the LiveView open.
3. Query the tenant database:

```sql
SELECT count(*)
FROM pg_stat_activity
WHERE application_name = 'realtime_dashboard_tenant_info';
```

A deterministic LiveView regression test sees `2` retained backends after one lookup (the disconnected and connected render phases).

### Actual behavior

Each PostgreSQL version lookup starts a connection with `Database.connect_db/1` but never stops it. Because the connection is linked to the LiveView process, it remains open until that dashboard session exits. Repeated tenant lookups in the same session accumulate additional tenant database connections.

### Expected behavior

The one-shot connection used for `SELECT version()` should be stopped immediately after the query, on both success and failure, while preserving the existing displayed result or error.

### Root cause

`RealtimeWeb.Dashboard.TenantInfo.fetch_pg_version/1` calls `Database.connect_db/1` and `Postgrex.query/3` without a cleanup boundary.

### Impact

Operational use of the Tenant Info dashboard can consume tenant PostgreSQL connection slots for the lifetime of an admin LiveView session. Repeated lookups amplify the connection usage and can reduce the tenant's available database capacity.

hsusul:
## Summary

- close the one-shot tenant PostgreSQL connection after the dashboard version query
- guarantee cleanup when the query returns an error or raises/exits
- add a LiveView regression test that checks `pg_stat_activity` after a tenant lookup

Fixes #2241.

## Problem

The Tenant Info page started a PostgreSQL connection for `SELECT version()` on each lookup and left it linked to the LiveView. One page load retained two tenant database backends (the disconnected and connected render phases), and repeated lookups accumulated more until the LiveView exited.

## Solution

Run the version query inside an `after` cleanup boundary and stop the connection when it is still alive. Existing success and error return shapes are unchanged.

## Tests

- Regression test on clean upstream: failed with `pg_stat_activity` count `2` (expected `0`)
- `mix test test/realtime_web/dashboard/tenant_info_test.exs`: 19 passed
- `mix test test/realtime_web/dashboard` with pinned pg-delta: 66 passed
- `mix compile --force --warnings-as-errors` on Elixir 1.19.5 / OTP 28: passed
- `mix format --check-formatted`: passed
- `mix credo`: passed
- `mix deps.unlock --check-unused`: passed
- `mix sobelow --config .sobelow-conf`: passed
- `git diff --check`: passed

## Broader local run

A full `mix test` run completed with 2,185/2,204 passing and 7 excluded. The 19 failures were outside this change: tests requiring more than the single configured external tenant database, a `wal_keep_size = 32MB` test environment, and one replication-slot timing assertion. The affected Tenant Info and complete dashboard suites pass.


edgurgel (approved):

edgurgel:
Good finding!

PR #2244, opened 2026-09-20
fix: preserve metrics for active tenants

### LINKED ISSUE #2243: Metrics cleaner prunes active tenant series when Connect unregisters
### Description

`Realtime.MetricsCleaner` independently tracks stale websocket vacancies and stale `Realtime.Tenants.Connect` unregistrations. If a tenant still has local websocket users while its Connect process is unregistered (for example during a tenant database outage or Connect recovery), the stale disconnected-tenant entry currently prunes every Peep series tagged with that tenant.

This removes metrics for a tenant that is still actively serving websocket connections, so the tenant disappears from per-tenant observability precisely while its database connectivity is degraded.

### Reproduction

1. Emit a tenant-scoped metric such as `[:realtime, :connections]`.
2. Join a local websocket member with `Forum.Census.join(:users, tenant_id, pid)`.
3. Start `Realtime.MetricsCleaner` with a short cleanup interval and threshold.
4. Emit `[:syn, Realtime.Tenants.Connect, :unregistered]` for the same tenant.
5. Wait until the disconnected-tenant marker is stale.
6. Read `Realtime.TenantPromEx.get_metrics/0`.

The reproduction is deterministic in `Realtime.MetricsCleanerTest`: after the threshold, the tenant series is absent even though `Forum.Census.local_member_count(:users, tenant_id)` remains nonzero.

### Expected behavior

Metrics must remain while the tenant has local websocket members. Once the final local member leaves, the retained stale marker should allow the next cleanup pass to prune the tenant series.

### Actual behavior

The disconnected-tenant cleanup path prunes the tenant series solely from the age of the Connect unregistration marker. It does not consult websocket occupancy.

### Root cause

`loop_and_cleanup_metrics_table/2` selects every stale entry and calls `Peep.prune_tags/2`. That helper is shared by both the websocket-vacancy table and the Connect-unregistration table, but it does not enforce the core cleanup invariant that a tenant with local websocket users is still active. The subsequent broad `:ets.select_delete/2` also discards the marker, preventing deferred cleanup after those users leave.

### Impact

Operators can lose active-tenant connection, channel, Presence, Broadcast, and other tenant-tagged metrics during Connect/database interruption and recovery. This creates an observability blind spot on a normal production failure path.

hsusul:

Bug fix for tenant metric lifecycle cleanup.

A stale `Realtime.Tenants.Connect` unregistration causes `Realtime.MetricsCleaner` to prune every metric series for that tenant even when the node still has active websocket users. This creates an observability gap during tenant database or Connect interruption and recovery.

## What is the new behavior?

Cleanup candidates with local websocket members are deferred. Their stale marker is retained, so the series is pruned on a later pass after the final local user leaves.

The regression test exercises the production lifecycle directly: emit a tenant metric, join a real `Forum.Census` user, unregister Connect, cross the cleanup threshold, then leave the user. It was red on clean upstream because the active tenant series disappeared, and is green with this change.

## Validation

- `mix test test/realtime/metrics_cleaner_test.exs` (11 passed)
- `mix test test/realtime/monitoring/prom_ex/plugins/tenant_test.exs` (30 passed)
- `mix format --check-formatted`
- `mix credo --strict`
- `mix deps.unlock --check-unused`
- `mix compile --force --warnings-as-errors` in the CI-pinned Elixir 1.19.5 / OTP 28.5.0.4 image
- `git diff --check`

Closes #2243


edgurgel:
I've added this to the issue: https://github.com/supabase/realtime/issues/2243#issuecomment-5752944760

Closing for now!

PR #2245, opened 2026-09-20
fix(auth): return 401 on a malformed authorization header

Snehil-Shah:
Just a tiny consistency fix. It used to fall back to the 500 handler when given an auth header without 2 parts. Now returns a 401, the same as every other bad-credential path.


edgurgel (approved):

edgurgel:
Nice finding!

